PR #30, opened 2020-02-11
Validate zones order

timuthy:
This PR forces a zone order immutability for updates on the infrastructure and worker portions of a shoot. Since the AWS provider uses the zone index to create networks via Terraform as well as `MachineDeployments`, the index should be kept stable.

/cc @rfranzke - thanks for noticing!

**Release note**:

```improvement operator
The AWS validator now checks that zone orders in a shoot resources is retained (only additions are allowed).
```

### DIFF
--- a/pkg/apis/aws/validation/shoot_test.go
+++ b/pkg/apis/aws/validation/shoot_test.go
@@ -23,6 +23,7 @@ import (
 	. "github.com/onsi/gomega"
 	. "github.com/onsi/gomega/gstruct"
 	"k8s.io/apimachinery/pkg/util/validation/field"
+	"k8s.io/utils/pointer"
 )
 
 var _ = Describe("Shoot validation", func() {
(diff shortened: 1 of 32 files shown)


ialidzhikov:
`$index` is used also in the alicloud infra chart. We can keep in mind to add the same validation check also there once we introduce the validator for provider-alicloud.

rfranzke:
@ismail can you add this info to https://github.com/gardener/gardener-extension-provider-alicloud/issues/15?

timuthy:
As discussed with @rfranzke, I added the validation for workers as well.

pkg/apis/aws/validation/shoot_test.go
```diff
@@ -165,10 +168,65 @@ var _ = Describe("Shoot validation", func() {
...
+				Expect(errorList).To(ConsistOf(
+					PointTo(MatchFields(IgnoreExtras, Fields{
+						"Type":  Equal(field.ErrorTypeInvalid),
+						"Field": Equal("workers[0].zones"),
+					})),
+				))
+			})
+		})
 	})
 })
```
rfranzke: Can you add one more test that checks that adding a new zone while changing the existing zones is not allowed?

PR #32, opened 2020-02-11
Restrict cross validation for infra config and cloud profile

timuthy:
Only validate zones in `InfrastructureConfig` against zones in `CloudProfile` when:
a) Shoot is created
b) Zones in `InfrastructureConfig` are updated

This ensures that already running shoots won't break after a zone is removed from a cloud profile.

/cc @rfranzke.
I'll also file a PR for g/g's worker validation:
https://github.com/gardener/gardener/blob/master/plugin/pkg/shoot/validator/admission.go#L433-L440

**Release note**:

```improvement operator
The AWS validator only validates configured zones in the `InfrastructureConfig` against available zones in the referring `CloudProfile` when shoots are created or the zone information in `InfrastructureConfig` is updated. This prevent immediate breakage of already running shoots clusters whose zones were removed from the `CloudProfile`.
```

### DIFF
--- a/pkg/validator/shoot_validator.go
+++ b/pkg/validator/shoot_validator.go
@@ -19,6 +19,7 @@ import (
 	"errors"
 	"reflect"
 
+	"github.com/gardener/gardener-extension-provider-aws/pkg/apis/aws"
 	awsvalidation "github.com/gardener/gardener-extension-provider-aws/pkg/apis/aws/validation"
 
 	"github.com/gardener/gardener/pkg/apis/core"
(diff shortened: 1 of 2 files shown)


pkg/validator/shoot_validator.go
```diff
@@ -126,9 +120,41 @@ func (v *Shoot) validateShootUpdate(ctx context.Context, oldShoot, shoot *core.S
...
+	// This ensures that already running shoots won't break after a zone is removed from a cloud profile.
+	if len(infraConfig.Networks.Zones) > len(oldInfraConfig.Networks.Zones) {
+		cloudProfile := &gardencorev1beta1.CloudProfile{}
+		if err := v.client.Get(ctx, kutil.Key(shoot.Spec.CloudProfileName), cloudProfile); err != nil {
+			return err
+		}
+
+		if errList := awsvalidation.ValidateInfrastructureConfigAgainstCloudProfile(infraConfig, shoot, cloudProfile, infraConfigFldPath); len(errList) != 0 {
+			return errList.ToAggregate()
+		}
```
rfranzke: Nit: Can this be put into a separate function to prevent replication at lines 150-157?

Otherwise lgtm

pkg/validator/shoot_validator.go
```diff
@@ -126,9 +120,41 @@ func (v *Shoot) validateShootUpdate(ctx context.Context, oldShoot, shoot *core.S
 		}
 	}
 
+	// Only validate against cloud profile when zones are updated.
+	// This ensures that already running shoots won't break after a zone is removed from a cloud profile.
+	if len(infraConfig.Networks.Zones) > len(oldInfraConfig.Networks.Zones) {
```
danielfoehrKn: What about the case when one zone is deleted & one other zone is added -> would that not lead to no validation of the infra config against the CloudProfile?

timuthy: Good question!
This immutability is already checked here https://github.com/gardener/gardener-extension-provider-aws/blob/4633575f4689330401f4e0504d306fa7e37bbe94/pkg/apis/aws/validation/infrastructure.go#L153

timuthy:
> Though, can you add some unit tests for the validation?
Unit tests already exist for each validation function called by `shoot_validator.go`

timuthy:
Thanks for your reviews guys :)

PR #37, opened 2020-02-14
Filter api expose objects

mvladev:

Do not mutate kube-apiserver exposure resources when gardener manages those. See https://github.com/gardener/gardener/pull/1929

Fixes # n/a

~This is WIP as https://github.com/gardener/gardener/commit/40864019e030923c80657ac5d46a34130bcf337d is still not available in any tags and vendor folder manually updated~

**Release note**:

```improvement user
Do not mutate `kube-apiserver` exposure resources which Gardener marks as managed by it with `core.gardener.cloud/apiserver-exposure: gardener-managed` label.
```

### DIFF
--- a/pkg/webhook/controlplaneexposure/ensurer_test.go
+++ b/pkg/webhook/controlplaneexposure/ensurer_test.go
@@ -49,19 +49,32 @@ var _ = Describe("Ensurer", func() {
 	)
 
 	Describe("#EnsureKubeAPIServerService", func() {
-		It("should add annotations to kube-apiserver service", func() {
-			var (
-				svc = &corev1.Service{
-					ObjectMeta: metav1.ObjectMeta{Name: "kube-apiserver"},
-				}
-			)
+		var (
+			svc     *corev1.Service
+			ensurer genericmutator.Ensurer
+		)
 
-			// Create ensurer
-			ensurer := NewEnsurer(etcdStorage, logger)
+		BeforeEach(func() {
+			svc = &corev1.Service{
+				ObjectMeta: metav1.ObjectMeta{Name: "kube-apiserver"},
+			}
+			ensurer = NewEnsurer(etcdStorage, logger)
+		})
 
-			// Call EnsureKubeAPIServerService method and check the result
+		It("should not modify kube-apiserver service if SNI is enabled", func() {
+			svc.Labels = map[string]string{"core.gardener.cloud/apiserver-exposure": "gardener-managed"}
+			svcCopy := svc.DeepCopy()
+
 			err := ensurer.EnsureKubeAPIServerService(context.TODO(), dummyContext, svc)
 			Expect(err).To(Not(HaveOccurred()))
+
+			Expect(svc).To(Equal(svcCopy))
+		})
+
+		It("should add annotations to kube-apiserver service", func() {
+			err := ensurer.EnsureKubeAPIServerService(context.TODO(), dummyContext, svc)
+			Expect(err).To(Not(HaveOccurred()))
+
 			Expect(svc.Annotations).To(HaveKeyWithValue("service.beta.kubernetes.io/aws-load-balancer-connection-idle-timeout", "3600"))
 			Expect(svc.Annotations).To(HaveKeyWithValue("service.beta.kubernetes.io/aws-load-balancer-backend-protocol", "ssl"))
 			Expect(svc.Annotations).To(HaveKeyWithValue("service.beta.kubernetes.io/aws-load-balancer-ssl-ports", "443"))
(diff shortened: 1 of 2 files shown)


rfranzke:
Why don't you just simply vendor this commit?

mvladev:
I'll rather wait for a release and then I've to update `gardener-extensions` and then this

rfranzke:
OK, fair enough. FYI: @timuthy wanted to validate a new g/g release next week.

mvladev:
Okay, this is ready for review

example/controller-registration.yaml
```diff
@@ -18,7 +18,7 @@ spec:
...
...QaKKlZ6AF6uHW9RR8inMnpIQZoitMnNdQH2Vlyq7081q80TUv66vIq/XV9FkBnVsvJP6Xb03FY8F2h8fpQbG1KvScd8WB7kceTlFljzk0HOcwq6VeQXg6h8ayf3Ohdkg9Ay5NF4E6ugWr0pktQWWZJueognM90BLa8ycKJXcHOPknhJyL224PJdLAaW/mvtClfOak2GQU7ZSMYFVCPSq+mWLlVe/8ol6AKtw+dBQScsgMImQPxmiIx25jH7dpzzhTbkhbxJ6y7eac9LM8LaSN52Fbsc0lEJcym2ErEYFksBjh/qb5k3mGEMbGHXaXK3XIvIPUq8vFMs/mZaFLnS8LmXBBgprWXKkof8r6AEz61dYv1KVK+BxLprycazCz72Hfslle/tPq5IHmIBV/l/3qHj/s324v//9OKU0/1sv0r/Yub+QKc9au43Oc/7fduf22RGBJR/7pNM5lGPIx/JlgJNGYMTsdDTwSeZ/5/WvTYld1EDV+m93WsX8n26vuV//j1E2rf90k/6sQZzPzaA/edlh/atLAvKGwW5GQNX673SbxfV/eLRf/49SdPx3hwRwU2OkF0eKPsxY1R+jrJSrjx0useyiLSS5SBuL+iR/bIhBBQZeK/V9fvtGnhkO70IaqJHIM+CQRtCr0Olayh6Yhe0vIpD7wLL9+r8J6QN/B77y/GfF/m91Ovvff3+UUji5xklWvwjrFpZ6fTUsos9/2nW96t/ooMg5dwdZUGT7K6TQ95YWRGaxSyM6Szkqy2XG+pV85qqjdGig4ok6rVICqBp1em4oMOSY2dhetvsSzoCq1r/K6Piw/wCiYv23OodF+7/ZOuru1/9jFJ01XVt3W7nsrrL+0XFs1Pg15ioNSN+fbR2qvJLcNeJSAHnBt9XE5svLr4Wrr8ubq980X3lp5fI2bD4iYE1pLGpGciWiMU99+ilpZsJonxxlzWSq8IZW5vnOhmZrU1/7ZEp9ebsnf/izAZc6/lnbYDVntE9++qVWyACVdSmz1FU3pjmernZ5a75WCMmk9Kp0dIxwGLnmo+kpF+f42/9SpS5/JbRPWJL+/l+mX6AfDAUvf1/d+Fn1Qp65oLM+kUaHqJnn8iAutWUcnrz/owbKF6nSO1aWybdZeePeVMsFWhRH0liLCtt87oW5L/uyL/uyL/uyL/uyL/uyL/uyL/vyEcv/AyD6BOQAeAAA
```
tim-ebert: Why is the `chart` changed although you didn't change anything under `charts/`?

tim-ebert: I guess, the CI also fails because of this change:
```
> Check
Executing check-generate
go generate needs to be run:
 M example/controller-registration.yaml
make: *** [Makefile:53: check] Error 1
The command '/bin/sh -c make install-requirements && make VERIFY=true all' returned a non-zero code: 2
```

mvladev: `make generate` produced that result... I guess something in my dependencies / tools.

We have to start using `docker` for those commands as well.

pkg/webhook/controlplaneexposure/ensurer_test.go
```diff
@@ -49,19 +49,32 @@ var _ = Describe("Ensurer", func() {
...
+			svc = &corev1.Service{
+				ObjectMeta: metav1.ObjectMeta{Name: "kube-apiserver"},
+			}
+			ensurer = NewEnsurer(etcdStorage, logger)
+		})
 
-			// Create ensurer
-			ensurer := NewEnsurer(etcdStorage, logger)
+		It("should not modify kube-apiserver service if SNI is enabled", func() {
+			svc.Labels = map[string]string{"core.gardener.cloud/apiserver-exposure": "gardener-managed"}
```
tim-ebert: Can you use constants here?
```suggestion
			svc.Labels = map[string]string{v1beta1constants.LabelAPIServerExposure: v1beta1constants.LabelAPIServerExposureGardenerManaged}
```

pkg/webhook/controlplaneexposure/ensurer_test.go
```diff
@@ -74,6 +87,37 @@ var _ = Describe("Ensurer", func() {
 	})
 
 	Describe("#EnsureKubeAPIServerDeployment", func() {
+		It("should not modify kube-apiserver deployment if SNI is enabled", func() {
+			var (
+				dep = &appsv1.Deployment{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:   v1beta1constants.DeploymentNameKubeAPIServer,
+						Labels: map[string]string{"core.gardener.cloud/apiserver-exposure": "gardener-managed"},
```
tim-ebert:
```suggestion
						Labels: map[string]string{v1beta1constants.LabelAPIServerExposure: v1beta1constants.LabelAPIServerExposureGardenerManaged}
```

rfranzke:
@mvladev can you quickly address @tim-ebert's comments so that we can get this PR in? :)

pkg/webhook/controlplaneexposure/ensurer_test.go
mvladev: The reason why I use string literals instead of constants is that it will detect if the constant's value is changed (possible backwards-incompatible change)

mvladev:
> @mvladev can you quickly address @tim-ebert's comments so that we can get this PR in? :) 

See https://github.com/gardener/gardener-extension-provider-aws/pull/37#discussion_r394246555 - it guarantees that the test would fail if the constant is changed and it might break previous providers.

pkg/webhook/controlplaneexposure/ensurer_test.go
rfranzke: Hm, makes sense, indeed. :)

tim-ebert: Fair enough, ok for me then. 👍

PR #17, opened 2020-02-03
allow adding gateway endpoints to vpc

tareqhs:
**What this PR does / why we need it**:
Allow adding gateway endpoints to the VPC of an AWS cluster. Example for adding an endpoint for S3:
```yaml
apiVersion: core.gardener.cloud/v1alpha1
kind: Shoot
spec:
  provider:
    type: aws
    infrastructureConfig:
      apiVersion: aws.provider.extensions.gardener.cloud/v1alpha1
      kind: InfrastructureConfig
      networks:
        vpc:
          gatewayEndpoints:
          - "s3"
```

More info about gateway endpoints
https://docs.aws.amazon.com/vpc/latest/userguide/vpce-gateway.html

Relevant terraform resource
https://www.terraform.io/docs/providers/aws/r/vpc_endpoint.html

**Release note**:
```improvement operator
Now it is possible to define VPC Gateway Endpoints for AWS shoots. The new list "gatewayEdnpoints" inside "vpc" sections is a list endpoint service names. Each item is mapped to a vpc gateway endpoint for a service "com.amazonaws.<region>.<service>", region being the shoot region.
```

PR #41, opened 2020-02-25
allow modifying gateway endpoints inside vpc after shoot creation

tareqhs:

Continues https://github.com/gardener/gardener-extension-provider-aws/pull/17

Currently the whole VPC section is immutable although gateway endpoints should be mutable. The change makes the immutable check more specific to `vpc.id` and `vpc.cidr`, and leaves `gatewayEndpoints` as a mutable field.

**Release note**:
```improvement operator
Now it is possible to modify the gateway endpoints list `networks.vpc.gatewayEndpoints` after the shoot is created.
```

### DIFF
--- a/pkg/apis/aws/validation/infrastructure_test.go
+++ b/pkg/apis/aws/validation/infrastructure_test.go
@@ -365,7 +365,26 @@ var _ = Describe("InfrastructureConfig validation", func() {
 			Expect(errorList).To(BeEmpty())
 		})
 
-		It("should forbid changing the VPC", func() {
+		It("should allow changing gateway endpoints inside vpc", func() {
+			newInfraConfig := infrastructureConfig.DeepCopy()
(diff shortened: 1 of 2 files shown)


pkg/apis/aws/validation/infrastructure_test.go
```diff
@@ -374,7 +380,7 @@ var _ = Describe("InfrastructureConfig validation", func() {
 
 			Expect(errorList).To(ConsistOf(PointTo(MatchFields(IgnoreExtras, Fields{
 				"Type":  Equal(field.ErrorTypeInvalid),
-				"Field": Equal("networks.vpc"),
+				"Field": Equal("networks.vpc.cidr"),
```
rfranzke: Can you also add a test that tries to change the id of the VPC?

tareqhs: done

PR #49, opened 2020-03-17
Prevent infra controller to exit when there is still state

ialidzhikov:
Currently we early exit If there is no given  output variables in the terraform.tfstate. However there might be still important resources in the tfstate which are there and we should not early exit depending on variable presence.
Also terraform destroy discards the output variables on state update no matter if resource is still there or not.

Fixes #

**Release note**:

```improvement operator
The infrastructure controller does no longer release the finalizer of Infrastructure CR if it cannot find a given output variables. This allows terraform destroy to be invoked even when there is no output variables in the terraform.tfstate.
```

### DIFF
--- a/pkg/controller/infrastructure/actuator_delete.go
+++ b/pkg/controller/infrastructure/actuator_delete.go
@@ -51,16 +51,6 @@ func (a *actuator) Delete(ctx context.Context, infrastructure *extensionsv1alpha
 		return fmt.Errorf("error while checking whether terraform config exists: %+v", err)
 	}
 
-	stateVariables, err := tf.GetStateOutputVariables(aws.VPCIDKey)
-	if err != nil {
-		if apierrors.IsNotFound(err) || terraformer.IsVariablesNotFoundError(err) {
-			a.logger.Info("Skipping explicit AWS load balancer and security group deletion because not all variables have been found in the Terraform state.")
-			return nil
-		}
-		return err
-	}
-	vpcID := stateVariables[aws.VPCIDKey]
-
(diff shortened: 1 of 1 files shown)


pkg/controller/infrastructure/actuator_delete.go
```diff
@@ -77,6 +67,16 @@ func (a *actuator) Delete(ctx context.Context, infrastructure *extensionsv1alpha
...
 			Fn: flow.TaskFn(func(ctx context.Context) error {
+				stateVariables, err := tf.GetStateOutputVariables(aws.VPCIDKey)
+				if err != nil {
+					if apierrors.IsNotFound(err) || terraformer.IsVariablesNotFoundError(err) {
+						a.logger.Info("Skipping explicit AWS load balancer and security group deletion because not all variables have been found in the Terraform state.")
+						return nil
+					}
+					return err
+				}
+				vpcID := stateVariables[aws.VPCIDKey]
```
rfranzke:
```suggestion
```

pkg/controller/infrastructure/actuator_delete.go
```diff
@@ -77,6 +67,16 @@ func (a *actuator) Delete(ctx context.Context, infrastructure *extensionsv1alpha
...
+
 				if err := a.destroyKubernetesLoadBalancersAndSecurityGroups(ctx, awsClient, vpcID, infrastructure.Namespace); err != nil {
```
rfranzke:
```suggestion
				if err := a.destroyKubernetesLoadBalancersAndSecurityGroups(ctx, awsClient, stateVariables[aws.VPCIDKey], infrastructure.Namespace); err != nil {
```

PR #50, opened 2020-03-17
Decouple netowrk interface name

### LINKED ISSUE #52: MTU customizer unit fails if eth0 not available
**What happened**:

When creating a Shoot with an ubuntu image, the mtu-customizer.service (systemd) fails because it expects the existence of eth0 network interface.
The mtu customizer should the mtu of the primary network interface to a common value 1460. 

Tested on AWS with coreos and ubuntu.

mtu-customizer service should have run successfully.
MTU should be 1460.

 - AWS Shoot with a machine type that is capable of [ENA](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/enhanced-networking-ena.html) 
like m5.large 
- Machine image ubuntu (used ubuntu in version 18.4.20190617)

Execute `systemctl status mtu-customizer.service` and observe that it is in failed status.
Execute `ifconfig | grp eth0` and see that eth0 is missing.

Ubuntu defaults to using the ENA network interface that is not eth0.
**Environment**:

- Extension version:
- Others:

vpnachev:
Remove the hard dependency the network interface to be named `eth0`.
 
Fixes #52

**Release note**:

```improvement operator
The `eth0`  interface name is no longer hard-coded, instead the non-virtual network interface names are now dynamically detected.
```


charts/gardener-extension-provider-aws/templates/daemonset-custom-mtu.yaml
```diff
@@ -28,12 +28,21 @@ spec:
       initContainers:
       - name: mtu-customizer
         image: {{ include "image.alpine" . }}
-        command: ["ip", "link", "set", "dev", "eth0", "mtu", "1460"]
+        command:
+        - /mtu-customizer/mtu-customizer.sh
         securityContext:
           capabilities:
             add: ["NET_ADMIN", "SYS_TIME"]
+        volumeMounts:
```
rfranzke: Can you please add an annotation checksum to the annotations of this pod sec?

pkg/webhook/controlplane/ensurer.go
```diff
@@ -264,13 +264,18 @@ func (e *ensurer) EnsureAdditionalUnits(ctx context.Context, ectx genericmutator
 		command              = "start"
 		trueVar              = true
 		customMTUUnitContent = `[Unit]
-    Description=Apply a custom MTU to eth0
-    Requires=eth0.network
+Description=Apply a custom MTU to network interfaces
+After=network-online.target
```
rfranzke: Is `network-online.target` present everywhere?

zanetworker: isn't this somehow  tied to `NetworkManager` being on the system or not?

vpnachev: `network-online.target` is independent from implementation details like NetworkManger.

I've no strong opinion whether it should be `network.target` or `network-online.target`, see details about them here https://wiki.freedesktop.org/www/Software/systemd/NetworkTarget/#runningservicesafterthenetworkisup

charts/gardener-extension-provider-aws/templates/daemonset-custom-mtu.yaml
vpnachev: done

PR #53, opened 2020-03-17
mtu customizer should use default network interface instead of eth0

### LINKED ISSUE #52: MTU customizer unit fails if eth0 not available

danielfoehrKn:

See attached issue.

Fixes #52 

**Release note**:

```improvement operator

```

### DIFF
--- a/pkg/webhook/controlplane/ensurer.go
+++ b/pkg/webhook/controlplane/ensurer.go
@@ -17,12 +17,17 @@ package controlplane
 import (
 	"bytes"
 	"context"
+	"fmt"
 	"regexp"
 
-	"github.com/gardener/gardener-extension-provider-aws/pkg/aws"
+	"github.com/gardener/gardener-extensions/pkg/controller/operatingsystemconfig/oscommon/cloudinit"
+	"github.com/gardener/gardener-extensions/pkg/util"
 	extensionswebhook "github.com/gardener/gardener-extensions/pkg/webhook"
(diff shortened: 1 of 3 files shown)


vpnachev:
duplicate of #50 :smile:

rfranzke:
Thanks @danielfoehrKn for scouting and pro-actively providing a fix! Though, #50 was opened shortly before... :( Can we close this in favor of #50?

zanetworker:
Nice :) 

I can think of some problems though (which at the moment we don't have).  But imagine that we have VM with more than one network interface, your script fetches only the default and sets the MTU on it, the other one will still be left with the old MTU.

danielfoehrKn:
Sure, closing this.

PR #69, opened 2020-04-15
remove blocked-network cidr from kcm

mvladev:

Removes access to blocked networks from `kube-controller-manager`.

Fixes #

**Release note**:

```improvement operator
`kube-controller-manageer` no longer has access to blocked CIDRs.
```

### DIFF
--- a/pkg/webhook/controlplane/ensurer.go
+++ b/pkg/webhook/controlplane/ensurer.go
@@ -198,16 +198,17 @@ func ensureKubeSchedulerCommandLineArgs(c *corev1.Container, csiEnabled, csiMigr
 }
 
 func ensureKubeControllerManagerLabels(t *corev1.PodTemplateSpec, csiEnabled, csiMigrationComplete bool) {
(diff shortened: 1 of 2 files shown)


pkg/webhook/controlplane/ensurer.go
```diff
@@ -198,16 +198,17 @@ func ensureKubeSchedulerCommandLineArgs(c *corev1.Container, csiEnabled, csiMigr
 }
 
 func ensureKubeControllerManagerLabels(t *corev1.PodTemplateSpec, csiEnabled, csiMigrationComplete bool) {
+	// make sure to always remove this label
+	delete(t.Labels, v1beta1constants.LabelNetworkPolicyToBlockedCIDRs)
```
ialidzhikov, line 202: I think this explicit removal is not required. I tested with the change I'm suggesting below and during reconciliation this label is being removed from the kube-controller-manager Pod.

```suggestion
```

PR #72, opened 2020-04-23
Enforce workerConfig for io1 volumes

rfranzke:
If no `iops` are set when `io1` volumes are used then we should return an error in the validation of the `Shoot`, otherwise the `Worker` controller will fail, eventually, because MCM will complain:

```
E0423 11:13:31.849445       1 awsmachineclass.go:126] Validation of AWSMachineClass failed spec.blockDevices[0].ebs.iops: Required value: Please mention a valid ebs volume iops
```

/kind bug

**Release note**:

```improvement user
When worker pools with the `io1` volume type are used then the validation does now enforce that the `.spec.provider.workers[].providerConfig` does contain a proper AWS `WorkerConfig` (that states the iops).
```

### DIFF
--- a/pkg/apis/aws/validation/shoot.go
+++ b/pkg/apis/aws/validation/shoot.go
@@ -21,7 +21,6 @@ import (
 
 	"github.com/gardener/gardener/pkg/apis/core"
 	"github.com/gardener/gardener/pkg/apis/core/validation"
-
 	apivalidation "k8s.io/apimachinery/pkg/api/validation"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/validation/field"
@@ -52,6 +51,10 @@ func ValidateWorkers(workers []core.Worker, zones []apisaws.Zone, fldPath *field
 			allErrs = append(allErrs, field.Required(fldPath.Index(i).Child("volume"), "must not be nil"))
 		} else {
 			allErrs = append(allErrs, validateVolume(worker.Volume, fldPath.Index(i).Child("volume"))...)
+
+			if worker.Volume.Type != nil && *worker.Volume.Type == string(apisaws.VolumeTypeIO1) && worker.ProviderConfig == nil {
+				allErrs = append(allErrs, field.Required(fldPath.Index(i).Child("providerConfig"), fmt.Sprintf("WorkerConfig must be set if volume type is %s", apisaws.VolumeTypeIO1)))
+			}
 		}
 
 		if len(worker.Zones) == 0 {
--- a/pkg/apis/aws/validation/worker.go
+++ b/pkg/apis/aws/validation/worker.go
@@ -15,6 +15,8 @@
 package validation
 
 import (
+	"fmt"
+
 	apisaws "github.com/gardener/gardener-extension-provider-aws/pkg/apis/aws"
 
 	"k8s.io/apimachinery/pkg/util/validation/field"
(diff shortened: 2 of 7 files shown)


pkg/apis/aws/validation/shoot.go
```diff
@@ -52,6 +52,10 @@ func ValidateWorkers(workers []core.Worker, zones []apisaws.Zone, fldPath *field
 			allErrs = append(allErrs, field.Required(fldPath.Index(i).Child("volume"), "must not be nil"))
 		} else {
 			allErrs = append(allErrs, validateVolume(worker.Volume, fldPath.Index(i).Child("volume"))...)
+
+			if worker.Volume.Type != nil && *worker.Volume.Type == "io1" {
```
tim-ebert, line 56: Hmm, I think this will also put out an error, even if the `providerConfig` is set, or am I missing something?

tim-ebert: Also, I think we should improve the validation logic here, to put out an error, if `.volume.iops` is missing (in case `volumeType=io1` and not only if it is set but not in the valid range: https://github.com/gardener/gardener-extension-provider-aws/blob/f9b9f455c6af7b6e915f3f2de4a438cb4f663bdf/pkg/apis/aws/validation/worker.go#L27-L38

Otherwise the problem can still occur, right?

rfranzke: > Hmm, I think this will also put out an error, even if the providerConfig is set, or am I missing something?

Sure, indeed, thank you...

rfranzke: The second comment should be already covered by https://github.com/gardener/gardener-extension-provider-aws/blob/master/pkg/validator/shoot_validator.go#L65-L89

pkg/apis/aws/validation/shoot.go
```diff
@@ -52,6 +52,10 @@ func ValidateWorkers(workers []core.Worker, zones []apisaws.Zone, fldPath *field
 			allErrs = append(allErrs, field.Required(fldPath.Index(i).Child("volume"), "must not be nil"))
 		} else {
 			allErrs = append(allErrs, validateVolume(worker.Volume, fldPath.Index(i).Child("volume"))...)
+
+			if worker.Volume.Type != nil && *worker.Volume.Type == "io1" && worker.ProviderConfig == nil {
```
timuthy, line 56: Nit: Can you introduce a constant for `io1`?

pkg/apis/aws/validation/worker.go
```diff
@@ -24,15 +26,25 @@ import (
...
-		case "gp2":
-			if *workerConfig.Volume.IOPS < 100 || *workerConfig.Volume.IOPS > 10000 {
-				allErrs = append(allErrs, field.Forbidden(field.NewPath("volume", "iops"), "range is 100-10000 iops for gp2 volumes"))
-			}
-		case "io1":
-			if *workerConfig.Volume.IOPS < 100 || *workerConfig.Volume.IOPS > 20000 {
-				allErrs = append(allErrs, field.Forbidden(field.NewPath("volume", "iops"), "range is 100-20000 iops for io1 volumes"))
+	if volumeType != nil {
+		if *volumeType == string(apisaws.VolumeTypeIO1) && (workerConfig.Volume == nil || workerConfig.Volume.IOPS == nil) {
+			allErrs = append(allErrs, field.Required(field.NewPath("volume", "iops"), "iops must be provided when using io1 volumes"))
```
tim-ebert, line 31:
```suggestion
			allErrs = append(allErrs, field.Required(field.NewPath("volume", "iops"), fmt.Sprintf("iops must be provided when using %s volumes", apisaws.VolumeTypeIO1)))
```

pkg/apis/aws/validation/worker_test.go
```diff
@@ -75,5 +87,14 @@ var _ = Describe("ValidateWorkerConfig", func() {
 				"Field": Equal("volume.iops"),
 			}))))
 		})
+
+		It("should return an error if a non-supported volume type is specified", func() {
```
tim-ebert, line 91:
```suggestion
		It("should return an error if iops is set for a non-supported volume type", func() {
```

tim-ebert: Actually, do we have a check, that only supported volume types are specified in `workers[].volume.type`?

rfranzke: This check is done by Gardener's `ShootValidator` admission plugin.

tim-ebert: Sure, forgot that this is cross-checked with the CloudProfile...

tim-ebert:
```
> Generate Check
go generate needs to be run:
 M pkg/apis/aws/zz_generated.deepcopy.go
make: *** [Makefile:109: check-generate] Error 1
```

PR #93, opened 2020-05-14
improve documentation for custom vpc

vpnachev:
Add statement in the documentation that custom VPCs needs properly enabled DNS support.

Fixes #

**Release note**:

```improvement operator
NONE
```

### DIFF
--- a/docs/usage-as-end-user.md
+++ b/docs/usage-as-end-user.md
@@ -52,7 +52,7 @@ If the flag is not provided it is defaulted to `true`.
 The `networks.vpc` section describes whether you want to create the shoot cluster in an already existing VPC or whether to create a new one:
 
 * If `networks.vpc.id` is given then you have to specify the VPC ID of the existing VPC that was created by other means (manually, other tooling, ...).
-Please make sure that the VPC has attached an internet gateway - the AWS controller won't create one automatically for existing VPCs.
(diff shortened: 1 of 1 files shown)


docs/usage-as-end-user.md
```diff
@@ -52,7 +52,7 @@ If the flag is not provided it is defaulted to `true`.
 The `networks.vpc` section describes whether you want to create the shoot cluster in an already existing VPC or whether to create a new one:
 
 * If `networks.vpc.id` is given then you have to specify the VPC ID of the existing VPC that was created by other means (manually, other tooling, ...).
-Please make sure that the VPC has attached an internet gateway - the AWS controller won't create one automatically for existing VPCs.
+Please make sure that the VPC has attached an internet gateway - the AWS controller won't create one automatically for existing VPCs. To make sure the nodes are able to join and operate in your cluster properly, please make sure that your VPC has enabled [DNS Support](https://docs.aws.amazon.com/vpc/latest/userguide/vpc-dns.html), explicitly the attributes `enableDnsHostnames` and `enableDnsSupport` must be set to `true`.
```
rfranzke, line 55: Is this really necessary for all OS on AWS or just for CoreOS/JeOS?

/cc @timuthy

timuthy: CoreOS for instance also works w/o `enableDnsHostnames`. It's rather required for the node metrics and this again is depended on the metrics server setting (https://github.com/gardener/gardener/pull/2019). Imo, we can consider the statement above as a best practice.

rfranzke: OK,  thanks, makes sense.

PR #96, opened 2020-05-18
Allow configuring existing Elastic IP allocation id

### LINKED ISSUE #63: NatGateway integration
**Static Public IP addresses** for Gardener-based Shoot Cluster **Egress** (Outbound) Internet Connectivity.

Allow users to bring their own public static IP addresses and/or public static IP addresses ranges / prefixes which should be attached to the NatGateway.
The specific IP addresses can be re-assigned in case cluster crashed or misconfigured or deleted, etc.
Or, even move specific IP addresses between different clusters - the classic use case can be moving IP addresses from main cluster to backup cluster during disaster recovery (DR) procedure.

The feature is needed for IP addresses whitelisting by customers (the end-users, who work with the shoot cluster):
- In development / test / validation systems, both within enterprise network and outside, like in public regulated market cloud.
- Also in production for specific products, such as HaaS and HANA Cloud, which have dedicated Source IP address whitelisting feature.
In these cases the customer can whitelist the Shoot Cluster Egress IP addresses.

rfranzke:
This PR allows end-users to bring their own, pre-created elastic IPs for NAT gateways.

Fixes #63

**Release note**:

```noteworthy user
It is now possible to bring your own, pre-created Elastic IP and let the AWS extension associate it when creating NAT gateways. Please consult [this documentation](https://github.com/gardener/gardener-extension-provider-aws/blob/master/docs/usage-as-end-user.md#infrastructureconfig) to get more information.
```


example/30-infrastructure.yaml
```diff
@@ -47,13 +47,12 @@ spec:
...
-      # - s3
-      #   given the region above, this translates to service name com.amazonaws.eu-west-1.s3
+      # gatewayEndpoints: # custom service names for adding gateway VPC endpoints
+      # - s3 # given the region above, this translates to service name com.amazonaws.eu-west-1.s3
       zones:
       - name: eu-west-1a
         internal: 10.250.112.0/22
         public: 10.250.96.0/22
         workers: 10.250.0.0/19
+      # elasticIPAllocationID: eipalloc-123456 # Allocation ID of the Elastic IP taht will be attached to the NAT gateway in this zone
```
vpnachev, line 57:
```suggestion
      # elasticIPAllocationID: eipalloc-123456 # Allocation ID of the Elastic IP that will be attached to the NAT gateway in this zone
```

pkg/apis/aws/v1alpha1/types_infrastructure.go
```diff
@@ -57,16 +57,24 @@ type Networks struct {
...
-	// Public is the  public subnet range to create (used for bastion and load balancers).
+	// Public is the public subnet range to create (used for bastion and load balancers).
 	Public string `json:"public"`
-	// Workers is the  workers  subnet range  to create (used for the VMs).
+	// Workers is the workers subnet range to create (used for the VMs).
 	Workers string `json:"workers"`
+	// ElasticIPAllocationID contains the allocation ID of an Elastic IP that will be attached to the NAT gateway in
+	// this zone (e.g., `eipalloc-123456`). If it's not provided then a new Elastic IP will be automatically created
+	// and attached.
+	// Important: If this field is changed then the already attached Elastic IP will be disassociated from the NAT gateway
```
vpnachev, line 73:
```suggestion
	// Important: If this field is changed then the already attached Elastic IP will be dissociated from the NAT gateway
```
Please, add the same documentation to the internal version.

rfranzke:
> I did not get how the extension distinguish between created and provided Elastic IP, so that it will not delete the provided one?

If no elastic IP allocation ID is provided then it's the same behaviour like before: As part of the Terraform execution a new EIP is created and used. As part of the Terraform destruction this IP will be deleted again because it is managed by us.

If the allocation ID of an existing elastic IP is provided then it's the same behaviour like if the ID of  an existing VPC is provided: As part of the Terraform execution we use the EIP to attach it to the NAT gateway that we create. As part of the Terraform destruction this EIP will be detached but NOT deleted because it is not managed by us.

/cc @vpnachev

vpnachev:
Thank you for the explanation!
I have played with it and it working as expected.

/lgtm
/approve

PR #123, opened 2020-06-16
Add configuration option to enable encryption of StorageClasses

### LINKED ISSUE #122: Add option to enable encryption to StorageClassses

/area security
/kind enhancement
/priority normal
/platform aws

Add option to configure the `StorageClasses` with `enrypted: "true"` option

**Why is this needed**:

Security

mvladev:

/area security
/kind enhancement
/priority normal
/platform aws

Security is always better. It adds the following things:
- New `shootStorageClassConfig.encrypted` in `ControllerConfiguration` allows to configure the encryption parameter of Shoot `StorageClasses`.
- New etcd `StorageClass`'s encryption is controlled via `config.etcd.storage.encrypted` value in `gardener-extension-provider-aws` chart.
- New `shootStorageClassConfig.encrypted` in `ControlPlaneConfig` allows to configure the encryption parameter of Shoot `StorageClasses`. This value overrides the one set in the controller's `ControllerConfiguration`.

Fixes #122 

**Release note**:

```noteworthy user
New `shootStorageClassConfig.encrypted` in `ControlPlaneConfig` allows to configure the encryption parameter of Shoot `StorageClasses`. This value overrides the one set in the controller's `ControllerConfiguration`.
```

```noteworthy operator
New `shootStorageClassConfig.encrypted` in `ControllerConfiguration` allows to configure the encryption parameter of Shoot `StorageClasses`.
```

```noteworthy operator
New etcd `StorageClass`'s encryption is controlled via `config.etcd.storage.encrypted` value in `gardener-extension-provider-aws` chart.
```


vlerenc:
Thank you for taking this up. Please consider that we have to support encryption for all infrastructures, not only AWS.

rfranzke:
Can you please open issues in the respective provider extension repos to track the same?

mvladev:
Is `ControlPlaneConfig` the right place to put this user-overriding configuration? I don't think that the control plane settings and end-user `StorageClasses` should be in the same place.

rfranzke:
Currently, the `ControlPlane` controller is responsible for creating this `StorageClass`, that's why I suggested it. I don't think we have another place to move this responsibility today, and probably it doesn't make sense to invent one now just for deploying the `StorageClass`es.

mvladev:
@rfranzke I've added an additional field in `ControlPlaneConfig` to override the controller setting.

rfranzke:
Thanks Martin, will check asap.

mvladev:
Okay.. I just tested with existing cluster and - `The StorageClass "default" is invalid: parameters: Forbidden: updates to parameters are forbidden.`

pkg/apis/config/v1alpha1/defaults.go
```diff
@@ -21,3 +21,16 @@ import (
 func addDefaultingFuncs(scheme *runtime.Scheme) error {
 	return RegisterDefaults(scheme)
 }
+
+// SetDefaults_ControllerConfiguration sets default values for ControllerConfiguration objects.
+func SetDefaults_ControllerConfiguration(obj *ControllerConfiguration) {
```
vpnachev, line 26: Hm, this type is not served by API, it is used just to parse CRDs, so I am not sure that this defaulting will be applied.

charts/gardener-extension-provider-aws/templates/storageclass.yaml
```diff
@@ -9,4 +9,7 @@ provisioner: {{ .Values.config.etcd.storage.provisioner }}
 allowVolumeExpansion: true
 parameters:
   type: gp2
+{{- if .Values.config.etcd.storage.encrypted }}
+  encrypted: "true"
```
vpnachev, line 13: How about:
```suggestion
  encrypted: {{ .Values.config.etcd.storage.encrypted }}
```

pkg/apis/config/v1alpha1/defaults.go
rfranzke: I think we have to explicitly call it, similar to https://github.com/gardener/gardener/blob/master/cmd/gardener-controller-manager/app/gardener_controller_manager.go#L129-L143, but I can't find it in https://github.com/gardener/gardener-extension-provider-aws/blob/master/pkg/cmd/config.go#L28:6.

charts/gardener-extension-provider-aws/templates/configmap.yaml
```diff
@@ -24,3 +24,7 @@ data:
 {{- if .Values.config.etcd.backup }}
 {{ toYaml .Values.config.etcd.backup | indent 6 }}
 {{- end }}
+{{- if .Values.config.shootStorageClassConfig.encrypted }}
+    shootStorageClassConfig:
+      encrypted: true
```
rfranzke, line 29:
```suggestion
{{- if .Values.config.shootStorageClassConfig }}
    shootStorageClassConfig:
      encrypted: {{ .Values.config.shootStorageClassConfig.encrypted }}
```

pkg/controller/controlplane/valuesprovider.go
```diff
@@ -302,17 +302,19 @@ var (
 )
 
 // NewValuesProvider creates a new ValuesProvider for the generic actuator.
-func NewValuesProvider(logger logr.Logger) genericactuator.ValuesProvider {
+func NewValuesProvider(logger logr.Logger, storgeClass *config.StorageClass) genericactuator.ValuesProvider {
```
rfranzke, line 305:
```suggestion
func NewValuesProvider(logger logr.Logger, storageClass *config.StorageClass) genericactuator.ValuesProvider {
```

pkg/controller/controlplane/valuesprovider.go
```diff
@@ -365,16 +367,35 @@ func (vp *valuesProvider) GetControlPlaneShootChartValues(
...
+
+	encrypt := false
+
+	// User-provided encryption overrides the one set in the controller.
+	if cpConfig.ShootStorageClassConfig != nil && cpConfig.ShootStorageClassConfig.Encrypted != nil {
+		encrypt = *cpConfig.ShootStorageClassConfig.Encrypted
+	} else if vp.storageClass != nil && vp.storageClass.Encrypted != nil {
+		encrypt = *vp.storageClass.Encrypted
+	}
+
```
rfranzke, line 395: I think somewhere here you would have to delete the existing `StorageClass` as it's immutable.

rfranzke: Ref https://github.com/gardener/gardener-extension-provider-openstack/pull/102

rfranzke:
@mvladev please update the PR
/needs rebase

phil9909:
> Okay.. I just tested with existing cluster and -
> 
> ```
> The StorageClass "default" is invalid: parameters: Forbidden: updates to parameters are forbidden.
> ```

You could always create two `StorageClass` objects (e.g. `default` and `default-encrypted`) and assign the `storageclass.kubernetes.io/is-default-class` annotation depending on the config. According to https://kubernetes.io/docs/tasks/administer-cluster/change-default-storage-class/ that should be possible.

The new default will then only effect newly created volumes.

afritzler:
I would also suggest to use two storage classes instead of overwriting the default one. Leads to less pain when upgrading an existing cluster.

ialidzhikov:
Hold until v1.11 of provider-aws is released.

/hold

vlerenc:
⚠️ @timuthy informed me that GRM now supports `resources.gardener.cloud/delete-on-invalid-update=true` that will allow to replace an existing storage class. If I understood him right, we need to add that to the storage classes here, but we discussed that it would be good to repeat the exercise also for all other storage classes in the other providers, so that we can formulate a roll out e-mail roughly saying the following:

> "Because of SGS security rule bla, we need to update the storage classes Gardener creates for your clusters. From now onwards they will be reconciled. So, any changes you might have done to the Gardener-managed storage classes (by replacing them) will be reverted with the next deployment/reconciliation and from then onward. If you need other storage classes, please create them under different names. Please note that we also annotate the default storage class as such. Should you have annotated another storage class as default (and removed the default annotation for the Gardener-managed storage class), you will have no default anymore, because defaults apply only if exactly one storage class is annotated as such."

charts/gardener-extension-provider-aws/templates/storageclass.yaml
mvladev: The default value is `false` so it should not update the resource if not needed.

charts/gardener-extension-provider-aws/templates/configmap.yaml
mvladev: The default value is `false` so we should not update the resource if not needed.

pkg/apis/config/v1alpha1/defaults.go
mvladev: I've updated the PR - now decoding happens via the `UniversalDecoder().Decode(...)`  method of the codex factory. Defaulting happens automatically. See `pkg/apis/config/loader/loader_test.go` - there is a test which tests exactly this behavior.

charts/internal/shoot-storageclasses/values.yaml
```diff
@@ -1 +1,2 @@
 useLegacyProvisioner: true
+encrypt: false
```
timuthy, line 2:
```suggestion
encrypted: false
```

to be consistent with the other value names

pkg/apis/config/loader/loader.go
```diff
@@ -15,33 +15,25 @@
...
 	install.Install(Scheme)
-	yamlSerializer := json.NewYAMLSerializer(json.DefaultMetaFactory, Scheme, Scheme)
-	Codec = versioning.NewDefaultingCodecForScheme(
-		Scheme,
-		yamlSerializer,
-		yamlSerializer,
-		schema.GroupVersion{Version: "v1alpha1"},
-		runtime.InternalGroupVersioner,
-	)
+	Codecs = serializer.NewCodecFactory(Scheme, serializer.EnableStrict)
```
timuthy, line 36: Any special reason we need the strict mode here? I've never done this in any Gardener code before. Did we miss something?

pkg/apis/config/loader/loader.go
```diff
@@ -57,16 +49,14 @@ func LoadFromFile(filename string) (*config.ControllerConfiguration, error) {
 // Load takes a byte slice and de-serializes the contents into ControllerConfiguration object.
 // Encapsulates de-serialization without assuming the source is a file.
 func Load(data []byte) (*config.ControllerConfiguration, error) {
-	cfg := &config.ControllerConfiguration{}
```
timuthy, line 60: We already have some logic which generically takes care about decoding probably, see this [example](https://github.com/gardener/gardener-extension-provider-aws/blob/a7ae362d82c1652e70ccfebb856f291b1582e9a3/pkg/validator/serialization.go).

rfranzke:
What's the status of this PR?

rfranzke:
Thanks @jia-jerry, yes, it is immutable. As the `StorageClass` is managed as part of a `ManagedResource` managed by by our [gardener/gardener-resource-manager](https://github.com/gardener/gardener-resource-manager) we can leverage one of its features, see https://github.com/gardener/gardener-resource-manager/pull/69.
@vpnachev wants to add this annotation to all `StorageClass`es for all provider extensions. This will also help to reconcile our desired `StorageClass` in case a user deletes and recreates it with his desired settings (which would otherwise result in a failing `SystemComponents` health check in the `Shoot` as GRM is not able to update it (due to immutability) without this annotation).

phil9909:
I think updating the `StorageClass` is misleading.

When I do a `kubectl get storageclasses.storage.k8s.io default -o yaml` after turning on encryption, I should eventually see something like this: 
```yaml
allowVolumeExpansion: true
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  annotations:
    resources.gardener.cloud/description: |-
      DO NOT EDIT - This resource is managed by gardener-resource-manager.
      Any modifications are discarded and the resource is returned to the original state.
    storageclass.kubernetes.io/is-default-class: "true"
  creationTimestamp: "2020-08-10T09:14:26Z"
  labels:
    shoot.gardener.cloud/no-cleanup: "true"
  name: default
  resourceVersion: "188"
  selfLink: /apis/storage.k8s.io/v1/storageclasses/default
  uid: 732a8425-ff41-41e4-b28f-439ab49cecb9
parameters:
  type: gp2
  encrypted: "true"
provisioner: kubernetes.io/aws-ebs
reclaimPolicy: Delete
volumeBindingMode: Immediate
```

So I will assume every `PersistentVolume` that uses the `StorageClass` `default` is encrypted
```
$ kubectl get PersistentVolume
NAME                                       CAPACITY   ACCESS MODES   RECLAIM POLICY   STATUS   CLAIM                                                                                         STORAGECLASS   REASON   AGE
pvc-0770f149-201a-4608-9e93-4bd487749eca   20Gi       RWO            Delete           Bound    sf-2f571956-8fa3-47c8-922b-becc09c821d2/pgdata-sapcp-2f571956-8fa3-47c8-922b-becc09c821d2-0   default                 2d15h
pvc-0a00dca0-e8ba-4ba8-8bdb-a7b9f552ed07   20Gi       RWO            Delete           Bound    sf-2769b11d-2c6a-477e-85ae-6ee90b0ada01/pgdata-sapcp-2769b11d-2c6a-477e-85ae-6ee90b0ada01-1   default                 2d15h
pvc-11a22e7c-bec4-44c6-a5cd-a20bb4bd4807   20Gi       RWO            Delete           Bound    sf-36cea663-505d-43ef-b794-5a55b5b2dc47/pgdata-sapcp-36cea663-505d-43ef-b794-5a55b5b2dc47-1   default                 2d18h
pvc-1fd8ee33-6222-4486-ac04-1769904f058b   20Gi       RWO            Delete           Bound    sf-19714102-c1a2-4ac7-83ab-956e6de9471f/pgdata-sapcp-19714102-c1a2-4ac7-83ab-956e6de9471f-1   default                 2d15h
pvc-445fc79c-49c3-4567-b22e-7768d09f24ec   20Gi       RWO            Delete           Bound    sf-85a37de4-ca31-4db1-96af-4d2fb440c33f/pgdata-sapcp-85a37de4-ca31-4db1-96af-4d2fb440c33f-0   default                 2d15h
pvc-44a021b9-15be-43db-84eb-6b0ccb14b964   20Gi       RWO            Delete           Bound    sf-bccbd64d-fc66-4cad-939a-1df4799cded7/pgdata-sapcp-bccbd64d-fc66-4cad-939a-1df4799cded7-1   default                 2d14h
pvc-48bec0e3-c0e8-47fc-935e-7518173907c9   20Gi       RWO            Delete           Bound    sf-2f571956-8fa3-47c8-922b-becc09c821d2/pgdata-sapcp-2f571956-8fa3-47c8-922b-becc09c821d2-1   default                 2d15h
pvc-4c1a08a1-50e9-40e4-87ed-e6b84f183459   20Gi       RWO            Delete           Bound    sf-7b0d9f6f-3c8d-4b83-bc36-1ce2c0f97634/pgdata-sapcp-7b0d9f6f-3c8d-4b83-bc36-1ce2c0f97634-0   default                 2d15h
```

But this is not true (only PVs created after the update will be encrypted, existing PVs won't be updated).

I even consider this a bug in k8s. It should either prevent `StorageClasses` which are still in use from being deleted or at least block recreating it.

mvladev:
Okay I'll update the PR with the following changes:

- Introduce `default-encrypted` `StorageClass`.
- Existing `StorageClasses` won't be modified.
- Add a new field that would allow end-users to override the `storageclass.kubernetes.io/is-default-class: "true"` annotation.

WDYT?

rfranzke:
Sounds good @mvladev

mvladev:
I'm just wondering if we should add this `disable settting default class` field to the `Shoot` resource. It's a generic setting that can be applied across providers.

rfranzke:
Hmm, not sure, so far we haven't added fields to the `Shoot` directly whose implementation/support directly depends on the extension controllers. @timebertt @ialidzhikov @vpnachev opinions?

timebertt:
> so far we haven't added fields to the `Shoot` directly whose implementation/support directly depends on the extension controllers

I would probably keep it this way, but no strong opinion actually

vpnachev:
Sounds good to me. Obviously not all shoot owners like the default class managed by Gardener, so we should allow them to provide their own default class while we still manage some storage classes inside the shoot.

rfranzke:
@mvladev do you still plan to work in this PR?

mvladev:
I'll work on it next week.

rfranzke:
@mvladev any further plans with this PR?

mvladev:
I've the updated version  with not settled ng the default class ready. I'll update the PR after new year.

mvladev:
This PR is reworked and continued ib #247 

/close

PR #124, opened 2020-06-16
Wait for terraformer before deleting infrastructure

### LINKED ISSUE #121: Infrastructure controller can wrongly early exit on deletion

/area quality
/kind bug
/priority normal
/platform aws

**What happened**:
Currently the infrastructure deletion can wrongly exit on detecting empty state and assuming that nothing was creating (because of invalid credentials or other reason).

1. Create an Infrastructure with terraform config that will timeout out to be created.

2. Delete the Infrastructure.

3. Ensure that the terraformer Pod from the creation can still run but the infrastructure controller can proceed with the deletion detecting empty terraform state config map and removing the Infrastucture finalizer.

Generally this can happen as the terraform pkg is only deleting the Pod and does not wait until the deletion happens after the grace period of time. The infra deletion can meanwhile wrongly complete when the state is empty but there is a terminating terraformer Pod.

**Environment**:

- Gardener version (if relevant): v1.5.2
- Extension version: v1.8.2
- Others:

tim-ebert:

/area quality
/area robustness
/kind bug
/priority normal
/platform aws

With this PR, the infrastructure actuator deletes all terraformer pods from previous reconciliations and waits for them to be deleted before checking the state cm for an empty state.
This is to ensure, that a terraformer pod, which creates the infrastructure for a freshly created cluster, is deleted / has finished, so we don't falsely exit early because the state configmap is still empty (and therefore don't delete the infrastructure at all).

Fixes #121 

✅ Depends on https://github.com/gardener/gardener/pull/2461, once it is merged, I will vendor ~~g/g@master~~ g/g@v1.6.2 and mark this PR as ready to review.

The issue probably also applies to other providers, so I will create similar PRs there, once this one is agreed on and merged.

**Release note**:

```improvement user
A bug has been fixed, that caused the `Infrastructure` not to be deleted for newly created clusters.
```


pkg/controller/infrastructure/actuator_delete.go
```diff
@@ -51,10 +51,14 @@ func Delete(
 		return fmt.Errorf("could not create the Terraformer: %+v", err)
 	}
 
+	// terraform pod from previous reconciliation might still be running, ensure they are gone before doing any operations
+	if err := tf.EnsureCleanedUp(ctx); err != nil {
```
tim-ebert, line 55: I am wondering, if we should rather use `WaitForCleanEnvironment` here to not interrupt any running `tf-destroy` pods if the deletion is triggered again (backoff on provider-aws side / timeout and retry on gardenlet side / ...)
The benefit of `EnsureCleanup` would be, that running `tf-apply` pods would be stopped directly instead of waiting for high timeouts.

rfranzke: Hmm, good question. My personal gud feeling would be that `WaitForCleanEnvironment` is a better starting point. Yes, we might potentially run into the timeout, but I cannot overlook the side effects that the termination of a running `destroy` pod could have.
Generally, PR lgtm, thanks.

tim-ebert: Ok, thanks for your opinion. Then let's go for `WaitForCleanEnvironment` 👍

tim-ebert:
https://github.com/gardener/gardener/pull/2461 is merged and released, vendored g/g@v1.6.2.
/ready

PR #133, opened 2020-06-26
Adapt the extension to Loki

Kristian-ZH:

/area logging
/priority normal
/platform aws

Adapt the extension to Loki

This PR depends on this one: gardener/gardener#2515 and should wait until it is merged

Fixes #

**Release note**:

```noteworthy operator
The logging configuration has been adapted to cater with Loki which was introduced as part of [gardener/gardener#2515](https://github.com/gardener/gardener/pull/2515). If an older version of this extension is used with a Gardener version >= v1.8 and the `Logging` feature gate is enabled then the fluent-bit matcher won't catch the log anymore (resulting in the logs being collected, but not nicely parsed and instead displayed in nested JSON). It is recommended update to at least this extension version when Gardener >= v1.8 is used and the `Logging` feature gate is enabled.
```


rfranzke:
Thanks. But what if somebody uses a newer Gardener version (with Loki) but an older AWS extension version (without this PR)?

Kristian-ZH:
> Thanks. But what if somebody uses a newer Gardener version (with Loki) but an older AWS extension version (without this PR)?

The log message will not be parsed and will be showed in Grafana like:
```
"log" {
   .................... // This is string containing everything(log, severity, process-id, timestamp)
}
```

rfranzke:
Hm, okay, I think this is okay, WDYT? The `Logging` feature is still alpha, so we can ask people to upgrade if they want to see the correctly parsed logs in Grafana and if they are already using Loki.

Issue #145, opened 2020-07-22
Add Default VolumeSnapshotClass 


runzexia:
ref https://github.com/gardener/gardener-extension-provider-gcp/issues/91

```
apiVersion: snapshot.storage.k8s.io/v1beta1
kind: VolumeSnapshotClass
metadata:
  name: default
  annotations:
    snapshot.storage.kubernetes.io/is-default-class: “true”
driver: ebs.csi.aws.com
deletionPolicy: Delete
```
/assign

PR #146, opened 2020-07-22
create default VolumeSnapshotClass if kubernetes version >= 1.18

runzexia:
Signed-off-by: RyderXia <ryder.xia@sap.com>

/area storage
/kind enhancement
/priority normal
/platform aws

Fixes https://github.com/gardener/gardener-extension-provider-aws/issues/145

**Release note**:

```improvement user
create default VolumeSnapshotClass if kubernetes version >= 1.18
```

### DIFF
--- a/charts/internal/shoot-storageclasses/templates/storageclasses.yaml
+++ b/charts/internal/shoot-storageclasses/templates/storageclasses.yaml
@@ -18,4 +18,13 @@ metadata:
 allowVolumeExpansion: true
 provisioner: ebs.csi.aws.com
 volumeBindingMode: WaitForFirstConsumer
+---
+apiVersion: snapshot.storage.k8s.io/v1beta1
+kind: VolumeSnapshotClass
+metadata:
+  name: default
+  annotations:
+    snapshot.storage.kubernetes.io/is-default-class: “true”
+driver: ebs.csi.aws.com
+deletionPolicy: Delete
 {{- end }}


charts/internal/shoot-storageclasses/templates/storageclasses.yaml
```diff
@@ -18,4 +18,15 @@ metadata:
 allowVolumeExpansion: true
 provisioner: ebs.csi.aws.com
 volumeBindingMode: WaitForFirstConsumer
+  {{- if semverCompare ">= 1.18" .Values.kubernetesVersion  }}
```
rfranzke, line 21: You don't need this `if` condition - this file contains the CSI storage classes, and when CSI is used the CSI snapshot components are always deployed as well. Hence, we can also always deploy the `VolumeSnapshotClass`.
The check for the minimum required Kubernetes version (1.18) is already done in the Golang code itself, so no need to take care here again. :)

runzexia: cool Golang code 👍

charts/internal/shoot-storageclasses/templates/storageclasses.yaml
```diff
@@ -18,4 +18,13 @@ metadata:
 allowVolumeExpansion: true
 provisioner: ebs.csi.aws.com
 volumeBindingMode: WaitForFirstConsumer
+---
+apiVersion: snapshot.storage.k8s.io/v1beta1
+kind: VolumeSnapshotClass
+metadata:
+  name: default
```
rfranzke, line 25: What about?

```suggestion
  name: gardener-default
```

runzexia: done

tim-ebert: Why `gardener-default`? Is `default` somehow reserved? Can't we keep it similar to the `default` StorageClass?

rfranzke: I'm fine with both. Many examples/guides call their storage classes "default" which conflicts with Gardener's default storage class. But there's no "hard" reason and just this soft criteria that came to my mind. If you prefer `default` over `gardener-default` then go ahead, I'm fine.

tim-ebert: Ok, then I would indeed go with `default` as the name for the `VolumeSnapshotClass` to make it consistent with the `StorageClass` and also other examples :)

charts/internal/shoot-storageclasses/templates/storageclasses.yaml
```diff
@@ -18,4 +18,13 @@ metadata:
 allowVolumeExpansion: true
 provisioner: ebs.csi.aws.com
 volumeBindingMode: WaitForFirstConsumer
+---
+apiVersion: snapshot.storage.k8s.io/v1beta1
+kind: VolumeSnapshotClass
+metadata:
+  name: gardener-default
```
tim-ebert, line 25: Sorry for the back and forth, see https://github.com/gardener/gardener-extension-provider-aws/pull/146#discussion_r459248601
```suggestion
  name: default
```

charts/internal/shoot-storageclasses/templates/storageclasses.yaml
```diff
@@ -18,4 +18,13 @@ metadata:
...
+  annotations:
+    snapshot.storage.kubernetes.io/is-default-class: “true”
```
tim-ebert, line 27: Should we also add the `resources.gardener.cloud/delete-on-invalid-update=true` annotation? (ref https://github.com/gardener/gardener-resource-manager/pull/69)
/cc @rfranzke

rfranzke: The `VolumeSnapshotClass` is a CRD, but we don't install a validating webhook for it. Maybe one of the CSI snapshot components creates it on its own.
But yes, we can simply add the annotation as well due to the similarly with normal storage classes.

charts/internal/shoot-storageclasses/templates/storageclasses.yaml
rfranzke: Sorry for the back and forth @runzexia :)

charts/internal/shoot-storageclasses/templates/storageclasses.yaml
tim-ebert: Ah. I just thought about the similarity to the storage classes and thought that `VolumeSnapshotClass`es might also be immutable. But I missed, that this is a CR...
Well, then we don't have to add it, I guess...

charts/internal/shoot-storageclasses/templates/storageclasses.yaml
runzexia: it's ok for me😊

PR #147, opened 2020-07-23
Enable encryption by default AWS instance EBS vols

### LINKED ISSUE #126: Enable encryption on disks attached to AWS VMs

/area security
/kind enhancement
/priority normal
/platform aws

Encryption to be enabled by default for all disks attached to EC2 instances

**Why is this needed**:
Security

prashanth26:

/area security
/kind enhancement
/priority normal
/platform aws

The PR enables encryption by default for EBS volumes used by AWS EC2 instances. 

Fixes #126 

This might cause issues with machines that do not support encryption by default. Refer list here - https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/EBSEncryption.html#ebs-encryption-requirements. 

However, this is something that can be overridden by setting the encryption flag to `false` on the [machine volume settings here](https://github.com/gardener/gardener/blob/master/example/90-shoot.yaml#L49). 

I checked with our cloud profile and I found 6 machine types with `g4dn` which don't have support for encryption. Something that we need to disable from the cloud profile?

**Release note**:

```noteworthy user
Enable encryption by default for AWS instance EBS volumes.
```
```action operator
Need to remove EC2 instances without encryption support in the cloud profile.
```

### DIFF
--- a/docs/usage-as-end-user.md
+++ b/docs/usage-as-end-user.md
@@ -121,9 +121,12 @@ If you don't want to configure anything for the `cloudControllerManager` simply 
 
 ## `WorkerConfig`
 
-The AWS extension supports encryption for volumes plus support for additional data volumes per machine.
-By default (if not stated otherwise), all the disks are unencrypted.
+The AWS extension supports encryption for volumes plus support for additional data volumes per machine. 
 For each data volume, you have to specify a name.
+By default (if not stated otherwise), all the disks (root & data volumes) are encrypted. 
+Please make sure that your [instance-type supports encryption](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/EBSEncryption.html). 
+If your instance-type doesn't support encryption, you will have to disable encryption (which is enabled by default) by setting `volume.encrpyted` to `false` (refer below shown YAML snippet).
+
 The following YAML is a snippet of a `Shoot` resource:
 
 ```yaml
@@ -135,7 +138,7 @@ spec:
       volume:
         type: gp2
         size: 20Gi
-        encrypted: true
+        encrypted: false
       dataVolumes:
       - name: kubelet-dir
         type: gp2
--- a/pkg/controller/worker/machines.go
+++ b/pkg/controller/worker/machines.go
@@ -274,7 +274,7 @@ func computeEBS(size string, volumeType *string, encrypted *bool) (map[string]in
 
 	ebs := map[string]interface{}{
 		"volumeSize":          volumeSize,
-		"encrypted":           false,
+		"encrypted":           true,
 		"deleteOnTermination": true,
 	}
 
--- a/pkg/controller/worker/machines_test.go
+++ b/pkg/controller/worker/machines_test.go
@@ -427,7 +427,7 @@ var _ = Describe("Machines", func() {
 									"volumeSize":          volumeSize,
 									"volumeType":          volumeType,
 									"deleteOnTermination": true,
-									"encrypted":           false,
+									"encrypted":           true,
 								},
 							},
 						},


timuthy:
> I checked with our cloud profile and I found 6 machine types with g4dn which don't have support for encryption. Something that we need to disable from the cloud profile?

The encrypted volumes feature was once [added](https://github.com/gardener/gardener/commit/6a9f80379ab1152b167d0d58914d1b42a7e48803) to the shoot API and if we want to exclude certain machine types, we'll need an analogue implementation for the CloudProfile + validation.

hardikdr:
Changes via /lgtm.
Though need to follow up on if it breaks existing clusters - eg unsupported machine types.

prashanth26:
> The encrypted volumes feature was once added to the shoot API and if we want to exclude certain machine types, we'll need an analogue implementation for the CloudProfile + validation.

Or we could retain the unsupported machines in the cloud profile as it is today. 

However, we could send out announcements mentioning shoot owners who would like to use machines without manual encryption support will have to update [their worker encryption flags] to false (https://github.com/gardener/gardener/blob/master/example/90-shoot.yaml#L49).

timuthy:
> However, we could send out announcements mentioning shoot owners who would like to use machines without manual encryption support will have to update their worker encryption flags to false.

I'd rather go the cloud profile way to be sure. You can never reach any stakeholder by such an announcements, imo.

rfranzke:
/assign @timuthy 
I get your point @timuthy, but given that not too many machine types would be affected I'd probably not block this PR and create an issue in g/g instead. For now, we can probably document that encryption is enabled by default and certain machine types don't support it (why it must be set to `false`).

prashanth26:
> Please adapt https://github.com/gardener/gardener-extension-provider-aws/blob/master/docs/usage-as-end-user.md#workerconfig

Sorry missed that. I have updated the docs to reflect the latest state. PTAL.

rfranzke:
Can you check the failing test?

prashanth26:
I have fixed the tests.

vpnachev:
@prashanth26 could you rephrase the second release note - I do not get it.
> Need to remove EC2 instances without encryption support in the cloud profile.

prashanth26:
Hi @vpnachev ,

I have rewritten the release note. Hope it's better now.

vpnachev:
Thanks! It is looking great now.

vlerenc:
FYI: AWS reported that issues with encrypted disks can happen regardless of the disk size, which was new to me (reported were issues only with large disks: https://github.com/kubernetes/kubernetes/issues/86064). A call is scheduled for 21:00 Thursday (German time) to clarify that and more (like KCM/CSI handling of that case). Should we hold back the release of this component, at least until Friday or should we "risk" it @vpnachev @timuthy @rfranzke? On the other hand, different than KCM/CSI handled volume handling, we can in MCM handle the local disks always differently and not fail, right @prashanth26?

prashanth26:
> On the other hand, different than KCM/CSI handled volume handling, we can in MCM handle the local disks always different and not fail, right @prashanth26?

The system disks (root & additional attached volumes) are encrypted by default using this PR. And this is agnostic of Kubernetes, as we are just attaching encrypted system disks now. I tried it out with a simple cluster, seems to work well. However, I am not sure about the size implication here. 

However, this is something that can be overridden by setting the encryption flag to `false` on the [machine volume settings here](https://github.com/gardener/gardener/blob/master/example/90-shoot.yaml#L49) in the worst case.

vpnachev:
To be honest, we had not to change the default behavior at all. Encrypted volumes were available even before this PR.
But now it is in and In case of emergency, the encryption can be explicitly disabled by the shoot cluster owner, so I do not think this should be a blocker for the next release.

rfranzke:
+1 @vpnachev

PR #154, opened 2020-08-04
Vendor gardener/gardener@v1.8.0

rfranzke:

/area networking
/kind bug
/priority normal
/platform aws

With this PR we reconcile all shoot webhook configurations to make sure potential port configuration changes are reflected correctly in the system.
Also, we explicitly specify the webhook service port as it's now different than the server port (443 vs. 10250), and the extensions library needs to be aware to properly register the webhook configuration with the right port.

✅ Depends on gardener/gardener#2663 and gardener/gardener#2665
/cc @stoyanr @timebertt @vpnachev 

**Release note**:

```improvement operator
NONE
```


cmd/gardener-extension-provider-aws/app/app.go
```diff
@@ -184,19 +184,17 @@ func NewControllerManagerCommand(ctx context.Context) *cobra.Command {
...
-				controllercmd.LogErrAndExit(err, "Could not add controllers to manager")
-			}
-
-			// Cleanup leaked ClusterRoles from worker controller.
-			// See https://github.com/gardener-attic/gardener-extensions/pull/378/files and https://github.com/gardener/gardener/issues/2144.
-			// TODO: This code can be removed in a future version.
+			// Update shoot webhook configuration in case the webhook server port has changed.
 			c, err := client.New(restOpts.Completed().Config, client.Options{})
 			if err != nil {
 				controllercmd.LogErrAndExit(err, "Error creating client for orphaned ClusterRole cleanup")
```
stoyanr, line 190: Since `CleanupLeakedClusterRoles` has been deleted (I hope on purpose), the error message here is wrong
```suggestion
				controllercmd.LogErrAndExit(err, "Error creating client for startup tasks")
```

Issue #156, opened 2020-08-08
network.vpc.gatewayEndpoints are not created when using existing VPC


ialidzhikov:
**How to categorize this issue?**

/area networking
/kind bug
/priority normal
/platform aws

**What happened**:
`network.vpc.gatewayEndpoints` are now usable only when new VPC is requested (`networks.vpc.id`)

**How to reproduce it (as minimally and precisely as possible)**:
1. Create an Infrastructure with `networks.vpc.id` and `networks.vpc.gatewayEndpoints`:

```yaml
apiVersion: aws.provider.extensions.gardener.cloud/v1alpha1
kind: InfrastructureConfig
networks:
  vpc:
    id: vpc-123456
    gatewayEndpoints:
    - s3
  zones:
  - name: eu-west-1a
    internal: 10.250.112.0/22
    public: 10.250.96.0/22
    workers: 10.250.0.0/19
```

2. Ensure that the corresponding VPC endpoint resources won't be created.

**Anything else we need to know?**:

**Environment**:

- Gardener version (if relevant):
- Extension version: v1.12.0
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
- Others:

ialidzhikov:
@rfranzke , just to make sure - the gateway endpoint (`aws_vpc_endpoint`) semantically is not related with only new VPCs and it is expected to have it created also when existing VPC is used, right?

rfranzke:
@vpnachev nice catch, thanks for opening the issue.
Yes, it is expected/desired behaviour.

PR #157, opened 2020-08-09
Add integration test for existing vpc case

### LINKED ISSUE #155: Rework the infrastructure integration test with envtest pkg

/area testing
/kind enhancement
/priority normal
/platform aws

**Why is this needed**:
Currently the infrastructure integration test directly interacts with the actuator's `Reconcile` and `Delete` funcs which leaves some points uncovered/untested - controller-runtime predicates and controller-runtime inner machinery that afterwards results in `Reconcile` and `Delete` invocations of the actuator. 
Looks like https://godoc.org/sigs.k8s.io/controller-runtime/pkg/envtest package has a good entry point for integration tests. It allows tests against existing cluster or against partial "ControlPlane" (apiserver and etcd started by the pkg, which might be enough for some cases). We could obtain the environment configuration, create a manager from it, add the controller we want to test to this manager (we already have the `AddToManager` funcs), start the manager and then finally start the test.
I think this will improve the tests as the test will be concentrated more on contract testing (applying and deleting CRs and checking its status) rather than controller/actuator inner working (for example instantiate a chartrenderer and pass it to the `Reconcile` func).

ialidzhikov:

/area quality
/kind test
/priority normal
/platform aws

This PR:
- reworks the existing infrastructure test to use `sigs.k8s.io/controller-runtime/pkg/envtest` (more details on https://github.com/gardener/gardener-extension-provider-aws/issues/155)
- adds integration test for infrastructure using existing VPC
- adds integration test for https://github.com/gardener/gardener-extension-provider-aws/pull/161

Fixes #155

Depends on:
- [x] https://github.com/gardener/gardener-extension-provider-aws/issues/156

**Release note**:

```improvement operator
NONE
```


ialidzhikov:
@schrodit , @dguendisch as there is a small change in the `TestDefinition`, if I trigger the test execution now which TestDefinition will be considered - the one in the master or the one from this PR?

dguendisch:
The TestDefinition should be picked up from this PR.

test/integration/infrastructure/infrastructure_test.go
```diff
@@ -82,193 +80,519 @@ func validateFlags() {
...
-	})
+	By("create namespace for test execution")
+	namespace = &corev1.Namespace{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: namespaceName,
+		},
+	}
+	if err := c.Create(ctx, namespace); err != nil {
+		return err
+	}
```
rfranzke, line 285:
```suggestion
	Expect(c.Create(ctx, namespace)).To(Succeed())
```

test/integration/infrastructure/infrastructure_test.go
```diff
@@ -82,193 +80,519 @@ func validateFlags() {
...
-			secretName = "cloudprovider"
+	By("create cluster")
+	cluster = &extensionsv1alpha1.Cluster{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: namespaceName,
+		},
+	}
+	if err := c.Create(ctx, cluster); err != nil {
+		return err
+	}
```
rfranzke, line 295:
```suggestion
	Expect(c.Create(ctx, cluster)).To(Succeed())
```

rfranzke: etc., you can follow the same for the rest of this function and similar functions

ialidzhikov:
Hmm, it actually failed with:
```
Message: "customresourcedefinitions.apiextensions.k8s.io \"clusters.extensions.gardener.cloud\" is forbidden: User \"system:serviceaccount:default:default\" cannot get resource \"customresourcedefinitions\" in API group \"apiextensions.k8s.io\" at the cluster scope",
```

@schrodit , @dguendisch  is it possible to adapt the RBAC so the test can install the required CRDs into the target cluster?

schrodit:
> is it possible to adapt the RBAC so the test can install the required CRDs into the target cluster?

yes it is possible. @dguendisch lets discuss some possible options here. 
We can as quick fix add a role/rolebinding that allows the creation of crds for the default user

test/aws/client/client.go
```diff
@@ -0,0 +1,62 @@
+// Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```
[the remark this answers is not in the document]
ialidzhikov: Now added also the needed IAM service to our client under `pkg/aws/client` and now the integration test is using it.

test/integration/infrastructure/infrastructure_test.go
ialidzhikov: Wrt to the end result, there is no change as the `runTest` func returns err and it is being checked in the test itself. I had to adapt the `runTest` func to be able to return err as we now have a test case that expects err to have occurred. And I didn't want to mix 
```
Expect(err).NotTo(HaveOccurred)
```

statements with err propagation

```
if err != nil {
    return err
} 
```

in the same func and that's why I decided to be consistent in the func and use only error propagation.

Let me know if that makes sense to you.

rfranzke: Yes, sure, it's fine with me. My point was also only consistency because I saw https://github.com/gardener/gardener-extension-provider-aws/pull/157/files#diff-10becb4271192c12b5fa414024464971R273-R274, but those are actually statements for the `CleanupAction` which I didn't see earlier.

Fine with me, all good. :)

PR #161, opened 2020-08-10
Infrastructure controller validates VPC attributes

### LINKED ISSUE #158: Infrastructure controller should check for correct VPC attributes when existing VPC is used

/area ops-productivity
/kind enhancement
/priority normal
/platform aws
/topology seed
/exp beginner

The AWS extension requires `enableDnsHostnames` and `enableDnsSupport` settings of VPC attributes to be set to `true` in order to function correctly. If these prerequisites are violated then the worker machines might not be able to join the cluster.
Unfortunately, we lack a check in our `Infrastructure` controller in order to submit a meaningful error message to the user.

**Why is this needed**:
Let's please add above mentioned check so that the can help himself.

rfranzke:

/area ops-productivity
/kind enhancement
/priority normal
/platform aws

When an existing VPC is used for the shoot infrastructure then we require both `enableDnsSupport` and `enableDnsHostnames` to be set to `true`, see also the user documentation: https://github.com/gardener/gardener-extension-provider-aws/blob/master/docs/usage-as-end-user.md#infrastructureconfig

This was not validated earlier, i.e., when such a VPC was used, the shoot reconciliation often failed because the worker nodes couldn't join the cluster. The end-user did not know why this is the case and required support.
With this PR, we check the VPC attributes upfront before the `Infrastructure` reconciliation to clearly indicate this error.

Fixes #158

/invite @ialidzhikov 

**Release note**:

```action user
The `Infrastructure` controller does now check for the required VPC attributes `enableDnsSupport` and `enableDnsHostnames` to be enabled when an existing VPC is used for the shoot infrastructure. If any of them is disabled then the reconciliation will fail (see also [user guide](https://github.com/gardener/gardener-extension-provider-aws/blob/master/docs/usage-as-end-user.md#infrastructureconfig)). Make sure to enable these attributes for your existing VPCs that are used for shoots.
```

### DIFF
--- a/pkg/aws/client/client.go
+++ b/pkg/aws/client/client.go
@@ -118,6 +118,27 @@ func (c *Client) GetInternetGateway(ctx context.Context, vpcID string) (string, 
 	return "", fmt.Errorf("no attached internet gateway found for vpc %s", vpcID)
 }
 
+// VerifyVPCAttributes checks whether the VPC attributes are correct.
+func (c *Client) VerifyVPCAttributes(ctx context.Context, vpcID string) error {
+	vpcAttribute, err := c.EC2.DescribeVpcAttributeWithContext(ctx, &ec2.DescribeVpcAttributeInput{VpcId: &vpcID, Attribute: aws.String("enableDnsSupport")})
+	if err != nil {
+		return err
+	}
+	if vpcAttribute.EnableDnsSupport == nil || vpcAttribute.EnableDnsSupport.Value == nil || !*vpcAttribute.EnableDnsSupport.Value {
+		return fmt.Errorf("invalid VPC attributes: `enableDnsSupport` must be set to `true`")
+	}
+
+	vpcAttribute, err = c.EC2.DescribeVpcAttributeWithContext(ctx, &ec2.DescribeVpcAttributeInput{VpcId: &vpcID, Attribute: aws.String("enableDnsHostnames")})
+	if err != nil {
+		return err
+	}
+	if vpcAttribute.EnableDnsHostnames == nil || vpcAttribute.EnableDnsHostnames.Value == nil || !*vpcAttribute.EnableDnsHostnames.Value {
+		return fmt.Errorf("invalid VPC attributes: `enableDnsHostnames` must be set to `true`")
+	}
+
+	return nil
+}
+
 // The following functions are only temporary needed due to https://github.com/gardener/gardener/issues/129.
 
 // ListKubernetesELBs returns the list of ELB loadbalancers in the given <vpcID> tagged with <clusterName>.
--- a/pkg/controller/infrastructure/actuator_reconcile.go
+++ b/pkg/controller/infrastructure/actuator_reconcile.go
@@ -123,14 +123,18 @@ func generateTerraformInfraConfig(ctx context.Context, infrastructure *extension
 			return nil, err
 		}
 		existingVpcID := *infrastructureConfig.Networks.VPC.ID
+		if err := awsClient.VerifyVPCAttributes(ctx, existingVpcID); err != nil {
+			return nil, err
+		}
 		existingInternetGatewayID, err := awsClient.GetInternetGateway(ctx, existingVpcID)
 		if err != nil {
 			return nil, err
 		}
 		vpcID = strconv.Quote(existingVpcID)
 		internetGatewayID = strconv.Quote(existingInternetGatewayID)
+
 	case infrastructureConfig.Networks.VPC.CIDR != nil:
-		vpcCIDR = string(*infrastructureConfig.Networks.VPC.CIDR)
+		vpcCIDR = *infrastructureConfig.Networks.VPC.CIDR
 	}
 
 	var zones []map[string]interface{}
--- a/pkg/aws/client/types.go
+++ b/pkg/aws/client/types.go
@@ -33,6 +33,7 @@ const (
 type Interface interface {
 	GetAccountID(ctx context.Context) (string, error)
 	GetInternetGateway(ctx context.Context, vpcID string) (string, error)
+	VerifyVPCAttributes(ctx context.Context, vpcID string) error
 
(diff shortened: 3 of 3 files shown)


pkg/aws/client/client.go
```diff
@@ -118,6 +118,27 @@ func (c *Client) GetInternetGateway(ctx context.Context, vpcID string) (string,
...
+	if vpcAttribute.EnableDnsSupport == nil || vpcAttribute.EnableDnsSupport.Value == nil || !*vpcAttribute.EnableDnsSupport.Value {
+		return fmt.Errorf("invalid VPC attributes: `enableDnsSupport` must be set to `true`")
+	}
+
+	vpcAttribute, err = c.EC2.DescribeVpcAttributeWithContext(ctx, &ec2.DescribeVpcAttributeInput{VpcId: &vpcID, Attribute: aws.String("enableDnsHostnames")})
+	if err != nil {
+		return err
+	}
+	if vpcAttribute.EnableDnsHostnames == nil || vpcAttribute.EnableDnsHostnames.Value == nil || !*vpcAttribute.EnableDnsHostnames.Value {
+		return fmt.Errorf("invalid VPC attributes: `enableDnsHostnames` must be set to `true`")
```
rfranzke, line 136: Ref https://github.com/gardener/gardener/pull/2702/files#diff-0f3cf517d12e4c9c07b01a0af943b3feR58 - such errors will be categorized with the `ERR_CONFIGURATION_PROBLEM` error code

rfranzke:
@ialidzhikov probably we can merge this PR before #157 so that you are able to add another test case for this situation?

pkg/aws/client/client.go
```diff
@@ -118,6 +118,27 @@ func (c *Client) GetInternetGateway(ctx context.Context, vpcID string) (string,
 	return "", fmt.Errorf("no attached internet gateway found for vpc %s", vpcID)
 }
 
+// VerifyVPCAttributes checks whether the VPC attributes are correct.
+func (c *Client) VerifyVPCAttributes(ctx context.Context, vpcID string) error {
```
ialidzhikov, line 122: I would vote to do not mess the awsclient pkg with requirements specific to the infra actuator. We can have a `DescribeVPCAttribute` func in the client but the real checks to happen in the actuator as the checks are specific to the actuator.

rfranzke: Hm, not sure, I can do it this way if you insist, no problem, but then I'd question this client package at all, i.e., why do we need it if we already have the EC2 client etc (i.e., why do we need to wrap it again if we not unpack our specifics in it)?

rfranzke: We already have Gardener-Extension-AWS-specific code in this client, see https://github.com/gardener/gardener-extension-provider-aws/blob/master/pkg/aws/client/client.go#L144 and following functions.

ialidzhikov:
> @ialidzhikov probably we can merge this PR before #157 so that you are able to add another test case for this situation?

Okay.

ialidzhikov:
Actually I see that `enableDnsHostnames` is required only for suse-chost Nodes to be able to join. CoreOS and gardenlinux nodes can join when `enableDnsHostnames`.
I rolled back to https://github.com/gardener/gardener-extension-os-suse-chost/issues/1 and the to https://github.com/gardener/gardener-extension-os-suse-chost/pull/12 - it says:

> 2. Reusing a VPC that already exists (VPC `id` specified in [InfrastructureConfig](https://github.com/gardener/gardener-extension-provider-aws/blob/master/docs/usage-as-end-user.md#infrastructureconfig)]
> 
> If the second option applies to your use-case please make sure that your VPC has enabled **DNS Support**. Otherwise SUSE JeOS based nodes aren't able to join or operate in your cluster properly.

How the statement above is now generalised like the one below?

> To make sure the nodes are able to join and operate in your cluster properly, please make sure that your VPC has enabled DNS Support, explicitly the attributes enableDnsHostnames and enableDnsSupport must be set to true.

Is there something that I miss? Any other component in the Shoot that requires `enableDnsHostnames`? `metrics-server`?

ialidzhikov:
Okay, I am replying to myself with https://github.com/gardener/gardener-extension-provider-aws/pull/93.

ialidzhikov:
Yep, `enable_dns_hostnames = true` is needle for metrics-server to be able to work properly - ref https://github.com/awslabs/amazon-eks-ami/issues/55#issuecomment-470721555.

Generally, lgtm.
I am a little bit concerned about the compatibility. Previous successful reconciliations or Shoot creations with existing vpc with `enable_dns_hostnames = false` will now start to fail. We need at least release note which is action required for users.

rfranzke:
@ialidzhikov done, thanks for checking

pkg/controller/infrastructure/actuator_reconcile.go
```diff
@@ -123,14 +123,18 @@ func generateTerraformInfraConfig(ctx context.Context, infrastructure *extension
 			return nil, err
 		}
 		existingVpcID := *infrastructureConfig.Networks.VPC.ID
+		if err := awsClient.VerifyVPCAttributes(ctx, existingVpcID); err != nil {
```
timebertt, line 126: BTW, what happens when a user specifies a non-existing vpc-id? I guess, this will fail here, right?
Would it make sense to check if the vpc exists explicitly so we can return a meaningful error message?

rfranzke: The error message of the `Infrastructure` controller will be

```
failed to generate Terraform config: InvalidVpcID.NotFound: The vpc ID 'vpc-0897467802289706f' does not exist\n\tstatus code: 400, request id: 443e0c6e-b212-4a49-b3a7-35198f322ba9"
```

timebertt: Well, that's meaningful enough :)
Thanks for checking!

PR #173, opened 2020-08-30
Add kcm mounts when not using csi

danielfoehrKn:

/area os
/kind enhancement
/kind bug
/priority normal
/platform aws

Adds a host path volume to the KCM that is required for certain OS.

The KCM discovers the Root CAs available on the node via [predefined locations](https://github.com/golang/go/blob/1bb247a469e306c57a5e0eaba788efb8b3b1acef/src/crypto/x509/root_linux.go#L7-L15).

```
	"/etc/ssl/certs/ca-certificates.crt",                // Debian/Ubuntu/Gentoo etc.
	"/etc/pki/tls/certs/ca-bundle.crt",                  // Fedora/RHEL 6
	"/etc/ssl/ca-bundle.pem",                            // OpenSUSE
	"/etc/pki/tls/cacert.pem",                           // OpenELEC
	"/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem", // CentOS/RHEL 7
	"/etc/ssl/cert.pem",                                 // Alpine Linux
```

This is why the kcm mounts the `etc/ssl` directory.

For certain OS (e.g coreos 2512.3.0), the `etc/ssl/certs` contains only symlinks to the actual `ca-certificates.crt` file.
```
lrwxrwxrwx. 1 root root    54 May 22 20:41  ca-certificates.crt -> ../../../usr/share/ca-certificates/ca-certificates.crt
```
 
However, the KCM cannot access `ca-certificates/ca-certificates.crt` , because it does not mount the correct directory.

Fixes #

Please also see [related issue on kubeadm for coreos](https://github.com/kubernetes/kubeadm/issues/671) and [this](https://github.com/kubernetes/kubernetes/pull/59122).

We can also think about mounting the other directories like `"/etc/pki/tls/certs/`. 
But I would like to talk about and first go ahead with this issue.

This change is independent of the provider, so should also be added for other providers as well.
However, we are having issues on AWS, so this should be first.

**Release note**:

```improvement operator
KCM mounts `/usr/share/ca-certificates` required for certain OS (e.g CoreOS 2512.3.0),
```

### DIFF
--- a/pkg/webhook/controlplane/ensurer.go
+++ b/pkg/webhook/controlplane/ensurer.go
@@ -246,6 +246,23 @@ var (
 		},
 	}
 
+	usrShareCaCerts            = "usr-share-cacerts"
+	directoryOrCreate          = corev1.HostPathDirectoryOrCreate
+	usrShareCaCertsVolumeMount = corev1.VolumeMount{
+		Name:      usrShareCaCerts,
+		MountPath: "/usr/share/ca-certificates",
+		ReadOnly:  true,
+	}
+	usrShareCaCertsVolume = corev1.Volume{
+		Name: usrShareCaCerts,
+		VolumeSource: corev1.VolumeSource{
+			HostPath: &corev1.HostPathVolumeSource{
+				Path: "/usr/share/ca-certificates",
+				Type: &directoryOrCreate,
+			},
+		},
+	}
+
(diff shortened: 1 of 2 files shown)


timuthy (approved):
/lgtm

vpnachev (approved):
/lgtm

rfranzke (approved):
/lgtm

PR #179, opened 2020-09-07
Change volumeBindingMode to WaitForFirstConsumer

### LINKED ISSUE #152: Specify volumeBindingMode:WaitForFirstConsumer in default storage class
/area storage
/kind enhancement
/priority normal
/platform aws

PVs shall be created in the zone where the pod will be scheduled to.

**Why is this needed**:
From https://kubernetes.io/docs/concepts/storage/storage-classes/#volume-binding-mode:
> By default, the Immediate mode indicates that volume binding and dynamic provisioning occurs once the PersistentVolumeClaim is created. For storage backends that are topology-constrained and not globally accessible from all Nodes in the cluster, PersistentVolumes will be bound or provisioned without knowledge of the Pod's scheduling requirements. This may result in unschedulable Pods.

We use `Immediate`, but should rather use `WaitForFirstConsumer`, wouldn't you agree?

ialidzhikov:

/area storage
/kind task
/priority normal
/platform aws

Change volumeBindingMode to WaitForFirstConsumer.

Fixes #152

**Release note**:

```noteworthy user
The `volumeBindingMode` of the StorageClasses managed by Gardener is now switched to `WaitForFirstConsumer`.
```

### DIFF
--- a/charts/internal/shoot-storageclasses/templates/storageclasses_legacy.yaml
+++ b/charts/internal/shoot-storageclasses/templates/storageclasses_legacy.yaml
@@ -9,6 +9,11 @@ metadata:
     storageclass.kubernetes.io/is-default-class: "true"
 allowVolumeExpansion: true
 provisioner: kubernetes.io/aws-ebs
+{{- if semverCompare ">= 1.12-0" .Capabilities.KubeVersion.GitVersion }}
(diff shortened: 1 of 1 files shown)


rfranzke (approved):
/lgtm

PR #191, opened 2020-10-01
Adding additional test definition provider-aws-extended 

kris94:
/kind enhancement
/priority normal
/platform aws

This PR adds additional test definition "provider-aws-extended" that includes all parameters from the respective test file, not only the mandatory, like the "provider-aws" test definition.
This test definition will be used by the integration tests where shoots with different networks from the defaults must be created (e.g. for registering a new seed for the control plane migration).
Fixes #

**Release note**:
NONE

### DIFF
--- a/.test-defs/provider-aws-extended.yaml
+++ b/.test-defs/provider-aws-extended.yaml
@@ -1,0 +1,21 @@
+kind: TestDefinition
+metadata:
+  name: gen-provider-aws-extended
+spec:
+  owner: gardener-oq@listserv.sap.com
+  description: Generates the aws provider specific configurations
+  activeDeadlineSeconds: 3600
+
+  command: [bash, -c]
+  args:
+  - >-
+    go run -mod=vendor ./test/tm/generator.go
+     --infrastructure-provider-config-filepath=$INFRASTRUCTURE_PROVIDER_CONFIG_FILEPATH
+     --controlplane-provider-config-filepath=$CONTROLPLANE_PROVIDER_CONFIG_FILEPATH
+     --network-vpc-cidr=$NETWORK_VPC_CIDR
+     --network-internal-cidr=$NETWORK_INTERNAL_CIDR
+     --network-public-cidr=$NETWORK_PUBLIC_CIDR
+     --network-worker-cidr=$NETWORK_WORKER_CIDR
+     --zone=$ZONE
+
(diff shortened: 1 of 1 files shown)


schrodit (approved):
/lgtm

rfranzke (approved):

.test-defs/provider-aws-extended.yaml
```diff
@@ -0,0 +1,21 @@
...
+  command: [bash, -c]
+  args:
+  - >-
+    go run -mod=vendor ./test/tm/generator.go
+     --infrastructure-provider-config-filepath=$INFRASTRUCTURE_PROVIDER_CONFIG_FILEPATH
+     --controlplane-provider-config-filepath=$CONTROLPLANE_PROVIDER_CONFIG_FILEPATH
+     --network-vpc-cidr=$NETWORK_VPC_CIDR
+     --network-internal-cidr=$NETWORK_INTERNAL_CIDR
+     --network-public-cidr=$NETWORK_PUBLIC_CIDR
+     --network-worker-cidr=$NETWORK_WORKER_CIDR
```
timebertt, line 18: Why do we have to add a second test def for these flags?
If they are optional, why can't we simply add them to the existing test def and just default them if unset?

timebertt (commented):
Sorry for jumping in just now, but I wonder why adding a second test def was necessary...

.test-defs/provider-aws-extended.yaml
plkokanov: We wondered whether to do it as you suggest.

If any of the outside supplied values (e.g. $NETWORK_INTERNAL_CIDR) are unset the command will take them as empty strings and not use the defaulted ones specified when defining the flags. So we would need some additional logic to handle empty strings.

dguendisch: @danielfoehrKn has for example done exactly that for another property he added a while ago. I recall a longer discussion about exactly what you mentioned: https://github.com/gardener/gardener/pull/2396#pullrequestreview-419317233
As you said, a defaulting in code would be needed.

timebertt: Yep, that sounds good :)

dguendisch: Yes, less clutter, less to remember 👍

PR #193, opened 2020-10-05
Changing the default parameters handling and adding all properties to the test definition

kris94:
How to categorize this PR?
/kind enhancement
/priority normal
/platform aws

What this PR does / why we need it:
This PR removes the additional test definition "provider-aws-extended" because it is obsolete. Also, it adds all properties to the "provider-aws"  test definition. In addition to this, as discussed in [PR:191](https://github.com/gardener/gardener-extension-provider-aws/pull/191), the default parameters handling is changed.
This way one test definition will be used by the integration tests where shoots with different networks from the defaults must be created (e.g. for registering a new seed for the control plane migration) and it will preserve the default behaviour for other integration tests (e.g. for shoot creation) that relies on the default values.
Which issue(s) this PR fixes:
Fixes #

Special notes for your reviewer:

Release note:
NONE


test/tm/generator.go
```diff
@@ -29,21 +30,43 @@ import (
...
+	networkPublicCidr                string
+	networkWorkerCidr                string
+	infrastructureProviderConfigPath string
+	controlplaneProviderConfigPath   string
+	zone                             string
+}
 
-	zone = flag.String("zone", "", "cloudprovider zone fo the shoot")
-)
+func init() {
```
timebertt, line 55: Can you move this to `main` or some function `addFlags` or similar?
I don't like the implicitness of `init` functions and try to use them only when necessary. Apart from that having it in a dedicated `addFlags` func will improve readability in general.

timebertt (changes requested):

timebertt (approved):
Thanks!
/lgtm

PR #209, opened 2020-11-16
Document rate limit issues

BeckerMax:
Add documentation regarding rate limiting.

Outcome from: https://github.com/gardener/gardener/issues/3120

### DIFF
--- a/docs/usage-as-end-user.md
+++ b/docs/usage-as-end-user.md
@@ -22,7 +22,10 @@ data:
 ```
 
 The [AWS documentation](https://docs.aws.amazon.com/general/latest/gr/aws-sec-cred-types.html#access-keys-and-secret-access-keys) explains the necessary steps to enable programmatic access, i.e. create **access key ID** and **access key**, for the user of your choice.
+
 ⚠️ For security reasons, we recommend creating a **dedicated user with programmatic access only**. Please avoid re-using a IAM user which has access to the AWS console (human user).
+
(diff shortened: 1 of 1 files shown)


docs/usage-as-end-user.md
```diff
@@ -22,8 +22,11 @@ data:
 ```
 
 The [AWS documentation](https://docs.aws.amazon.com/general/latest/gr/aws-sec-cred-types.html#access-keys-and-secret-access-keys) explains the necessary steps to enable programmatic access, i.e. create **access key ID** and **access key**, for the user of your choice.
+
 ⚠️ For security reasons, we recommend creating a **dedicated user with programmatic access only**. Please avoid re-using a IAM user which has access to the AWS console (human user).
 
+⚠️ We have seen rate limiting issues with credential reuse between Shoot clusters. If you plan to update fifty or more clusters in parallel in the same AWS region please spread them accross different AWS accounts.
```
rfranzke, line 28: Where is this artificial number 50 coming from? :-) I would not state this because it highly depends on what the user is doing with his AWS account, which other workloads (not gardener/shoots) are using the AWS API, etc.

I would only state that depending on the AWS API usage it can be problematic to have more than a certain number of shoots because of rate limits, and in this case multiple AWS accounts should be used if the number of shoots shall be increased further.

BeckerMax: corrected it

rfranzke (approved):
/lgtm

timebertt (approved):
/lgtm

PR #226, opened 2020-11-26
Expose csi-driver and cloud-controller-manager logs

Kristian-ZH:

/area logging
/kind enhancement
/priority normal
/platform aws

Add two new dashboards exposed to the end users which will show logs for CKM and the csi-driver-controller

Fixes #

@rfranzke @vlvasilev 

**Release note**:

```improvement user
New dashboards which expose logs for `cloud-controller-manager` and `csi-driver-controller`.
```


charts/internal/seed-controlplane/charts/csi-driver-controller/templates/csi-driver-controller-deployment.yaml
```diff
@@ -76,7 +76,7 @@ spec:
         - name: socket-dir
           mountPath: {{ .Values.socketPath }}
 
-      - name: csi-provisioner
+      - name: aws-csi-provisioner
```
ialidzhikov, line 79: I don't get why you change the container names? There is nothing aws specific in this container to call it `aws-csi-provisioner`.

Kristian-ZH: Because when the seed contains multiple provider-extensions, in the fluent-bit Configmap will be added the same rewrite_tag filter: https://github.com/gardener/gardener-extension-provider-aws/pull/226/files#diff-019be355db9cbaece5133ea2a8f44876159eb1b7d2a119906d1c9b3d3c09a9c3R31
multiple times

rfranzke: It is the same for ccm/mcm

rfranzke (dismissed):
/lgtm

charts/internal/seed-controlplane/charts/csi-driver-controller/templates/csi-driver-controller-deployment.yaml
ialidzhikov: Yeah but ccm has only 1 container it is cloud-provider specific and you can name it `{provider}-cloud-controller-manager`. For CSI we have 1 cloudprovider specific container and the other 5 sidecar containers are not.

ialidzhikov: IMO for the CSI case it is better to rename the deployment to `{provider}-csi-driver-controller`. Calling the external-provisioner `aws-csi-provisioner` sounds misleading.

charts/internal/seed-controlplane/charts/cloud-controller-manager/templates/configmap-observability.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v1
 kind: ConfigMap
 metadata:
-  name: cloud-controller-manager-monitoring-config
```
vpnachev, line 4: Is this the configmap that will leak?

charts/internal/seed-controlplane/charts/csi-driver-controller/templates/csi-driver-controller-deployment.yaml
rfranzke: @ialidzhikov if we do this then I suggest to also rename the CCM/MCM deployments so that it's consistent. WDYT?

charts/internal/seed-controlplane/charts/cloud-controller-manager/templates/configmap-observability.yaml
Kristian-ZH: Yes it is.

charts/internal/seed-controlplane/charts/csi-driver-controller/templates/csi-driver-controller-deployment.yaml
ialidzhikov: Sounds good to me.

vpnachev: With the out-of-tree of MCM, it already have the [provider name](https://github.com/gardener/gardener-extension-provider-gcp/blob/89f8007cdfa09cb5939eb000f399fc1eb1a0d372/charts/internal/machine-controller-manager/seed/templates/deployment.yaml#L42-L71) in its containers:
- gcp-machine-controller-manager
- controller-manager-provider-gcp

Maybe the general controller can then remove the provider name from the container name?

charts/gardener-extension-provider-aws/templates/configmap-logging.yaml
```diff
@@ -20,6 +20,20 @@ data:
         Key_Name            log
         Parser              kubeapiserverParser
         Reserve_Data        True
+
```
rfranzke, line 23: line 12 and 19 should be adapted, right?

pkg/controller/controlplane/provider_actuator.go
```diff
@@ -0,0 +1,142 @@
+// Copyright (c) 2019 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```
rfranzke, line 1:
```suggestion
// Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```

pkg/controller/controlplane/provider_actuator.go
```diff
@@ -0,0 +1,142 @@
...
+
+// InjectClient injects the given client into the valuesProvider.
+func (a *providerActuator) InjectClient(client client.Client) error {
+	injector, ok := a.genericActuator.(inject.Client)
+	if ok {
+		a.client = client
+		return injector.InjectClient(client)
+	}
+
+	return fmt.Errorf("Cotroller generic actuator does not implement Config interface")
```
rfranzke, line 115: typo

pkg/controller/controlplane/valuesprovider.go
```diff
@@ -197,7 +197,7 @@ var (
 				Objects: []*chart.Object{
 					{Type: &corev1.Service{}, Name: aws.CloudControllerManagerName},
 					{Type: &appsv1.Deployment{}, Name: aws.CloudControllerManagerName},
```
rfranzke, line 199: This needs to be adapted, right?

pkg/controller/controlplane/valuesprovider.go
```diff
@@ -216,6 +216,7 @@ var (
 					// csi-driver-controller
 					{Type: &appsv1.Deployment{}, Name: aws.CSIControllerName},
```
rfranzke, line 217: This needs to be adapted, right?

pkg/controller/controlplane/provider_actuator.go
```diff
@@ -0,0 +1,142 @@
...
+
+// InjectConfig injects the given config into the actuator.
+func (a *providerActuator) InjectConfig(config *rest.Config) error {
+
+	injector, ok := a.genericActuator.(inject.Config)
+	if ok {
+		return injector.InjectConfig(config)
+	}
+
+	return fmt.Errorf("Cotroller generic actuator does not implement Config interface")
```
rfranzke, line 104: typo

pkg/controller/controlplane/provider_actuator.go
```diff
@@ -0,0 +1,142 @@
...
+}
+
+// InjectFunc enables injecting Kubernetes dependencies into actuator's dependencies.
+func (a *providerActuator) InjectFunc(f inject.Func) error {
+	injector, ok := a.genericActuator.(inject.Injector)
+	if ok {
+		return injector.InjectFunc(f)
+	}
+
+	return fmt.Errorf("Cotroller generic actuator does not implement Injector interface")
```
rfranzke, line 93: typo

pkg/controller/controlplane/provider_actuator.go
```diff
@@ -0,0 +1,142 @@
...
+		return injector.InjectClient(client)
+	}
+
+	return fmt.Errorf("Cotroller generic actuator does not implement Config interface")
+}
+
+func (a *providerActuator) Reconcile(ctx context.Context, cp *extensionsv1alpha1.ControlPlane, cluster *extensionscontroller.Cluster) (bool, error) {
+	//Delete leaked resources
+	a.logger.Info("Cleanup leaked resources", "controlplane", kutil.ObjectName(cp))
+	if err := a.client.Delete(ctx, &corev1.ConfigMap{ObjectMeta: metav1.ObjectMeta{Name: "cloud-controller-manager-monitoring-config", Namespace: cp.Namespace}}); err != nil && !errors.IsNotFound(err) {
```
rfranzke, line 121:
```suggestion
	if err := a.client.Delete(ctx, &corev1.ConfigMap{ObjectMeta: metav1.ObjectMeta{Name: "cloud-controller-manager-monitoring-config", Namespace: cp.Namespace}}); client.IgnoreNotFound(err) != nil {
```

pkg/controller/controlplane/provider_actuator.go
```diff
@@ -0,0 +1,142 @@
...
+		return false, err
+	}
+
+	if err := a.client.Delete(ctx, &appsv1.Deployment{ObjectMeta: metav1.ObjectMeta{Name: "cloud-controller-manager", Namespace: cp.Namespace}}); err != nil && !errors.IsNotFound(err) {
```
rfranzke, line 125:
```suggestion
	if err := a.client.Delete(ctx, &appsv1.Deployment{ObjectMeta: metav1.ObjectMeta{Name: "cloud-controller-manager", Namespace: cp.Namespace}}); client.IgnoreNotFound(err) != nil {
```

pkg/controller/controlplane/provider_actuator.go
```diff
@@ -0,0 +1,142 @@
...
+	"k8s.io/client-go/rest"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/runtime/inject"
+)
+
+type providerActuator struct {
+	client client.Client
+	logger logr.Logger
+
+	genericActuator controlplane.Actuator
```
rfranzke, line 45:
```suggestion
	controlplane.Actuator
```

pkg/controller/controlplane/provider_actuator.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+
+	if err := a.client.Delete(ctx, &appsv1.Deployment{ObjectMeta: metav1.ObjectMeta{Name: "cloud-controller-manager", Namespace: cp.Namespace}}); err != nil && !errors.IsNotFound(err) {
+		return false, err
+	}
+
+	return a.genericActuator.Reconcile(ctx, cp, cluster)
+}
+
+func (a *providerActuator) Delete(ctx context.Context, cp *extensionsv1alpha1.ControlPlane, cluster *extensionscontroller.Cluster) error {
```
rfranzke, line 132: This function should also be adapted, or?

pkg/controller/controlplane/provider_actuator.go
```diff
@@ -0,0 +1,142 @@
...
+	return a.genericActuator.Delete(ctx, cp, cluster)
+}
+
+func (a *providerActuator) Restore(ctx context.Context, cp *extensionsv1alpha1.ControlPlane, cluster *extensionscontroller.Cluster) (bool, error) {
+	return a.genericActuator.Restore(ctx, cp, cluster)
+}
+
+func (a *providerActuator) Migrate(ctx context.Context, cp *extensionsv1alpha1.ControlPlane, cluster *extensionscontroller.Cluster) error {
+	return a.genericActuator.Migrate(ctx, cp, cluster)
+}
```
rfranzke, line 142: With above suggestion (inline the generic actuator into the providerActuator struct) these lines could probably be deleted.

rfranzke (changes requested):
I miss the deletion of the old machine-controller-manager deployment. Also, wouldn't the approach taken in https://github.com/gardener/gardener-extension-provider-alicloud/pull/64 be easier?

Anyways, after thinking again about this renaming change, it's actually not very nice that the deployment names will differ between all infrastructures. Sorry for that, but maybe we should keep the deployment names as they are and rather rename the containers and prefix them with the infra name. @ialidzhikov @vpnachev @Kristian-ZH do you agree? It's way easier for operators/users of Gardener if we keep the deployment names, there might even be tools like gardenctl that potentially break, etc.

ialidzhikov:
> Anyways, after thinking again about this renaming change, it's actually not very nice that the deployment names will differ between all infrastructures. Sorry for that, but maybe we should keep the deployment names as they are and rather rename the containers and prefix them with the infra name. @ialidzhikov @vpnachev @Kristian-ZH do you agree? It's way easier for operators/users of Gardener if we keep the deployment names, there might even be tools like gardenctl that potentially break, etc.

Okay, let's keep the deployment names as they are and rename the containers (as it was done in the initial change before my comment). @Kristian-ZH , I am sorry for the confusion.

Kristian-ZH:
The changes are reverted to the initial approach (changing the container names) + the cleanup logic is now similar like the  logic @rfranzke mentioned above from Alicloud

vpnachev (approved):
/lgtm
/invite @wyb1 @istvanballok @vlvasilev

vlvasilev (approved):
/lgtm

rfranzke (approved):
/lgtm

Great work @Kristian-ZH thank you!
/invite @ialidzhikov 
/assign @ialidzhikov

rfranzke:
ping @ialidzhikov

ialidzhikov:
> ping @ialidzhikov

I am already on it. No need for additional `ping`s.

ialidzhikov:
I tried the change locally and the `CSI Driver Controller` dashboard does not show any logs to me for any of the containers. This applies for the users grafana and for the operators grafana as well.

Screenshot 2020-12-03 at 10 07 36

When I search for the logs of this pod, I see that there are logs in loki.

Screenshot 2020-12-03 at 10 09 25

Is this also reproducible for you?

Kristian-ZH:
Nope, if you have a time we can check this together?

ialidzhikov:
The issue was on my side, I forgot to update the chart field of my ctrlreg in my local setup.

PR #235, opened 2020-12-09
Terraformer v2

dkistner:

/area quality robustness
/kind enhancement
/priority normal
/platform aws

This PR let the infrastructure actuator use the terraformer v2.

**Release note**:

```noteworthy operator
The terraformer version has been upgraded to version `v2.0.0`.
```

/invite @timebertt

### DIFF
--- a/charts/images.yaml
+++ b/charts/images.yaml
@@ -2,7 +2,7 @@ images:
 - name: terraformer
   sourceRepository: github.com/gardener/terraformer
   repository: eu.gcr.io/gardener-project/gardener/terraformer
-  tag: "v1.5.0"
+  tag: "v2.0.0"
 
 - name: cloud-controller-manager
   sourceRepository: github.com/kubernetes/kubernetes
--- a/pkg/controller/infrastructure/actuator.go
+++ b/pkg/controller/infrastructure/actuator.go
@@ -50,6 +50,8 @@ func newTerraformer(logger logr.Logger, restConfig *rest.Config, purpose string,
 	}
 
 	return tf.
+		UseV2(true).
+		SetLogLevel("debug").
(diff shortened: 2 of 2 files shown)


timebertt (approved):
/lgtm
/test

dkistner:
Can we merge this? (I have no privileges to do it)

rfranzke (approved):

timebertt:
@gardener/test-infra-maintainers We seem to have some problems with triggering bot tests lately, see [above](https://github.com/gardener/gardener-extension-provider-aws/pull/235#pullrequestreview-547983671), where `/test` was commented but no test was triggered.
I have observed this several times in other PRs, too. 
Is this a known issue?

ialidzhikov:
@timebertt , I think it didn't work in your comment as there are also other command (something) before the test command. Looks like for now it is searching for exact match. I think that it didn't work in the past when there was also trailing whitespace. But just FYI, these are not major issues.

schrodit:
>  , I think it didn't work in your comment as there is also other command (something) before the test command. Looks like for now it is searching for exact match

The bot should parse one command for each line so `\n` should separate the commands.

I cannot find anything in the logs of the bot so I can also suspect that the webhook wasn't delivered correctly.
@timebertt can you check in the github webhook delivery log if the event was correctly delivered to the bot (I have no access to check that myself)

But maybe there is an issue with the parsing. I will try to reproduce it locally.

timebertt:
> I cannot find anything in the logs of the bot so I can also suspect that the webhook wasn't delivered correctly.
> @timebertt can you check in the github webhook delivery log if the event was correctly delivered to the bot (I have no access to check that myself)

Me neither...

PR #237, opened 2020-12-14
Support for Kubernetes v1.20

rfranzke:
**How to categorize this PR?**

/area open-source usability
/kind enhancement
/priority normal
/platform aws
/exp intermediate
/topology garden seed shoot

**What this PR does / why we need it**:
This PR adds support for 1.20 to the extension.

**Which issue(s) this PR fixes**:
Part of gardener/gardener#3260

**Special notes for your reviewer**:
* :white_check_mark: Depends on gardener/gardener#3296
* I have successfully validated the functionality as follows:
  * :white_check_mark: Create new clusters with versions < 1.20
  * :white_check_mark: Create new clusters with version  = 1.20
  * :white_check_mark: Upgrade old clusters from version 1.19 to version 1.20
  * :white_check_mark: Delete clusters with versions < 1.20
  * :white_check_mark: Delete clusters with version  = 1.20

**Release note**:

```feature user
The AWS extension does now support shoot clusters with Kubernetes version 1.20. You should consider the [Kubernetes release notes](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.20.md) before upgrading to 1.20. 
```

PR #212, opened 2020-11-17
Migration of in-tree to out-of-tree MCM providers

prashanth26:
- This also includes migration from AWSMachineClass to MachineClass. This migration occurs implicitly without causing rollouts of existing nodes/VMs.
- Added tests to verify deletion of old machineClass objects

**How to categorize this PR?**

/area dev-productivity
/kind  api-change
/priority normal
/platform aws

**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Release note**:

```breaking operator
⚠️ Before upgrading your `gardener/gardener-extension-provider-aws` to [>= v1.20.0](https://github.com/gardener/gardener-extension-provider-aws/releases/tag/v1.20.0), please upgrade your `gardener/gardener` component version to [>= v1.14.0](https://github.com/gardener/gardener/releases/tag/v1.14.0) to avoid breaking of clusters that are using the scale from/to zero feature (clusters that allowing scaling from/to 0 worker pools). If used with an older `gardener/gardener` version, this would lead to failure of clusters making use of this feature. 
```
```other developer
Migration of MCM provider from in-tree to out-of-tree. Refer - [MCM provider AWS](https://github.com/gardener/machine-controller-manager-provider-aws).
```
```other developer
Migration of `AWSMachineClass` to `MachineClass`. This migration occurs implicitly without causing rollouts of existing nodes/VMs.
```

PR #238, opened 2020-12-16
update mcm to use worker secret ref. for the machine class credentialsSecretRef

danielfoehrKn:

/area quality
/kind enhancement
/priority normal
/platform aws

Uses the machine classes  .{spec.}credentialsSecretRef field. It is a reference to a secret containing only the credentials, while today's .{spec.}secretRef still contains the user-data. See [here for more details](https://github.com/gardener/machine-controller-manager/pull/578).

Uses the worker's secret reference as the credentialsSecretRef.
This means all worker pools use the same "cloudprovider" secret containing only the cloud provider credentials.
The existing MachineClass SecretReference then only contains the user data that is different for each pool. 

Fixes #

**Release note**:

```other operator
Reducing credential update complexity by all the machine classes using the new .{spec.}credentialsSecretRef field.
This means all worker pools use the same "cloudprovider" secret containing only the cloud provider credentials.
The existing MachineClass SecretReference only contains the user data that is different for each pool.
```


rfranzke:
Changes look great, will test it and come back

prashanth26:
We can go ahead with merging this PR first. I will rebase these changes on https://github.com/gardener/gardener-extension-provider-aws/pull/212 on this is merged.

rfranzke:
@danielfoehrKn can you rebase now? I guess no more revendoring is necessary because it's already done on `master` (#237)

prashanth26 (approved):
/lgtm

ialidzhikov (approved):
/lgtm
/squash

rfranzke (approved):
/lgtm
/hold until v1.19 got released

prashanth26:
Hi @danielfoehrKn , Hi Colleagues,

I was going through all the changes for `provider-aws` release over the weekend. And I realized that there is a corner case that we might have missed that could potentially cause clusters hanging during deletion. 

We have [this method](https://github.com/gardener/gardener/blob/master/extensions/pkg/controller/worker/genericactuator/actuator_delete.go#L160) that waits until all machine objects are cleaned up by MCM before deleting it. However, this does not check for the MCM finalizers on `cloudprovider` which will exist with this change as MCM uses `cloudprovider` secret now. 

I tried to reproduce this scenario with several shoot deletions however was unsuccessful has I think this case can be hit only when `cloudprovider` finalizer update fails (as currently finalizer removal is too fast and the race condition isn't hit). However, I am pretty confident that we need this check to avoid such cases, and we could verify this with unit tests to reproduce this case. 

The check would be to wait until the finalizer is removed for `cloudprovider` secret similar to this - https://github.com/gardener/gardener/blob/master/extensions/pkg/controller/worker/genericactuator/actuator_delete.go#L224. I think we need this change in g/g and revendor it to all providers before releasing any of the extension providers. WDYT?

ialidzhikov:
In our sync about this we agreed to add the `gardener.cloud/purpose=machineclass` label to the worker `.spec.secretRef` (aka cloudprovider Secret) and adapt the check in https://github.com/gardener/gardener/blob/ee91337bdf786a45a9b025cfc5ca883936240475/extensions/pkg/controller/worker/genericactuator/actuator_delete.go#L231-L235 to check only for the removal of the mcm finalizer.

But checking the code now again, I guess this won't work as we have a `cleanupMachineClassSecrets` func (ref https://github.com/gardener/gardener/blob/v1.15.3/extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go#L181-L184) that will the delete the cloudprovider secret assuming that it is a machine class secret that is stale.

prashanth26:
> https://github.com/gardener/gardener/blob/ee91337bdf786a45a9b025cfc5ca883936240475/extensions/pkg/controller/worker/genericactuator/actuator_delete.go#L231-L235

Yes, Ismail, this wouldn't work. 

The check would be to wait until the finalizer is removed for `cloudprovider` secret similar to this here. However instead of waiting for the secret deletion, we would have to wait only until finalizer is removed.

PR #240, opened 2020-12-17
Apply --ignore-operation-annotation for backupentry and backupbucket controllers

ialidzhikov:
/kind bug

https://github.com/gardener/gardener/pull/2290 adds the IgnoreOperationAnnotation option to the backupentry and backupbucket controllers but this config is not applied and always is set to false currently.

**Release note**:

```bugfix operator
An issue causing the backupentry and backupbucket controllers to not respect the `--ignore-operation-annotation` is now fixed.
```


ialidzhikov:
I am really tired of this.

/close

PR #251, opened 2021-01-07
Do not gitignore *-packr.go files

### LINKED ISSUE #81: *-packr.go files are ignored by git
**What happened**:
Currently changes in `charts/images.yaml` are not applied in the docker image if `make generate` is not run.
Also `check-generate` won't detect any outdated generated file.

After updating an entry in `charts/images.yaml` and building a docker image after that, I would expect my change to be respected.

1. Update an image tag in `charts/images.yaml`

For example update the `csi-snapshotter` image tag from `v2.1.0` to `v2.1.1`

```diff
 - name: csi-snapshotter
   sourceRepository: github.com/kubernetes-csi/external-snapshotter
   repository: quay.io/k8scsi/csi-snapshotter
-  tag: "v2.1.0"
+  tag: "v2.1.1"
```

2. Build and push an image

```
$ docker build -t foo/gardener-extension-provider-aws:bar -f Dockerfile --target gardener-extension-provider-aws .
$ docker push foo/gardener-extension-provider-aws:bar
```

3. Update the ControllerRegistration with the newly built image and ensure that the `charts/images.yaml` file in the new container is updated with the change from step 1.

4. Ensure that the update step from step 1 has no effect and the extension controller is using the old tag (`v2.1.0`).

To fix this issue, the developer needs to first run `make generate` and then to build and push his docker image.

**Environment**:

- Extension version:
- Others:

ialidzhikov:
/kind bug
/platform aws

Fixes #81

**Release note**:

```bugfix developer
provider-aws does no longer gitignore `*-packr.go` files. Until now `*-packr.go` files were gitignored and a developer had to run always `make generate` before running `make start` or before building an image to make sure that provider-aws won't use and outdated images.yaml file.
```


rfranzke (approved):
/lgtm

ialidzhikov:
From https://github.com/gobuffalo/packr/tree/v2.8.1/v2#building-a-binary:

> *NOTE*: It is not recommended to check-in these generated -packr.go files. They can be large, and can easily become out of date if not careful. It is recommended that you always run packr2 clean after running the packr2 tool.

`packr2` generates different files (different output) when it is executed from a container and from my machine for example. My understanding from the above statement is that it is not recommended to commit and push the generated `-packr.go` files.

timebertt:
Should we then close #81?

PR #253, opened 2021-01-07
Slightly refactor validation of infra config against CloudProfile

rfranzke:

/area dev-productivity
/kind cleanup bug
/priority normal
/platform aws

After https://github.com/gardener/gardener-extension-provider-openstack/pull/204 this PR now slightly refactors the logic of validating the `.spec.provider.infrastructureConfig` of a `Shoot` against the constraints in the `CloudProfile`.

**Release note**:

```bugfix operator
It is now possible to remove zones from the `CloudProfile` without breaking the possibility of adding new, still allowed zones to the `.spec.provider.infrastructureConfig` of `Shoot`s which are using the removed zone.
```

### DIFF
--- a/pkg/apis/aws/validation/infrastructure.go
+++ b/pkg/apis/aws/validation/infrastructure.go
@@ -15,7 +15,6 @@
 package validation
 
 import (
-	"fmt"
 	"regexp"
 	"strings"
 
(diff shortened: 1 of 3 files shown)


ialidzhikov (approved):
Played with the PR locally, it worked well -> the validator now allows adding new (supported zones in the CloudProfile) to an infrastructureConfig that has a zone that is no longer present in the CloudProfile.

/lgtm

PR #263, opened 2021-01-29
vendor g/g@1.17.0 and configurable resource lock

timebertt:

/area scalability
/kind enhancement
/priority normal
/platform aws

This PR makes the leader election resource lock of provider-aws configurable by vendoring `g/g@v1.17.0` and adding the chart value `leaderElection.resourceLock`.

Fixes #

✅ depends on https://github.com/gardener/gardener/pull/3479

**Release note**:

```feature operator
It is now possible to specify the leader election resource lock via the command line flag `--leader-election-resource-lock` (defaults to `configmapsleases`) and the chart value `leaderElection.resourceLock`. Please be careful when changing the resource lock and always migrate via multilocks in order to prevent situations where multiple instances of the controller are running with leader election and thus acting on the same resources.
```


timebertt:
Updated the PR to vendor `g/g@v1.17.0` and adapting to the other breaking changes of the release.

/ready
/title Vendor g/g@1.17.0 and configurable resource lock

rfranzke (approved):
/lgtm

PR #281, opened 2021-02-21
Switch to kaniko oci builder

ialidzhikov:
/kind enhancement

See https://github.com/gardener/gardener-extension-os-coreos/pull/19 and https://github.com/gardener/gardener-extension-os-coreos/pull/20 for more details.

**Release note**:

```other operator
NONE
```

PR #283, opened 2021-02-25
Fix the erroneous pipeline definition

ialidzhikov:
/kind bug

It seems that the publish step is no longer valid after https://github.com/gardener/gardener-extension-provider-aws/pull/281

**Release note**:

```other operator
NONE
```


vpnachev (approved):
/lgtm

PR #288, opened 2021-03-03
activate minAllowed for cloud-controller-manager VPA

MartinWeindel:

/area control-plane
/area auto-scaling
/kind enhancement
/priority normal
/platform aws

To avoid OOM killing of the AWS cloud controller manager it is needed to activate `minAllowed` settings in its VPA object.

Fixes #

**Release note**:

```other operator
The cloud-controller-manager VPA does no specify minAllowed values to prevent too low resource recommendations from VPA that lead to OOM.
```

### DIFF
--- a/charts/internal/seed-controlplane/charts/cloud-controller-manager/templates/vpa.yaml
+++ b/charts/internal/seed-controlplane/charts/cloud-controller-manager/templates/vpa.yaml
@@ -10,3 +10,9 @@ spec:
     name: cloud-controller-manager
   updatePolicy:
     updateMode: Auto
+  resourcePolicy:
+    containerPolicies:
+    - containerName: '*'
+      minAllowed:
+        cpu: 20m
+        memory: 40M


charts/gardener-extension-provider-aws/values.yaml
```diff
@@ -10,10 +10,10 @@ maxSurge: 50%
 resources: {}
 vpa:
   enabled: true
-# resourcePolicy:
```
ialidzhikov, line 13: @MartinWeindel , these are the VPA values for the extension itself. If you want to adapt the values for the CCM (as stated in the PR description), this would be the right place - https://github.com/gardener/gardener-extension-provider-aws/blob/master/charts/internal/seed-controlplane/charts/cloud-controller-manager/templates/vpa.yaml.

ialidzhikov (changes requested):

charts/gardener-extension-provider-aws/values.yaml
MartinWeindel: It was too busy yesterday, thanks. I will fix it.

rfranzke (approved):
/lgtm

ialidzhikov (approved):
/lgtm

PR #290, opened 2021-03-05
Skip explicit LB deletion for Kubernetes >= v1.16

rfranzke:
**How to categorize this PR?**

/area ops-productivity robustness scalability
/kind enhancement
/priority critical
/platform aws

**What this PR does / why we need it**:
We are now skipping the explicit LB deletion when the shoot is of at least Kubernetes v1.16

See also: https://github.com/gardener/gardener/issues/129

**Release note**:

```other user
The load balancers and security groups are no longer explicitly deleted by the AWS provider extension when a shoot cluster of at least Kubernetes v1.16 is being deleted. Instead, it now relies on the service-controller in the `cloud-controller-manager` to properly clean up.
```

PR #291, opened 2021-03-05
[release-v1.20] Skip explicit LB deletion for Kubernetes >= v1.16

rfranzke:
Cherry-pick of #290.

**Release note**:

```other user
The load balancers and security groups are no longer explicitly deleted by the AWS provider extension when a shoot cluster of at least Kubernetes v1.16 is being deleted. Instead, it now relies on the service-controller in the `cloud-controller-manager` to properly clean up.
```


ialidzhikov (approved):
/lgtm

PR #293, opened 2021-03-10
Add controlplane label to csi-driver-controller

ialidzhikov:
`csi-driver-controller` Deployment is now annotated with the controlplane labels to allow the `dependency-watchdog` to restart the csi-driver-controller when the apiserver recovers.
Ref https://github.com/gardener/gardener/blob/v1.18.0/docs/usage/seed_bootstrapping.md#endpoint

**Release note**:

```other operator
NONE
```

### DIFF
--- a/charts/internal/seed-controlplane/charts/csi-driver-controller/templates/csi-driver-controller-deployment.yaml
+++ b/charts/internal/seed-controlplane/charts/csi-driver-controller/templates/csi-driver-controller-deployment.yaml
@@ -28,6 +28,8 @@ spec:
       labels:
         app: csi
         role: controller
+        garden.sapcloud.io/role: controlplane
+        gardener.cloud/role: controlplane
         networking.gardener.cloud/to-dns: allowed
         networking.gardener.cloud/to-public-networks: allowed
         networking.gardener.cloud/to-shoot-apiserver: allowed


charts/internal/seed-controlplane/charts/csi-driver-controller/templates/csi-driver-controller-deployment.yaml
```diff
@@ -28,6 +28,8 @@ spec:
       labels:
         app: csi
         role: controller
+        garden.sapcloud.io/role: controlplane
```
vpnachev, line 31: What do you think about using 
```yaml
              matchExpressions:
              - key: networking.gardener.cloud/to-shoot-apiserver
                operator: In
                values:
                - allowed
```

instead of
https://github.com/gardener/gardener/blob/3cbde2d2dc25fbad2916682966b18f988fe57b77/charts/seed-bootstrap/charts/dependency-watchdog/templates/endpoint-configmap.yaml#L16-L20 ?

At least for some time, till all know extensions migrate from `garden.sapcloud.io/role` to `gardener.cloud/role`.

The reason is that components that must talk with the shoot's API already have this label and it can be used by the DWD to easily find them.

vpnachev (approved):
/lgtm

rfranzke (approved):

PR #297, opened 2021-03-22
Vendor gardener/gardener@v1.19.0

ialidzhikov:
**Release note**:

```other developer
`github.com/gardener/gardener` dependency is now updated to `v1.19.0`. For the complete list of changes, see the [release notes](https://github.com/gardener/gardener/releases/tag/v1.19.0).
```


pkg/controller/controlplane/valuesprovider.go
```diff
@@ -370,6 +370,16 @@ func (vp *valuesProvider) GetControlPlaneShootChartValues(
...
+// GetControlPlaneShootCRDsChartValues returns the values for the control plane shoot CRDs chart applied by the generic actuator.
+// Currently the provider extension does not specify a control plane shoot CRDs chart. That's why we simply return empty values.
+func (vp *valuesProvider) GetControlPlaneShootCRDsChartValues(
+	_ context.Context,
+	_ *extensionsv1alpha1.ControlPlane,
+	_ *extensionscontroller.Cluster,
+) (map[string]interface{}, error) {
+	return map[string]interface{}{}, nil
+}
+
```
rfranzke, line 382: Hm, probably this should be added to https://github.com/gardener/gardener/blob/master/extensions/pkg/controller/controlplane/genericactuator/noopvaluesprovider.go, or?

rfranzke (approved):
/lgtm

pkg/controller/controlplane/valuesprovider.go
ialidzhikov: Nice catch, thank you! It seems that I missed this in the initial PR in the extension library. I now opened https://github.com/gardener/gardener/pull/3755.

PR #294, opened 2021-03-17
[ci:component:github.com/gardener/terraformer:v2.3.0->v2.4.0]

gardener-robot-ci-2:
**Release Notes*:
``` other operator github.com/gardener/terraformer #84 @ialidzhikov
The following terraform provider plugins are updated:
- hashicorp/terraform-provider-aws: 3.18.0 -> 3.32.0
- hashicorp/terraform-provider-google: 3.27.0 -> 3.59.0
- hashicorp/terraform-provider-google-beta: 3.27.0 -> 3.59.0
```

PR #301, opened 2021-03-29
Revert "Upgrade terraformer #294"

ialidzhikov:
/kind bug

This reverts commit c57ddb75ab948788b13474d9c36d3d85137363d8.

The new version of terraformer comes with a new version of terraform-provider-aws - ref https://github.com/gardener/terraformer/pull/84. The new version of terraform-provider-aws requires permissions for `iam:ListRolePolicies` which is currently not part of our officially recommended IAM policy. So it is incompatible change:

```
* reading inline policies for IAM role shoot--foo--bar-nodes, error: AccessDenied: User: technical-user is not authorized to perform: iam:ListRolePolicies on resource: role shoot--foo--bar-nodes
	status code: 403, request id: <omitted>
```

**Release note**:

```bugfix user
An issue causing Infrastructure reconciliation to fail because of insufficient privileges is now fixed.
```

### DIFF
--- a/charts/images.yaml
+++ b/charts/images.yaml
@@ -2,7 +2,7 @@ images:
 - name: terraformer
   sourceRepository: github.com/gardener/terraformer
   repository: eu.gcr.io/gardener-project/gardener/terraformer-aws
-  tag: "v2.4.0"
+  tag: "v2.3.0"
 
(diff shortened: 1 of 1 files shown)


vpnachev (approved):
/lgtm

PR #311, opened 2021-04-05
Fix the infrastructure-test TestDefinition

ialidzhikov:
/kind bug

After https://github.com/gardener/gardener-extension-provider-aws/pull/307, the test execution fails with:

```
# github.com/gardener/gardener-extension-provider-aws/test/integration/infrastructure
vendor/github.com/gardener/gardener/pkg/operation/botanist/component/extensions/operatingsystemconfig/downloader/downloader.go:19:2: cannot find package "." in:
	/go/src/github.com/gardener/gardener-extension-provider-aws/vendor/embed
FAIL	github.com/gardener/gardener-extension-provider-aws/test/integration/infrastructure [setup failed]
FAIL
```

The embed pkg is only present in go@1.16, so we should also update the golang version in the TestDefinition.

**Release note**:

```other operator
NONE
```

### DIFF
--- a/.test-defs/infrastructure-test.yaml
+++ b/.test-defs/infrastructure-test.yaml
@@ -16,4 +16,4 @@ spec:
     --secret-access-key=$SECRET_ACCESS_KEY
     --region=$REGION
 
-  image: golang:1.15.3
+  image: golang:1.16.3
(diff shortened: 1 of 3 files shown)


vpnachev (approved):
Thanks!
/lgtm

PR #312, opened 2021-04-08
Increase revisionHistoryLimit

timebertt:

/area ops-productivity
/kind enhancement
/platform aws

This PR increases `revisionHistoryLimit` to `5` for the `admission-aws` and `provider-aws` Deployments, and to `1` for all Deployments in the shoot control plane.
With `revisionHistoryLimit` set to `0`, operators often lack information about why a certain deployment was rolled.

Ref https://github.com/gardener/gardener/issues/3325

**Release note**:

```other operator
The `revisionHistoryLimit` of different `Deployment`s was increased.
```


charts/gardener-extension-admission-aws/charts/runtime/templates/deployment.yaml
```diff
@@ -6,7 +6,7 @@ metadata:
   labels:
 {{ include "labels" . | indent 4 }}
 spec:
-  revisionHistoryLimit: 0
+  revisionHistoryLimit: 5
```
rfranzke, line 9: Why 5 and not 1 (like for all others)? Do we have to expect that many revisions so that 5 is indeed needed?

timebertt: Hmm, no not really. I just set it to 5, because I already had a few occasions where I was still lacking some information with `revisionHistoryLimit=1` because a Deployment was rolled multiple times over an hour or so and when I was trying to investigate the history I was looking for was already purged.

Do you see any problems with setting it to 5? These Deployments are singletons per Garden/Seed cluster, so scalability shouldn't be a concern as for per-controlplane components.

rfranzke: No, I was just curious about the reason.

timebertt: Now you know 😉

vpnachev (approved):
/lgtm

PR #315, opened 2021-04-09
Support for Kubernetes v1.21

rfranzke:

/area open-source usability
/kind enhancement
/priority 3
/platform aws
/exp intermediate
/topology garden seed shoot

This PR adds support for 1.21 to the extension.

Part of gardener/gardener#3851

* :white_check_mark: Depends on gardener/gardener#3860
* I have successfully validated the functionality as follows:
  * :white_check_mark: Create new clusters with versions < 1.21
  * :white_check_mark: Create new clusters with version  = 1.21
  * :white_check_mark: Upgrade old clusters from version 1.20 to version 1.21
  * :white_check_mark: Delete clusters with versions < 1.21
  * :white_check_mark: Delete clusters with version  = 1.21

**Release note**:

```feature user
The AWS extension does now support shoot clusters with Kubernetes version 1.21. You should consider the [Kubernetes release notes](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.21.md) before upgrading to 1.21. 
```


rfranzke:
/invite @dkistner @prashanth26 - maybe you can have a look here as well

dkistner (approved):
I played around with it. Cluster creation/deletion and cluster upgrade works as expected. Thank you.

/lgtm
/unassign

pkg/webhook/controlplane/ensurer.go
```diff
@@ -58,6 +59,18 @@ func (e *ensurer) InjectClient(client client.Client) error {
...
+func computeCSIMigrationCompleteFeatureGate(cluster *extensionscontroller.Cluster) (string, error) {
+	k8sGreaterEqual121, err := versionutils.CompareVersions(cluster.Shoot.Spec.Kubernetes.Version, ">=", "1.21")
+	if err != nil {
+		return "", err
+	}
+
+	if k8sGreaterEqual121 {
+		return "InTreePluginAWSUnregister", nil
+	}
+	return "CSIMigrationAWSComplete", nil
```
prashanth26, line 71: It would be better to have string constants here and reuse them across the package and tests.

prashanth26 (approved):
Just a minor suggestion. Feel free to ignore it if you feel it's unnecesarry. 
/lgtm otherwise

pkg/webhook/controlplane/ensurer.go
rfranzke: Yeah, generally I agree, however, in this case the only usage of those names are in this file, and this is even the only occurrence. For the tests I explicitly don't want to reuse the constants to be able to detect issues in case someone (accidentally?) changes a constant. Also, for the other provider's it wasn't done like this, so I suggest to proceed as is :)

prashanth26: >  in this case the only usage of those names are in this file, and this is even the only occurrence

Agreed. That's why i kept it optional. 
/lgtm otherwise 👍

rfranzke:
Thanks @dkistner and @prashanth26!!!

PR #316, opened 2021-04-09
Replace packr with native go:embed

rfranzke:

/area dev-productivity
/kind cleanup
/platform aws

Meanwhile, we have updated to Golang 1.16 and can use the native `go:embed` feature which allows us to eliminate packr.

**Release note**:

```other operator
NONE
```


charts/images.go
```diff
@@ -0,0 +1,21 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package charts
+
+import _ "embed"
+
+// ImagesYAML contains the content of the images.yaml file
+//go:embed images.yaml
```
ialidzhikov, line 20: Is it possible to move this file under `pkg/imagevector` and to have `//go:embed ../../charts/images.yaml`? It is a little bit strange to have `.go` file under `charts/`.

rfranzke: I think it's not possible to use relative paths with `go:embed`, unfortunately.

From https://golang.org/pkg/embed/#hdr-Directives:

> The patterns are interpreted relative to the package directory containing the source file. The path separator is a forward slash, even on Windows systems. Patterns may not contain ‘.’ or ‘..’ or empty path elements, nor may they begin or end with a slash.

ialidzhikov: Ah, ok, thanks for explaining.

ialidzhikov (approved):
/lgtm

PR #321, opened 2021-04-20
Use VPC ID from InfrastructureConfig if possible

rfranzke:

/area ops-productivity
/kind bug
/platform aws

Today, the explicit LB/SG deletion relies on the Terraform output variable for the VPC ID. However, in case an existing VPC is used (`.networks.vpc.id != nil`), during `terraform destroy`, the first successfully deleted resource by Terraform removes this output variable and never puts it back. If the destruction now fails and the deletion flow is re-invocated, the explicit LG/SG deletion is skipped and never executed again.

With this PR, we no longer rely on the Terraform output variable if `.networks.vpc.id != nil`. Hence, the explicit LB/SG deletion will be properly retried for each invocation as long as the Terraform job did not succeed. This helps to destroy LB/SG resources which were created after the first invocation.

/invite @timebertt 

**Release note**:

```bugfix operator
When deleting an `Infrastructure`, the explicit load balancer and security group deletion is now properly retried in case an existing VPC is used.
```

### DIFF
--- a/pkg/controller/infrastructure/actuator_delete.go
+++ b/pkg/controller/infrastructure/actuator_delete.go
@@ -51,7 +51,11 @@ func Delete(
 ) error {
 	infrastructureConfig := &awsapi.InfrastructureConfig{}
 	if _, _, err := decoder.Decode(infrastructure.Spec.ProviderConfig.Raw, nil, infrastructureConfig); err != nil {
-		return fmt.Errorf("could not decode provider config: %+v", err)
+		// If we cannot decode the provider config, e.g. due to the recently introduced strict mode (see
+		// https://github.com/gardener/gardener-extension-provider-aws/pull/307), we don't return here and just log the
+		// error message.
+		logger.Error(err, "could not decode provider config")
+		infrastructureConfig = nil
 	}
 
 	tf, err := newTerraformer(logger, restConfig, aws.TerraformerPurposeInfra, infrastructure)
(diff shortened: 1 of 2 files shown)


pkg/controller/infrastructure/actuator_delete.go
```diff
@@ -79,16 +87,23 @@ func Delete(
...
-						logger.Info("Skipping explicit AWS load balancer and security group deletion because not all variables have been found in the Terraform state.")
-						return nil
+				var vpcID string
+				if id := infrastructureConfig.Networks.VPC.ID; id != nil {
+					vpcID = *id
+				} else {
+					stateVariables, err := tf.GetStateOutputVariables(ctx, aws.VPCIDKey)
+					if err != nil {
+						if apierrors.IsNotFound(err) || terraformer.IsVariablesNotFoundError(err) {
+							logger.Info("Skipping explicit AWS load balancer and security group deletion because not all variables have been found in the Terraform state.")
```
timebertt, line 97: Can you move this log down to have a general log if the explicit LB deletion is skipped?
It's already hard to debug because the flow doesn't log its steps here, so it would be helpful to have that log in both cases (where `vpcID` is empty).

timebertt (changes requested):
Thanks for taking care!

pkg/controller/infrastructure/actuator_delete.go
```diff
@@ -79,18 +91,28 @@ func Delete(
...
-						return nil
+				var vpcID string
+
+				if infrastructureConfig != nil && infrastructureConfig.Networks.VPC.ID != nil {
+					vpcID = *infrastructureConfig.Networks.VPC.ID
+				} else {
+					stateVariables, err := tf.GetStateOutputVariables(ctx, aws.VPCIDKey)
+					if err == nil {
+						vpcID = stateVariables[aws.VPCIDKey]
+					} else if err != nil && !apierrors.IsNotFound(err) && !terraformer.IsVariablesNotFoundError(err) {
```
timebertt, line 102: Nit: redundant condition
```suggestion
					} else if !apierrors.IsNotFound(err) && !terraformer.IsVariablesNotFoundError(err) {
```

timebertt (approved):
Thanks!
You can decide, if you want to address the nit comment :)
/lgtm

timebertt (approved):
/lgtm

PR #325, opened 2021-04-26
Add integration test case for invalid credentials

timebertt:

/area usability
/kind test
/platform aws

This PR tests https://github.com/gardener/terraformer/pull/93 and https://github.com/gardener/gardener/pull/3950 by adding an integration test case for invalid credentials.
It creates an infrastructure resource but uses invalid credentials and ensures that the `ERR_INFRA_UNAUTHORIZED` is correctly detected from the termination message and that the infrastructure can still be deleted afterwards (as it's a no-op).

Ref https://github.com/gardener/terraformer/issues/67

**Release note**:

```other operator
NONE
```

### DIFF
--- a/test/integration/infrastructure/infrastructure_test.go
+++ b/test/integration/infrastructure/infrastructure_test.go
@@ -18,6 +18,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"flag"
 	"fmt"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


timebertt:
Dropped `replace g/g` and terraformer dev version.
/ready
/test

test/integration/infrastructure/infrastructure_test.go
```diff
@@ -250,6 +253,102 @@ var _ = Describe("Infrastructure tests", func() {
...
+				infra     *extensionsv1alpha1.Infrastructure
+			)
+
+			framework.AddCleanupAction(func() {
+				By("cleaning up namespace and cluster")
+				Expect(client.IgnoreNotFound(c.Delete(ctx, namespace))).To(Succeed())
+				Expect(client.IgnoreNotFound(c.Delete(ctx, cluster))).To(Succeed())
+			})
+
+			defer func() {
```
ialidzhikov, line 278: `defer` won't be called if the test execution is interrupted by SIGINT/SIGTERM. You could use `AddCleanupAction` instead to make sure that the test won't leak any infra if it is interrupted.

timebertt: Well, in this test we can be sure, that no infra is leaked as we pass invalid credentials.

As for the other test cases, we should actually also move the deletion logic to a `defer` and only do actual cleanup on AWS in a `CleanupAction` (see https://github.com/gardener/terraformer/pull/57 for more background on that).
However, I didn't tackle this topic in this PR.

ialidzhikov: > Well, in this test we can be sure, that no infra is leaked as we pass invalid credentials.

Ah, yep, I missed this point.

ialidzhikov (approved):
/lgtm

rfranzke (approved):

PR #332, opened 2021-05-06
Replace aws-infra Helm chart with native Go templates

rfranzke:

/area open-source
/kind cleanup
/platform aws

In order to prepare for Helm 3 and to go in line with gardener/gardener#2754, the `aws-infra` Helm chart is eliminated and replaced with native Golang templates.

See discussion in https://github.com/gardener/gardener/pull/3994#pullrequestreview-652016478 ff.

**Release note**:

```other operator
NONE
```


pkg/controller/infrastructure/terraform.go
```diff
@@ -0,0 +1,89 @@
...
+}`
+)
+
+// Helper functions for template rendering
+
+func joinQuotes(data []string) string {
+	var out string
+
+	for _, v := range data {
+		out += fmt.Sprintf("%q", v)
```
timebertt, line 69: Comma seems to be missing if I'm not mistaken.
```suggestion
		out += fmt.Sprintf("%q,", v)
```

Maybe we should add multiple ignored tags in the infra test to also test this case: https://github.com/gardener/gardener-extension-provider-aws/blob/67201b176e1460d5a51f44ada0e58011bd120536/test/integration/infrastructure/infrastructure_test.go#L515-L518
(or add a unit test for the template func)

timebertt (changes requested):
Very nice, thank you!
/test

pkg/controller/infrastructure/terraform.go
rfranzke: Thanks, very nice catch, fixed and test adapted.

timebertt (approved):
Thanks for addressing my suggestion!
/lgtm

Unfortunately, the `bastion-test` failed :(

PR #335, opened 2021-05-10
[GEP-15] fix wrong base64 encoding for bastion instance userdata

xrstf:

/area ops-productivity
/kind bug
/platform aws

The testcase wasn't representing what is actually stored in the Bastion. `[]byte` is already stored as base64.

**Release note**:

```other operator
NONE
```


rfranzke (approved):
/lgtm

PR #337, opened 2021-05-16
Use separate ManagedResource for ControlPlane CRDs

ialidzhikov:
/kind enhancement
/platform aws

This PR moves the volumesnapshot CRDs to separate ManagedResource to prevent race in the deletion of the `extension-controlplane-shoot` ManagedResource.

Part of gardener/gardener#2227

This PR relies on the `resources.gardener.cloud/mode=Ignore` functionality in gardener-resource-manager (see https://github.com/gardener/gardener-resource-manager/pull/118) to migrate CRDs from one ManagedResource to another. gardener-resource-manager with the corresponding functionality is present in Gardener >= v1.21.0.

_PS: Initially I had the idea to introduce a feature gate to do not enforce the strict version requirement to Gardener. But generally the feature gate is something that can be enabled/disabled and this brings much more complexity - i.e. the feature gate was enabled (the CRDs were migrated to separate ManagedResource), then the feature gate was disabled and this would require CRDs to be migrated once again to the old ManagedResource._

**Release note**:

```bugfix operator
provider-aws is now using a separate ManagedResource for ControlPlane CRDs (`volumesnapshot` related CRDs) that are installed in the Shoot cluster to separate the deletion of CRDs from the deletion of the RBAC for controller leader election.
```

```breaking operator
This version of provider-aws requires at least Gardener v1.21.0. Before upgrading to this version of provider-aws, make sure that you upgraded to at least Gardener v1.21.0.
```

### DIFF
--- a/charts/internal/shoot-crds/charts/volumesnapshots/templates/crd-volumesnapshotclasses.yaml
+++ b/charts/internal/shoot-crds/charts/volumesnapshots/templates/crd-volumesnapshotclasses.yaml
@@ -4,7 +4,7 @@ kind: CustomResourceDefinition
 metadata:
   name: volumesnapshotclasses.snapshot.storage.k8s.io
   annotations:
-    resources.gardener.cloud/mode: Ignore
+    resources.gardener.cloud/keep-object: "true"
 spec:
   additionalPrinterColumns:
     - JSONPath: .driver
--- a/charts/internal/shoot-crds/Chart.yaml
+++ b/charts/internal/shoot-crds/Chart.yaml
@@ -1,0 +1,4 @@
+apiVersion: v1
+description: A Helm chart for control plane CRDs resources that are installed in the Shoot cluster
+name: shoot-crds
+version: 0.1.0
--- a/charts/internal/shoot-crds/charts/volumesnapshots/Chart.yaml
+++ b/charts/internal/shoot-crds/charts/volumesnapshots/Chart.yaml
@@ -1,0 +1,4 @@
+apiVersion: v1
+description: A Helm chart for volumesnapshot CRDs.
+name: volumesnapshots
+version: 0.1.0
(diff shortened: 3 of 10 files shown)


charts/internal/shoot-crds/charts/volumesnapshots/templates/crd-volumesnapshotclasses.yaml
```diff
@@ -0,0 +1,76 @@
+---
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: volumesnapshotclasses.snapshot.storage.k8s.io
+  annotations:
+    resources.gardener.cloud/keep-object: "true"
+spec:
```
stoyanr, line 8: I noticed that `resources.gardener.cloud/keep-object` is set to `true` for `volumesnapshotclasses`, but not for the other 2 CRDs. Is there a special reason for this? I understand that it's "safe" to keep the volumesnapshotclasses (and hence the CRDs) as there are no provider resources backing them, but would it actually hurt if we deleted them? Is it perhaps to avoid another "race condition", e.g. `volumesnapshots` could not be deleted if their `volumesnapshotclass` was deleted?

stoyanr (approved):
/lgtm

Just a question for my own understanding.

charts/internal/shoot-crds/charts/volumesnapshots/templates/crd-volumesnapshotclasses.yaml
ialidzhikov: > Is it perhaps to avoid another "race condition", e.g. volumesnapshots could not be deleted if their volumesnapshotclass was deleted?

Yep. Explained better in https://github.com/gardener/gardener-extension-provider-azure/pull/302.

rfranzke (approved):

PR #345, opened 2021-05-26
Expose logs from user exposed components in the new way

Kristian-ZH:

/area logging
/kind enhancement
/platform aws

Adapt the extension to https://github.com/gardener/gardener/pull/4060

Fixes #

**Release note**:

```other operator
NONE
```

### DIFF
--- a/charts/internal/seed-controlplane/charts/cloud-controller-manager/templates/configmap-observability.yaml
+++ b/charts/internal/seed-controlplane/charts/cloud-controller-manager/templates/configmap-observability.yaml
@@ -59,3 +59,8 @@ data:
   dashboard_users: |
     cloud-controller-manager-dashboard.json: |-
 {{- .Files.Get "ccm-dashboard.json" | nindent 6 }}
+
+  observedComponents: |
+    observedPods:
+    - podPrefix: cloud-controller-manager
+      isExposedToUser: true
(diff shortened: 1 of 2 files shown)


rfranzke:
This seems to not be backwards compatible with older Gardener versions, correct?

Kristian-ZH:
Yes, it will completely replace the older version

rfranzke:
I think @ialidzhikov asked this already as well. Generally, we try to be backwards-compatible for at least a couple of releases, so what's the proposed way forward with these PRs?

Kristian-ZH:
My idea to not make it backwards-compatible was that the users users/operators will have two different dashboards which will look in the same way and will contain the same information. Also currently we are making one dashboard for each exposed component, which makes the navigation between them a way harder. 

But if you see any value in keeping both approaches alive for some time then I will do it.

rfranzke:
My concern is rather: When somebody deploys an AWS extension version with this PR and a Gardener version without https://github.com/gardener/gardener/pull/4060 - what will happen? My understanding is that there won't be any dashboard at all, right?

Kristian-ZH:
Yes, if that happens there won't be any dashboards for ccm and csi-driver

rfranzke (approved):
/lgtm
/cc @ialidzhikov as he's currently a new release

ialidzhikov:
> Yes, if that happens there won't be any dashboards for ccm and csi-driver

Is this PR still a breaking change against a Gardener version without gardener/gardener#4060?

Kristian-ZH:
No, it is backwards compatible now.

PR #351, opened 2021-06-09
Enhance validation of cloud provider secrets

stoyanr:

/area usability
/area ops-productivity
/kind enhancement
/platform aws

Enhances the validation of AWS cloud provider secrets:
* `accessKeyID` must only contain alphanumeric characters, see https://docs.aws.amazon.com/IAM/latest/APIReference/API_AccessKey.html
* `secretAccessKey` must only contain base64 characters, see https://aws.amazon.com/blogs/security/a-safer-way-to-distribute-aws-credentials-to-ec2/

Fixes #

The above AWS security blog is not authoritative documentation. On the other hand, all `secretAccessKeys` we have seen so far do indeed contain only base64 characters. I think the proposed validation will in practice always match valid keys while it will also reject values that have e.g. leading or trailing whitespace or other non-base64 characters due e.g. copy / paste mistakes.

**Release note**:

```other operator
Validation of AWS cloud provider secrets is enhanced to reject `accessKeyID` that does not only contain alphanumeric characters, and `secretAccessKey` that does not only contain base64 characters.
```

### DIFF
--- a/pkg/apis/aws/validation/secrets.go
+++ b/pkg/apis/aws/validation/secrets.go
@@ -16,6 +16,7 @@ package validation
 
 import (
 	"fmt"
+	"regexp"
 
 	"github.com/gardener/gardener-extension-provider-aws/pkg/aws"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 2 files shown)


vpnachev (commented):
/lgtm
/needs second-opinion

vpnachev (approved):

rfranzke (approved):

Issue #234, opened 2020-12-08
Image overwrite for images defined by _images.yaml are not overwritten


schrodit:
**How to categorize this issue?**

/area delivery
/kind bug
/priority normal
/platform aws

**What happened**:

Currently images that are defined by the [_images.yaml](https://github.com/gardener/gardener-extension-provider-aws/blob/master/charts/gardener-extension-provider-aws/templates/_images.tpl) are added to the component descriptor (see the script [here](https://github.com/gardener/gardener/blob/master/hack/.ci/component_descriptor#L26)).

But these images are not overwritten by the image vector overwrite, therefore the opensource images are used instead of the overwritten image references.

cc @AndreasBurger @ccwienk 

**What you expected to happen**:

The images should be overwritten as defined by the component descriptor.

PR #358, opened 2021-06-15
make additional images used in the helm chart configurable

schrodit:

/area quality
/kind enhancement
/platform aws

This PR makes the additional images that are used in the aws-extension helm chart configurable by the `values.yaml`.
Also integrated the new configuration in the Landscaper component so that the images are overwritten with the value defined in the component descriptor.

The `_images.yaml` definition are kept for now so that the component-descriptor generation still works.

Fixes https://github.com/gardener/gardener-extension-provider-aws/issues/234

**Release note**:

```other operator
It is now possible to overwrite the `alpine` and `pause` container images that are used in the extension controller helm chart
```


rfranzke (approved):
/lgtm

PR #361, opened 2021-06-18
Switch landscaper deployment to ControllerDeployment

robertgraeff:

/area dev-productivity
/kind enhancement
/platform aws

This PR switches the landscaper based deployment to the new ControllerDeployment resource.

Furthermore, it adds some make targets to create and push component descriptors in a dev scenario.

Fixes #

**Release note**:

```other operator
NONE
```


Makefile
```diff
@@ -15,11 +15,12 @@
 EXTENSION_PREFIX            := gardener-extension
 NAME                        := provider-aws
 ADMISSION_NAME              := admission-aws
-REGISTRY                    := eu.gcr.io/gardener-project/gardener
```
vpnachev, line 18: Why do you remove this variable?
I would like to be able to easily re-build the images and push them to my own OCI registry, e.g.
```bash
$ make REGISTRY=my-regeistry docker-images
```

hack/create-installation.sh
```diff
@@ -0,0 +1,122 @@
+#!/bin/bash
+#
+# Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```
vpnachev, line 3:
```suggestion
# Copyright (c) 2021 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```

hack/create-installation.sh
```diff
@@ -0,0 +1,122 @@
+#!/bin/bash
```
vpnachev, line 1:
```suggestion
#!/usr/bin/env bash
```

hack/create-installation.sh
```diff
@@ -0,0 +1,122 @@
...
+          - classification: supported
+            cri:
+            - containerRuntimes:
+              - type: gvisor
+              name: containerd
+            version: 184.0.0
+        - name: flatcar
+          versions:
+          - classification: preview
+            version: 2605.11.0
```
vpnachev, line 58: Who will update these values?
Please, try to avoid landscape specification here in the provider repository.

hack/create-installation.sh
```diff
@@ -0,0 +1,122 @@
...
+        version: 1.20.6
+      - classification: deprecated
+        expirationDate: '2021-09-15T23:59:59Z'
+        version: 1.20.5
+      - classification: deprecated
+        expirationDate: '2021-08-15T23:59:59Z'
+        version: 1.20.4
+      - classification: deprecated
+        expirationDate: '2021-07-31T23:59:59Z'
+        version: 1.20.2
```
vpnachev, line 84: Same as above, those values will need to be actively maintained and will just introduce unwanted effort.

hack/create-installation.sh
```diff
@@ -0,0 +1,122 @@
...
+        - name: cloud-controller-manager
+          repository: eu.gcr.io/sap-se-gcr-k8s-public/k8s_gcr_io/hyperkube
+          sourceRepository: github.com/kubernetes/kubernetes
+          tag: sha256:12d877b29fb0d7c0cb90f4e8de8a98bb644df623be39ce7bab5420d2e21c2edc
+          targetVersion: 1.15.12
+        - name: cloud-controller-manager
+          repository: eu.gcr.io/sap-se-gcr-k8s-public/k8s_gcr_io/hyperkube
+          sourceRepository: github.com/kubernetes/kubernetes
+          tag: sha256:ad97b353f1d8c37950248e1879d9ed48ce0eaeccba58d4d04f8f2a788c467ffe
+          targetVersion: 1.16.15
```
vpnachev, line 119: Also same as above. These values must not be here.

hack/generate-cd.sh
```diff
@@ -0,0 +1,81 @@
+#!/bin/bash
```
vpnachev, line 1:
```suggestion
#!/usr/bin/env bash
```

hack/generate-cd.sh
```diff
@@ -0,0 +1,81 @@
...
+#
+# Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```
vpnachev, line 3:
```suggestion
# Copyright (c) 2021 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```

hack/generate-cd.sh
```diff
@@ -0,0 +1,81 @@
...
+IMAGE_REGISTRY="$(${SOURCE_PATH}/hack/get-image-registry.sh)"
+CD_REGISTRY="$(${SOURCE_PATH}/hack/get-cd-registry.sh)"
+
+CA_PATH="$(mktemp -d)"
+TMP_COMPONENT_DESCRIPTOR_PATH="${CA_PATH}/component-descriptor.yaml"
+
+if ! which component-cli 1>/dev/null; then
+  echo -n "component-cli is required to generate the component descriptors"
+  echo -n "Trying to installing it..."
+  go get github.com/gardener/component-cli/cmd/component-cli
```
vpnachev, line 19: Move this to the `install-requirements` target in the Makefile.

vpnachev (commented):
Ok, this is the second PR that I am checking and I already see the point that most of the changes here needs to be put into a library and just imported in the repositories.

Makefile
achimweigel: I have removed that change.

achimweigel:
> just imported in the



> Ok, this is the second PR that I am checking and I already see the point that most of the changes here needs to be put into a library and just imported in the repositories.
> 
> /cla

Yes, we will do this in the near future. Currently we are adding this to 5 projects to find out how it differs, e.g. we identified already one project with no dependency to gardener so the link approach via the vendor folder does not work here.

hack/create-installation.sh
achimweigel: I have removed that

hack/create-installation.sh
achimweigel: I have removed that

hack/create-installation.sh
achimweigel: I have removed that

hack/generate-cd.sh
achimweigel: I'm not sure if we should do this because if we move the script to a central location do we know if all projects have some install-requirements target? And we have to adapt this for all other projects.

vpnachev:
> we identified already one project with no dependency to gardener so the link approach via the vendor folder does not work here.

It must not be https://github.com/gardener/gardener, it can be https://github.com/gardener/landscaper or some other repo, but for sure whatever can be implemented as library should be done so.

achimweigel:
Just moving the script to e.g. landscaper might also not help because the script itself references a script in gardener which computes the image overwrite vector. Currently we are working on replacing this script by a Landscaper component. 

I totally agree with you that we will provide a central script. But we want to wait until the current implementation is more mature. Probably we need changes and don't want to make a release in some central project and then upgrade the reference in all other projects. And we want to wait until the image overwrite vector computation is done by a Landscaper component.

achimweigel:
Hi  @vpnachev, 

should we perhaps just remove the changes in the makefile and the scripts and only keep the blueprint adaptions? We could add these later again when we have some central implementation wherever this will be.

vpnachev:
Makes sense.

achimweigel:
Hi @vpnachev, I have removed all critical files.

vpnachev (approved):
/lgtm
/needs second-opinion
from @gardener/landscaper-maintainers team.

/invite @gardener/landscaper-maintainers

schrodit (approved):
lgtm

PR #362, opened 2021-06-19
Add DNSRecord controller

stoyanr:

/area control-plane
/kind api-change
/kind enhancement
/platform aws

* Adds a AWS-specific implementation of a DNSRecord controller, see https://github.com/gardener/gardener/pull/4231.
* Vendors g/g v1.26.0-dev and k8s v0.21.1 (in a separate commit to make reviews easier)

Fixes #

Note the following important details:
* The provider type is `aws-route53` (like in `DNSProvider`), not `aws` (like in other controllers) for reasons of compatibility and ease of migration.
* The credentials secret keys are again the ones compatible with existing domain secrets, e.g. `AWS_ACCESS_KEY_ID` instead of `accessKeyID`.
* The functions `CreateOrUpdateDNSRecord` and `DeleteDNSRecord` return after the DNS record is created / updated / deleted and don't wait until the change has been synced to all nameservers. (/cc @MartinWeindel)

I tested creating, updating, and deleting DNS records with the controller locally and it seems to work as expected.

**Release note**:

```other operator
NONE
```


MartinWeindel:
I don't understand the rationale of this PR at all. I'm surprised to see the DNS functionality apparently distributed to multiple extensions. 
1. DNS server is now linked with the infrastructure, although there is no direct dependency between specification of DNS provider and infrastructure. Should the aws-provider now be deployed on all seeds, if the DNS is hosted on AWS? How do you deal if a Gardener installation uses netlify, cloudflare or infoblox as DNS provider? Do you want to create new infrastructure providers for this purpose? What happens if internal and external DNS domains are on different backends?
Additionally the Gardenlet now needs to watch what DNS providers it needs on a seed and needs to make sure, that the corresponding infrastructure controller is deployed as soon as the a hosted zone is used from DNS provider type which is not deployed currently.  
2. I don't understand how the `DNSProvider` can be dropped without creating a similar new specification. Even if by chance DNS domain and infrastructure are using the same backend/SDK, they are using **different accounts** practically all the time. This means you need different secrets for infrastructure and DNS hosted zone management. As the same DNSProvider is often used for the `external` DNS entry and services/ingresses via shoot-dns-service, you have a duplication of the secrets.
3. "DNS hosted zones are cached": This is ok for a draft, but this is not sufficient for production purposes. You have to use a cache with TTL and renew the cache periodically. Especially if end users use their own zones, you need to be aware of such new zones in a short time. 
4. "`spec.dns.domain` can't be managed by a custom provider anymore" Is this really your target? If yes, how do you deal with all the clusters using their own hosted zones for the external and ingress DNS currently?

stoyanr:
@MartinWeindel Thanks for your feedback! The design is a direct consequence of the proposal we discussed. To your questions.

> Should the aws-provider now be deployed on all seeds, if the DNS is hosted on AWS? 

Yes. We might optimize it so that only the DNS controller is started in this case. This can be done later as it's pure configuration.

> How do you deal if a Gardener installation uses netlify, cloudflare or infoblox as DNS provider? Do you want to create new infrastructure providers for this purpose? 

Yes, they will only contain a DNS controller.

> What happens if internal and external DNS domains are on different backends?

We will have both extensions installed on the corresponding seed.

> Additionally the Gardenlet now needs to watch what DNS providers it needs on a seed and needs to make sure, that the corresponding infrastructure controller is deployed as soon as the a hosted zone is used from DNS provider type which is not deployed currently.

Yes, and this is exactly the mechanism used for all other extensions (provider, runtime, network, etc.), and so we'll actually be happy to use it for DNS as well.

> I don't understand how the DNSProvider can be dropped without creating a similar new specification. Even if by chance DNS domain and infrastructure are using the same backend/SDK, they are using different accounts practically all the time. This means you need different secrets for infrastructure and DNS hosted zone management. As the same DNSProvider is often used for the external DNS entry and services/ingresses via shoot-dns-service, you have a duplication of the secrets.

We don't want the complication of DNS provider, in fact getting rid of it is one of our main motivations for this change. If different accounts are used, we will have different secrets for them, exactly as today. If the same DNS provider is used as you mention above (and only in this case), then we would have the same credentials in 2 different secrets - one created for the DNSProvider, and another one for the DNSRecord (so 1 secret more than today). We don't see this as much of an issue.

> "DNS hosted zones are cached": This is ok for a draft, but this is not sufficient for production purposes. 

In fact, we think it's better to get rid of caching entirely and minimize the need for getting the DNS hosted zones from route53 via other means. DNSRecords will only be reconciled when the shoot is created and when it's migrated to a different seed (similarly to other extension controllers, it uses the `gardener.cloud/operation` annotation to implement this restriction). Since the hosted zone never changes, once it's written to the status during the first reconciliation it can be used directly, which leaves us with one single getting of hosted zones per shoot for the shoot's lifetime. To minimize this even further, we will add the zone as an optional field to the DNSRecord spec and actually pass it for our own hosted zones. With all that, we believe the need for caching goes away entirely.

> "spec.dns.domain can't be managed by a custom provider anymore" Is this really your target? If yes, how do you deal with all the clusters using their own hosted zones for the external and ingress DNS currently?

I was thinking about deprecating `spec.dns.providers.primary: true` in favor of fields like `spec.dns.type`, `spec.dns.region`, `spec.dns.zone`, etc., similarly to e.g. `network` or `backup` - so it's still possible to use a user-specified type, zone, etc. for Gardener DNS records, but not via `spec.dns.providers`. We still see this as an option for a future version of the API, however for now we would like to avoid incompatible changes, so we will allow `spec.dns.providers.primary: true` and simply ignore the `domains` and `zones` in this case for Gardener in this case.

/cc @rfranzke @timebertt @plkokanov

stoyanr:
Corrected the PR (and also https://github.com/gardener/gardener/pull/4231) to reflect the discussion above, /cc @timebertt @MartinWeindel if you would like to take another look.

MartinWeindel (changes requested):
Found one problem in the findZoneForName function.
Additiionally, alias targets should also be supported for AWS to reduce costs.
See https://github.com/gardener/external-dns-management/issues/27
and [code for creating alias target records instead of A records](https://github.com/gardener/external-dns-management/blob/master/pkg/controller/provider/aws/aliastarget.go#L98-L115)

MartinWeindel:
> In fact, we think it's better to get rid of caching entirely and minimize the need for getting the DNS hosted zones from route53 via other means. DNSRecords will only be reconciled when the shoot is created and when it's migrated to a different seed (similarly to other extension controllers, it uses the gardener.cloud/operation annotation to implement this restriction). Since the hosted zone never changes, once it's written to the status during the first reconciliation it can be used directly, which leaves us with one single getting of hosted zones per shoot for the shoot's lifetime. To minimize this even further, we will add the zone as an optional field to the DNSRecord spec and actually pass it for our own hosted zones. With all that, we believe the need for caching goes away entirely.

Good idea

pkg/aws/client/client.go
```diff
@@ -418,6 +423,90 @@ func (c *Client) DeleteSecurityGroup(ctx context.Context, id string) error {
...
+
+func newChangeResourceRecordSetsInput(zoneId, action, name, recordType string, values []string, ttl int64) *route53.ChangeResourceRecordSetsInput {
+	var resourceRecords []*route53.ResourceRecord
+	for _, value := range values {
+		if recordType == route53.RRTypeTxt {
+			value = encloseInQuotes(value)
+		}
+		resourceRecords = append(resourceRecords, &route53.ResourceRecord{
+			Value: aws.String(value),
+		})
```
MartinWeindel, line 483: Load balancers in AWS are using domain names instead of IP addresses. Therefore `CNAME` records need to be supported (and converted into `ALIAS` target records typically). But as far as I can see, the implementation already supports it, as the record type is not validated. `CNAME` records can only have a single value, but an explicit validation is probably not needed as the AWS route53 service will report an error.

pkg/controller/dnsrecord/actuator.go
```diff
@@ -0,0 +1,145 @@
...
+		return zone, nil
+	}
+}
+
+func findZoneForName(zones map[string]string, name string) string {
+	longestZoneName, result := "", ""
+	for zoneName, zoneId := range zones {
+		if strings.HasSuffix(name, zoneName) && len(zoneName) > len(longestZoneName) {
+			longestZoneName, result = zoneName, zoneId
+		}
```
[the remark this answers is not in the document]
stoyanr: It will not (`strings.HasSuffix("d.c.b.a", "ccc.b.a") == false`). You probably mean "d.cccc.b.a" must not match "ccc.b.a", in which case I agree. Will use the propose matching function, thanks!

pkg/aws/client/client.go
stoyanr: Thanks a lot! So if I understand correctly, in this case:
* The `RecordType` of the DNSRecord in this case should be `CNAME`. We'll add explicit validation for only a single value in this case (as well as for other things, e.g. the supported values for `recordType`) to https://github.com/gardener/gardener/pull/4231 before merging it.
* Only for the `aws-route53` implementation, this can be converted to an "alias target" using the code you referred to above as an example. 

Question: What is the advantage of using "alias target" instead of a regular `CNAME`? Could we simply live with `CNAME` everywhere?

stoyanr: OK, I found https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/resource-record-sets-choosing-alias-non-alias.html and this somewhat answers my last question. Perhaps the important one is "Pricing for DNS queries" (e.g. alias records are "free" and CNAME are not). Are there any other considerations?

Can the hardcoded list https://github.com/gardener/external-dns-management/blob/2b72ef8cb031137d2945b8253a27dcb1042e733d/pkg/controller/provider/aws/aliastarget.go#L29 be avoided? How likely is it change and to have to be actively maintained?

MartinWeindel: You guessed right. The usage of alias was requested from the SAP Hana team to reduce costs for DNS queries.

We have found no other solution than to hardcode the list. There seems to be no API to fetch this list. We even considered scraping HTML from the page [Elastic Load Balancing endpoints and quotas](https://docs.aws.amazon.com/general/latest/gr/elb.html), but this is of course not robust. The assumption is that the zone IDs never change. If there are new AWS regions we may not be able to generate `ALIAS` as long as they are not appended to the list.

stoyanr: OK, thanks, will do as suggested then.

stoyanr: Done.

pkg/controller/dnsrecord/actuator.go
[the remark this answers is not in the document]
stoyanr: Done.

pkg/controller/dnsrecord/actuator.go
```diff
@@ -0,0 +1,145 @@
...
+	case dns.Spec.Zone != nil && *dns.Spec.Zone != "":
+		return *dns.Spec.Zone, nil
+	default:
+		// The zone is not specified in the resource status or spec. Try to determine the zone by
+		// getting all hosted zones of the account and searching for the longest zone name that is a suffix of dns.spec.Name
+		zones, err := awsClient.GetDNSHostedZones(ctx)
+		if err != nil {
+			return "", fmt.Errorf("could not get DNS hosted zones: %+v", err)
+		}
+		a.logger.Info("Got DNS hosted zones", "zones", zones)
```
timebertt, line 125: We probably want to have some ref to the dnsrecord object in all of those logs.

timebertt (commented):
Well, there is not much to comment on. Looks very good already :)
I didn't pay too much attention to the DNS details, as I'm not so experienced in this area as @MartinWeindel. The controller structure and so on looks very good.

Will play around with it and test it a bit next.

pkg/aws/secret.go
```diff
@@ -19,50 +19,79 @@ import (
...
+func ReadCredentialsSecret(secret *corev1.Secret, useDNSKeys bool) (*Credentials, error) {
 	if secret.Data == nil {
 		return nil, fmt.Errorf("secret does not contain any data")
 	}
 
-	accessKeyID, ok := secret.Data[AccessKeyID]
+	accessKeyIDKey, secretAccessKeyKey, regionKey := AccessKeyID, SecretAccessKey, Region
+	if useDNSKeys {
+		accessKeyIDKey, secretAccessKeyKey, regionKey = DNSAccessKeyID, DNSSecretAccessKey, DNSRegion
+	}
```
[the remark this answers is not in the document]
stoyanr: I could do it, but I would like to be sure that there is a use case for this. If gardener would never pass DNS secrets in the second vairant, then the added complexity is not worth it.

> I remember, there is a use case in gardener, where you specify a non-default domain in Shoot.spec.dns.domain but no custom dns provider (secret). In that case the shoots credentials are used. 

If I specify a non-default domain while having a `default-domain` secret I get an error during validation:
```
Error from server (BadRequest): error when creating "dev/shoot-i500152-gcp.yaml": shoot uses a default domain but does not match expected scheme: <shoot-name>.<project-name>.<default-domain>
```

And if I delete the `default-domain` secret, the shoot is accepted, but the reconciliation fails immediately with:
```
Could not initialize a new operation for Shoot reconciliation: unable to figure out which secret should be used for dns.
```

Without a `default-domain` secret, the only way I could reconcile such a shoot I found so far is to add a provider (with a secret) and set `primary` to `true`. So I don't think what you describe is actually possible. Could I be missing something?

Having in mind that:
* `external-dns-management` would so far only accept the "DNS format" of secrets. 
* gardener doesn't perform any mapping of secret keys (and I think it shouldn't, since it's provider-specific). 

The only reason to accept the other secret format would be to allow users who never used `external-dns-management` before to specify a secret in this format for a custom provider. I am sceptical that this justifies the increased complexity of supporting both secret variants in all providers (this is just one of them).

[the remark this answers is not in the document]
stoyanr: Note: I double checked, `external-dns-management` indeed supports the "other format" via "alternative keys", e.g.:
```
accessKeyID, err := c.GetRequiredProperty("AWS_ACCESS_KEY_ID", "accessKeyID")
```
Still, the above consideration still applies. Is there a use case in which this format is actually used currently? If not, are we sure we want to introduce one?

pkg/aws/client/client.go
```diff
@@ -418,6 +423,90 @@ func (c *Client) DeleteSecurityGroup(ctx context.Context, id string) error {
 	return ignoreNotFound(err)
 }
 
+// GetDNSHostedZones returns a map of all DNS hosted zone names mapped to their IDs.
```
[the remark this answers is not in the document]
stoyanr: Yes, you are right, will extract it.

pkg/controller/dnsrecord/actuator.go
stoyanr: Good point, will add it.

pkg/controller/dnsrecord/actuator.go
```diff
@@ -0,0 +1,142 @@
...
+}
+
+// Migrate migrates the DNSRecord.
+func (a *actuator) Migrate(ctx context.Context, dns *extensionsv1alpha1.DNSRecord, cluster *extensionscontroller.Cluster) error {
+	return nil
+}
+
+func (a *actuator) getZone(ctx context.Context, dns *extensionsv1alpha1.DNSRecord, awsClient awsclient.Interface) (string, error) {
+	switch {
+	case dns.Status.Zone != nil && *dns.Status.Zone != "":
```
timebertt, line 114: This means, once the zone in the status is set, there is no way to influence it anymore via the spec field.

Just to be on the safe side:
Is there a use case where we might want to prefer the spec over the status field? I could imagine that if you deleted the record and recreated the hosted zone in between reconciliations, the controller would not be able to recover from this even if you specify the new hosted zone in the spec. 
Though, I'm not sure if that's entirely correct, so I'm asking here :)

pkg/controller/dnsrecord/actuator.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+}
+
+func (a *actuator) InjectClient(client client.Client) error {
+	a.client = client
+	return nil
+}
+
+// Reconcile reconciles the DNSRecord.
+func (a *actuator) Reconcile(ctx context.Context, dns *extensionsv1alpha1.DNSRecord, cluster *extensionscontroller.Cluster) error {
```
timebertt, line 50: One thing I noticed during my tests:
The controller uses the default workqueue rate limiter, which means failures are retried very quickly.
E.g. when specifying the wrong zone ID, this will result in a bunch of requests in a few seconds.

We might want to
- a) use a rate limiter with different qps/baseDelay settings or
- b) set `RetryAfter` (or influence it somehow from the actuator) or
- c) try to detect such errors (e.g. reusing error code detection early on here in the actuator) and only retry a few times
- or a combination of those options

in order to not immediately exhaust the accounts rate limits on such configuration issues.

pkg/aws/secret.go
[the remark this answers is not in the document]
stoyanr: I double checked some more and found this code: https://github.com/gardener/gardener/blob/54930d430301a5f0aadfa5367e10012371c542d0/pkg/operation/shoot/shoot.go#L466-L475
Trying to test it now ...

pkg/controller/dnsrecord/actuator.go
```diff
@@ -0,0 +1,142 @@
...
+
+	// Update resource status
+	return extensionscontroller.TryUpdateStatus(ctx, retry.DefaultBackoff, a.client, dns, func() error {
+		dns.Status.Zone = &zone
+		return nil
+	})
+}
+
+// Delete deletes the DNSRecord.
+func (a *actuator) Delete(ctx context.Context, dns *extensionsv1alpha1.DNSRecord, cluster *extensionscontroller.Cluster) error {
```
timebertt, line 78: Ok, this might be a bit theoretical and I'm not sure, if we need to think about this case, but hear me out :)

If the record type changes for some reason (might be because the corresponding LB changes from IP to Hostname, or because the order in the values list is changed and now the first one isn't part of a canonical zone anymore, so we change from ALIAS to CNAME, or whatever else), we will orphan the old record in the infrastructure.

If this is indeed a problem, we might want to track the record type that was created in the infrastructure and accordingly a) deny switching the record type or b) delete the old entry before creating one of the new type, or something similar.

Actually, the same applies to a changing name field value (though this one should definitely not change, I suppose). So another option might be to make those fields immutable, though I'm not sure if that's easily achievable via CRDs or if we would need validating webhooks.

pkg/controller/dnsrecord/actuator.go
```diff
@@ -0,0 +1,142 @@
...
+
+	// Determine DNS hosted zone ID
+	zone, err := a.getZone(ctx, dns, awsClient)
+	if err != nil {
+		return err
+	}
+
+	// Create or update DNS record
+	ttl := extensionsv1alpha1helper.GetDNSRecordTTL(dns.Spec.TTL)
+	a.logger.Info("Creating or updating DNS record", "zone", zone, "name", dns.Spec.Name, "type", dns.Spec.RecordType, "values", dns.Spec.Values)
```
timebertt, line 65: I don't remember the details: is the region an important detail in Route53? If so, I might make sense to add a respective field to this log message.

I don't see anything about regions in the dashboard and its not "regionized", so I suppose it's not important. But I don't know the details about the gov cloud use case.

timebertt (commented):
Ok, finished testing it. Worked really great!

A few more question arose while playing around with it, added them here to make sure we don't miss any details.

pkg/aws/secret.go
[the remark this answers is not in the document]
timebertt: > I double checked some more and found this code: https://github.com/gardener/gardener/blob/54930d430301a5f0aadfa5367e10012371c542d0/pkg/operation/shoot/shoot.go#L466-L475
> Trying to test it now ...

Yep, exactly. Also had to dig a bit, but this is the use case that I remembered.
So, a non-default domain is used in conjunction with a primary dns provider that doesn't reference a secret.

pkg/controller/dnsrecord/actuator.go
stoyanr: I think this is really minor, but it would be easy to just reverse the 2 cases and make sure the spec takes precedence if it's specified. I will do it.

pkg/controller/dnsrecord/actuator.go
stoyanr: I don't think this is what will happen if the record type changes. If you specify a record with the correct name and wrong record type, you will get an error from route53 different from "not found". So the record will not be orphaned, instead the deletion will fail. Perhaps this is good enough for a "theoretical" issue like this?

Regarding the name, this should definitely not change. It should be made immutable (and perhaps the record type as well). I will check what is the best way to do it and add it to https://github.com/gardener/gardener/pull/4231.

pkg/controller/dnsrecord/actuator.go
stoyanr: I also don't think the region is important. It's needed to create the client, but apart from that it doesn't seem to make a difference. I think the same records will be managed if the region changes. So I would avoid making it too prominent.

pkg/controller/dnsrecord/actuator.go
stoyanr: OK, good point. I think b) is easiest since we have `ReconcileErr` to override the `RequeueAfter` value returned by the reconciler. We should perhaps set this to e.g. 30 seconds for all errors that are coming from the route53 API, so no sophisticated error detection is needed? 

Regarding "only retry a few times" this would mean saving the number of retries in the status. I would avoid it unless you think it's really important.

timebertt: > OK, good point. I think b) is easiest since we have `ReconcileErr` to override the `RequeueAfter` value returned by the reconciler. We should perhaps set this to e.g. 30 seconds for all errors that are coming from the route53 API, so no sophisticated error detection is needed?

Hmmm, if we use a constant backoff of `30s` this might even make more calls in the long run that an exponential backoff, right?
But I'm not too sure about what's the best option here. We can take this to our in-person discussion tomorrow.

> Regarding "only retry a few times" this would mean saving the number of retries in the status. I would avoid it unless you think it's really important.

I can't really judge as of now, as I don't have enough hands-on experience with those things. Maybe @MartinWeindel can already give an assessment based on his experiences.
But in any case, we can still introduce such mechanism later on, if it becomes necessary.

pkg/controller/dnsrecord/actuator.go
timebertt: Ok, good. Then let's keep it like it is.

pkg/aws/secret.go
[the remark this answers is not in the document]
stoyanr: Done. Tested, works the use case mentioned above, works fine.

pkg/controller/dnsrecord/actuator.go
stoyanr: Done.

pkg/aws/client/client.go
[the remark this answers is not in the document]
stoyanr: Done.

pkg/controller/dnsrecord/actuator.go
stoyanr: Done.

pkg/controller/dnsrecord/actuator.go
stoyanr: Implemented and tested it with a hardcoded requeueAfter of 30 seconds and a wrongly configured zone. The actual requests happen more often than 30 seconds in the beginning because gardenlet also retries. After gardenlet stops retrying, they continue happening every 30 seconds. Perhaps the value should be increased to 1 minute? I don't think it should be made configurable though, it's just a precaution against rate limits.

stoyanr:
@timebertt I believe I addressed all your feedback (in a separate commit).
/squash

pkg/controller/dnsrecord/actuator.go
stoyanr: > Hmmm, if we use a constant backoff of 30s this might even make more calls in the long run that an exponential backoff, right?

Yes, but I think rate limits are per unit of time, so this doesn't really matter? The important thing is not to have too many requests quickly in succession, especially if many shoots are reconciled at the same time with a wrong configuration. I am not really sure either, but as I said I would not start with some elaborate mechanism at this point just to mitigate this issue.

BTW, I think gardenlet retries mentioned above could be prevented by correctly classifying such errors as user errors. Right now this is not the case, so I am seeing more requests than optimal. I guess 30 seconds is perhaps ok in this case.

timebertt (commented):
Thanks for incorporating my suggestions!
Also no further feedback to this PR right now from my side :)

pkg/controller/dnsrecord/actuator.go
stoyanr: When thinking more about it, I think the one practical way to deal with the issue would be to combine b) and c) - detect configuration errors if at all possible (e.g. the error code `NoSuchHostedZone` can't be anything else but a configuration error), and for such errors return a really long `RequeueAfter`, e.g. 1 hour. For all other route53 errors, return e.g. 30 seconds or simply propagate the error as is.
We could also recognize rate limit errors (I believe this is the `ThrottlingException` code), and `RequeueAfter` 1 hour here as well (on errors), since if we are throttled it's obviously pointless to retry more often.

timebertt: True, all good what you say. 
However, I think it's enough to save these ideas for later, once we actually have the problems. The current approach is not the nicest, but is probably already enough.

pkg/controller/dnsrecord/actuator.go
stoyanr: > It seems the behavior depends on the record type ...

In fact I was wrong, the same happens when trying to switch from A to CNAME and vice versa - the update failed as I expected, but then the deletion of the DNSRecord succeeded (it failed on route53 but the error returned was "not found" and so ignored), and so the actual record in route53 was orphaned.

Unresolving this since it's not so trivial, we should at least give it some thought.

pkg/controller/dnsrecord/actuator.go
stoyanr: In fact, it seems the region is not used **at all**. I tried with regions `foo` and `bar` and everything still works ... Still, this is route53-specific behavior, so I believe we should keep the region in the API as it *might* be important on other infrastructures.

timebertt: Sounds good, thanks for double-checking.

MartinWeindel: The region **is used** in some special cases:
1. "Route 53 in the Beijing and Ningxia Regions: specify cn-northwest-1." see https://docs.aws.amazon.com/general/latest/gr/r53.html
2. we make use of regions  for NS2 (AWS regional API access). It also needs another endpoint: see https://github.com/gardener/external-dns-management/blob/master/pkg/controller/provider/aws/handler.go#L99-L101
/cc @danielfoehrKn

pkg/controller/dnsrecord/actuator.go
stoyanr: We discussed the issue and decided not to do any further changes. Instead, we should rely on gardener properly categorizing these errors as user errors via https://github.com/gardener/gardener/blob/677eaffc5acdf78d19f09566945f4e91d97da742/pkg/apis/core/v1beta1/helper/errors.go#L56-L64 (the regexes should be extended appropriately with the corresponding route53 errors, as well as those from other providers). In this case, the shoot will rather quickly become `Failed`, and our extension controller will stop reconciling its resources, which will prevent further requests to the API. In essence, we already have an appropriate mechanism to deal with this, and we should use it instead of introducing an a different one.

pkg/controller/dnsrecord/actuator.go
stoyanr: We discussed this issue and decided not to introduce more complexity at this stage, apart from making the `recordType` field also immutable and making sure that it's actually validated by Gardener.

stoyanr:
Added unit tests in a new commit. I consider this PR ready now except for the fact that it's based on https://github.com/gardener/gardener/pull/4231, which is still not merged, so keeping the Draft state until this is the case.

pkg/controller/dnsrecord/actuator.go
```diff
@@ -159,3 +169,14 @@ func findZoneForName(zones map[string]string, name string) string {
 	}
 	return result
 }
+
+func getRegion(dns *extensionsv1alpha1.DNSRecord, credentials *aws.Credentials) string {
+	switch {
+	case dns.Spec.Region != nil && *dns.Spec.Region != "":
+		return *dns.Spec.Region
+	case credentials.Region != nil:
```
timebertt, line 177: Nit
```suggestion
	case len(credentials.Region) > 0:
```

timebertt (commented):
Codewise lgtm, will test once https://github.com/gardener/gardener/pull/4312 is merged and vendored.

pkg/controller/dnsrecord/actuator.go
stoyanr: Done

stoyanr:
Now this is properly based on g/g and therefore can be merged if there is no further feedback. Removing `Draft`.

timebertt (approved):
Thank you very much!
Code-wise
/lgtm
now!

Won't been able to test it myself again, but my first tests were already very good and not much has changed since then.

pkg/aws/secret.go
```diff
@@ -19,50 +19,77 @@ import (
...
+	extensionscontroller "github.com/gardener/gardener/extensions/pkg/controller"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/utils/pointer"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 // GetCredentialsFromSecretRef reads the secret given by the the secret reference and returns the read Credentials
 // object.
-func GetCredentialsFromSecretRef(ctx context.Context, client client.Client, secretRef corev1.SecretReference) (*Credentials, error) {
+func GetCredentialsFromSecretRef(ctx context.Context, client client.Client, secretRef corev1.SecretReference, allowDNSKeys bool) (*Credentials, error) {
```
rfranzke, line 31: Why "allow"DNSKeys?

rfranzke (approved):
/lgtm

pkg/aws/secret.go
stoyanr: Because they can be either allowed or not. If they are allowed, then a secret containing them will be parsed successfully. If not, parsing such a secret will fail. The idea is, that a secret that contains DNS keys but is not used for DNS will still fail to be parsed (we don't want to suddenly allow a new set of secret keys for everything).

rfranzke: Makes sense, thank you!

PR #365, opened 2021-06-21
Setting CSI AWS topology label on machine deployment creation to allow scale-up

### LINKED ISSUE #364: Scale from zero fails to work on k8s >= 1.18 AWS

/area auto-scaling
/kind enhancement
/priority 3
/platform aws

AWS CSI driver specific label `topology.ebs.csi.aws.com/zone` needs to be added while creation of machineDeployment . This provider specific label needs to be added by extension-provider-aws until the aws-ebs-csi-driver team fixes and starts using the well known label `topology.kubernetes.io/zone`

**Why is this needed**:

AWS starts using CSI volume provisioners from k8s>=1.18 and the PV needs csi driver specific label to mount to node rather than the well known `topology.kubernetes.io/zone`label. This leads to cluster Autoscaler not scaling up the node from zero in worker pool as the csi driver specific label is not added by anyone.

himanshu-kun:

/area auto-scaling
/kind enhancement
/platform aws

adds driver specific CSI label `topology.ebs.csi.aws.com/zone` during machineDeployment creation 
Fixes #364 

also made changes to unit test written for machineDeployment generation by adding relevant variables
**Release note**:

```other operator
AWS CSI driver specific topology label  "topology.ebs.csi.aws.com/zone" is set as a label on machine deployments to allow the proper working of cluster-autoscaler during scale-up. 
```

### DIFF
--- a/pkg/controller/worker/helper.go
+++ b/pkg/controller/worker/helper.go
@@ -28,10 +28,6 @@ import (
 	"k8s.io/client-go/util/retry"
 )
 
-const (
-	awsCSIDriverTopologyKey = "topology.ebs.csi.aws.com/zone"
-)
-
 func (w *workerDelegate) decodeWorkerProviderStatus() (*api.WorkerStatus, error) {
 	workerStatus := &api.WorkerStatus{}
 
@@ -62,22 +58,4 @@ func (w *workerDelegate) updateWorkerProviderStatus(ctx context.Context, workerS
 		w.worker.Status.ProviderStatus = &runtime.RawExtension{Object: workerStatusV1alpha1}
 		return nil
 	})
-}
-
-func makeCopyOfMap(mp map[string]string) map[string]string {
-	var mpDeepCopy = make(map[string]string)
-	for k, v := range mp {
-		mpDeepCopy[k] = v
-	}
-	return mpDeepCopy
-}
-
-func addAwsCsiDriverTopologyLabel(poolLabels map[string]string, zone string) map[string]string {
-	if _, exists := poolLabels[AwsCsiDriverTopologyKey]; !exists {
-		var poolLabelsDeepCopy = makeCopyOfMap(poolLabels)
-		poolLabelsDeepCopy[AwsCsiDriverTopologyKey] = zone
-		return poolLabelsDeepCopy
-	}
-
-	return poolLabels
 }
--- a/pkg/controller/worker/helper.go
+++ b/pkg/controller/worker/helper.go
@@ -29,7 +29,7 @@ import (
 )
 
 const (
-	AwsCsiDriverTopologyKey = "topology.ebs.csi.aws.com/zone"
+	awsCSIDriverTopologyKey = "topology.ebs.csi.aws.com/zone"
 )
 
 func (w *workerDelegate) decodeWorkerProviderStatus() (*api.WorkerStatus, error) {
--- a/pkg/controller/worker/helper.go
+++ b/pkg/controller/worker/helper.go
@@ -28,6 +28,10 @@ import (
 	"k8s.io/client-go/util/retry"
 )
 
+const (
+	AwsCsiDriverTopologyKey = "topology.ebs.csi.aws.com/zone"
+)
+
 func (w *workerDelegate) decodeWorkerProviderStatus() (*api.WorkerStatus, error) {
 	workerStatus := &api.WorkerStatus{}
 
@@ -58,4 +62,22 @@ func (w *workerDelegate) updateWorkerProviderStatus(ctx context.Context, workerS
 		w.worker.Status.ProviderStatus = &runtime.RawExtension{Object: workerStatusV1alpha1}
 		return nil
 	})
+}
+
+func makeCopyOfMap(mp map[string]string) map[string]string {
+	var mpDeepCopy = make(map[string]string)
+	for k, v := range mp {
+		mpDeepCopy[k] = v
+	}
+	return mpDeepCopy
+}
+
+func addAwsCsiDriverTopologyLabel(poolLabels map[string]string, zone string) map[string]string {
+	if _, exists := poolLabels[AwsCsiDriverTopologyKey]; !exists {
+		var poolLabelsDeepCopy = makeCopyOfMap(poolLabels)
+		poolLabelsDeepCopy[AwsCsiDriverTopologyKey] = zone
+		return poolLabelsDeepCopy
+	}
+
+	return poolLabels
 }
--- a/pkg/controller/worker/machines.go
+++ b/pkg/controller/worker/machines.go
@@ -24,7 +24,7 @@ import (
 	awsapi "github.com/gardener/gardener-extension-provider-aws/pkg/apis/aws"
 	awsapihelper "github.com/gardener/gardener-extension-provider-aws/pkg/apis/aws/helper"
 	"github.com/gardener/gardener-extension-provider-aws/pkg/aws"
-
+	"github.com/gardener/gardener/extensions/pkg/controller/csimigration"
 	"github.com/gardener/gardener/extensions/pkg/controller/worker"
 	genericworkeractuator "github.com/gardener/gardener/extensions/pkg/controller/worker/genericactuator"
 	v1beta1constants "github.com/gardener/gardener/pkg/apis/core/v1beta1/constants"
@@ -91,6 +91,11 @@ func (w *workerDelegate) generateMachineConfig() error {
 	}
 
 	nodesSecurityGroup, err := awsapihelper.FindSecurityGroupForPurpose(infrastructureStatus.VPC.SecurityGroups, awsapi.PurposeNodes)
+	if err != nil {
+		return err
+	}
+
+	csiEnabled, _, err := csimigration.CheckCSIConditions(w.cluster, aws.GetCSIMigrationKubernetesVersion(w.cluster))
 	if err != nil {
 		return err
 	}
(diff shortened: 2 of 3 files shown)


gardener-robot-ci-3:
Thank you @himanshu-kun for your contribution. Before I can start building your PR, a member of the organization must set the required label(s) {'reviewed/ok-to-test'}. Once started, you can check the build status in the PR checks section below.

ialidzhikov:
@himanshu-kun , which one of my comments in #364 is not clear?

pkg/controller/worker/machines_test.go
```diff
@@ -636,6 +647,9 @@ var _ = Describe("Machines", func() {
 					result, err := workerDelegate.GenerateMachineDeployments(ctx)
 					Expect(err).NotTo(HaveOccurred())
 					Expect(result).To(Equal(machineDeployments))
+					// klog.Errorf("%+v\n", result)
+					// klog.Errorf("\n")
+					// klog.Errorf("%+v\n", machineDeployments)
```
prashanth26, line 652:
```suggestion
```
Remove commented code.

pkg/controller/worker/machines.go
```diff
@@ -186,6 +187,11 @@ func (w *workerDelegate) generateMachineConfig() error {
 				MachineConfiguration: genericworkeractuator.ReadMachineConfiguration(pool),
 			})
 
+			//TODO : remove when AWS CSI driver stops using the aws csi topology key
+
+			//adding aws csi driver topology label if its not specified
```
prashanth26, line 192:
```suggestion
			// TODO: remove when AWS CSI driver stops using the aws csi topology key - <link-to-issue>
			// add aws csi driver topology label if its not specified
```
Can you please link the open issue regarding this on `aws-csi-driver` in this TODO comment?
Also please keep spaced-comments.

pkg/controller/worker/machines.go
```diff
@@ -186,6 +187,11 @@ func (w *workerDelegate) generateMachineConfig() error {
...
+			addAwsCsiDriverTopologyLabel(&machineDeploymentIndex, zone, pool.Labels, machineDeployments)
```
prashanth26, line 193: `zoneIndex` could've been used instead of creating `machineDeploymentIndex`. 

However, I feel that this `addAwsCsiDriverTopologyLabel` can be updated to get rid of passing machineDeployment and index info. You can rather keep it to just `addAwsCsiDriverTopologyLabel(pool.Labels, zone)` and invoke it directly while adding the label [here](https://github.com/gardener/gardener-extension-provider-aws/blob/58b5865d35527675548c2ef860b6e9381aa6ecfb/pkg/controller/worker/machines.go#L184), similar to [this](https://github.com/gardener/gardener-extension-provider-aws/blob/58b5865d35527675548c2ef860b6e9381aa6ecfb/pkg/controller/worker/machines.go#L181). This way you avoid passing the machineDeployment and tracking the index.

prashanth26 (commented):
Thanks for the PR @himanshu-kun . Some suggested changes.

pkg/controller/worker/machines.go
himanshu-kun: yes that would be much neater. Thanks ,will do that

himanshu-kun: find the changes to helper function 36619f9

pkg/controller/worker/machines_test.go
```diff
@@ -208,7 +212,11 @@ var _ = Describe("Machines", func() {
 				zone1 = region + "a"
 				zone2 = region + "b"
 
+				AwsCsiDriverTopologyKey = "topology.ebs.csi.aws.com/zone"
```
prashanth26, line 215: Can you reuse [this const](https://github.com/gardener/gardener-extension-provider-aws/pull/365/files#diff-b34b29ff5c37b47738c2ac7ef24c03eaca1a5c54e7635201747422c924661a32R32) here?

pkg/controller/worker/helper.go
```diff
@@ -59,3 +63,30 @@ func (w *workerDelegate) updateWorkerProviderStatus(ctx context.Context, workerS
...
+	return mpDeepCopy
+}
+
+// func addAwsCsiDriverTopologyLabel(machineDeploymentIndex *int, zone string, poolLabels map[string]string, machineDeployments worker.MachineDeployments) {
+// 	if _, exists := poolLabels[AwsCsiDriverTopologyKey]; !exists {
+// 		var poolLabelsDeepCopy = makeCopyOfMap(poolLabels)
+// 		poolLabelsDeepCopy[AwsCsiDriverTopologyKey] = zone
+// 		machineDeployments[*machineDeploymentIndex].Labels = poolLabelsDeepCopy
+// 		*machineDeploymentIndex++
+// 	}
```
prashanth26, line 81:
```suggestion
```
Always remove unwanted code. You can always look up past changes on your git commit history.

prashanth26 (commented):
Just minor comments. /lgtm otherwise.

pkg/controller/worker/helper.go
himanshu-kun: oh again :/ , sorry , I will keep it mind now

pkg/controller/worker/machines_test.go
```diff
@@ -911,3 +918,11 @@ func addNameAndSecretToMachineClass(class map[string]interface{}, name string, c
...
 	class["secret"].(map[string]interface{})["labels"] = map[string]string{v1beta1constants.GardenerPurpose: genericworkeractuator.GardenPurposeMachineClass}
 }
+
+func makeCopyOfMap(mp map[string]string) map[string]string {
+	var mpDeepCopy = make(map[string]string)
+	for k, v := range mp {
+		mpDeepCopy[k] = v
+	}
+	return mpDeepCopy
+}
```
himanshu-kun, line 928: this has been declared again in this file although it was declared in file `helper.go` as importing the package of `helper.go` makes lot of changes in code( the name of the imported package comes in front of many functions)

prashanth26: Aren't both of these in the same package? It should be accessible without any imports then?

himanshu-kun: no they aren't ...one is in `worker` package and the other is in `worker_test` package. and both the packages are in the `pkg/controller/worker`

prashanth26: Sorry i forgot about that. Isn't something like this possible then - https://github.com/gardener/gardener-extension-provider-aws/blob/8eb133cbb838bc2598865e7a288e6d2900e7278a/pkg/controller/worker/machines_test.go#L403.

pkg/controller/worker/helper.go
```diff
@@ -28,6 +28,10 @@ import (
 	"k8s.io/client-go/util/retry"
 )
 
+const (
+	AwsCsiDriverTopologyKey = "topology.ebs.csi.aws.com/zone"
```
rfranzke, line 32: why is this exported? If needed, add a doc string

pkg/controller/worker/helper.go
```diff
@@ -59,3 +63,21 @@ func (w *workerDelegate) updateWorkerProviderStatus(ctx context.Context, workerS
...
+
+func addAwsCsiDriverTopologyLabel(poolLabels map[string]string, zone string) map[string]string {
+	if _, exists := poolLabels[AwsCsiDriverTopologyKey]; !exists {
+		var poolLabelsDeepCopy = makeCopyOfMap(poolLabels)
+		poolLabelsDeepCopy[AwsCsiDriverTopologyKey] = zone
+		return poolLabelsDeepCopy
+	}
+
+	return poolLabels
+}
```
rfranzke, line 83: Why do we need a copy?

rfranzke (changes requested):

pkg/controller/worker/helper.go
himanshu-kun: The label is being used in `machine_test.go` as well. Sure I will add the docstring

rfranzke: If that's the only reason, don't use it in the test. Use the string in the test. This is what we try to do for most test packages. It will enable detecting (potentially undesired) changes of the constant's value.

pkg/controller/worker/helper.go
himanshu-kun: if I made the changes to pool.Labels directly , I would have to call another helper function to revert those changes as pool.Labels might be used somewhere else
I thought its better to make a copy in that case, even though copying will add extra time complexity

pkg/controller/worker/helper.go
himanshu-kun: sure, I will do that , thanks.

pkg/controller/worker/helper.go
rfranzke: > I would have to call another helper function to revert those changes as pool.Labels might be used somewhere else

I can't follow, can you explain again? Why do you need to revert the changes again? 😕

prashanth26: The pool.Labels are generic to that worker which might be hosting multiple zones. Hence he is making a copy to create one specific for the machineDeployment for a `zone`.

rfranzke: Hm, I see. You might want to use https://github.com/gardener/gardener/blob/b72ac4af0c9d00454be8b75eba00d7eaf8df8b48/pkg/utils/miscellaneous.go#L70-L91 to save some code, i.e.

```golang
				Labels:               utils.MergeMaps(pool.Labels, map[string]string{AwsCsiDriverTopologyKey: zone}),
```

WDYT?

prashanth26: I was wondering if there is already util for this earlier. However, I overlooked it. Good point, we can use this. WDYT @himanshu?

pkg/controller/worker/helper.go
vpnachev:
```suggestion
	awsCSIDriverTopologyKey = "topology.ebs.csi.aws.com/zone"
```

pkg/controller/worker/helper.go
```diff
@@ -59,3 +63,21 @@ func (w *workerDelegate) updateWorkerProviderStatus(ctx context.Context, workerS
...
+
+func makeCopyOfMap(mp map[string]string) map[string]string {
+	var mpDeepCopy = make(map[string]string)
+	for k, v := range mp {
+		mpDeepCopy[k] = v
+	}
+	return mpDeepCopy
+}
+
+func addAwsCsiDriverTopologyLabel(poolLabels map[string]string, zone string) map[string]string {
```
vpnachev, line 75: Please, add the csi labels only if the CSI is enabled for the cluster. 
You can see an example here https://github.com/gardener/gardener-extension-provider-aws/blob/0f39e3d1c1eed0075e3e5661793da74fa94feea1/pkg/webhook/controlplane/ensurer.go#L396 how to check whether it is enabled or not.

vpnachev (changes requested):

pkg/controller/worker/helper.go
himanshu-kun: thanks for that @vpnachev

pkg/controller/worker/helper.go
himanshu-kun: yeah thanks @rfranzke , it will save writing  helper functions, I will use it

himanshu-kun (commented):
I have removed the unit tests which I had added earlier in `machines_test.go` but have tested the changes on local AWS shoot.
For AWS cluster , CSI is enable from k8s>=1.18, so I have tested with local shoot of k8s=1.17.17 and k8s =1.20.2 and both gave the expected result of not adding and adding the csi label respectively.
Hope thats ok @vpnachev @rfranzke

pkg/controller/worker/machines.go
```diff
@@ -185,7 +191,11 @@ func (w *workerDelegate) generateMachineConfig() error {
 				Taints:               pool.Taints,
 				MachineConfiguration: genericworkeractuator.ReadMachineConfiguration(pool),
 			})
-
+			// TODO: remove when AWS CSI driver stops using the aws csi topology key - https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/899
+			// add aws csi driver topology label if its not specified
+			if csiEnabled {
+				machineDeployments[len(machineDeployments)-1].Labels = utils.MergeStringMaps(pool.Labels, map[string]string{awsCSIDriverTopologyKey: zone})
```
vpnachev, line 197: I do not like very much this line. 
You can firstly populate a machine deployment object and then append it to the list of all machine deployments. 
Other option is to have a helper (lambda) function that calculate the labels and instead of 
```golang
Labels: pool.Labels,
```

you can have something like 
```golang
Labels: func() map[string]string {
	if !csiEnabled {
		return pool.Labels
	}
	return utils.MergeStringMaps(pool.Labels, map[string]string{awsCSIDriverTopologyKey: zone})
}(),

```

himanshu-kun: find the changes here df1b320

vpnachev (commented):
/ok-to-test

Can you add a small test about this change in https://github.com/gardener/gardener-extension-provider-aws/blob/master/pkg/controller/worker/machines_test.go ?

pkg/controller/worker/machines.go
```diff
@@ -24,7 +24,7 @@ import (
 	awsapi "github.com/gardener/gardener-extension-provider-aws/pkg/apis/aws"
 	awsapihelper "github.com/gardener/gardener-extension-provider-aws/pkg/apis/aws/helper"
 	"github.com/gardener/gardener-extension-provider-aws/pkg/aws"
-
+	"github.com/gardener/gardener/extensions/pkg/controller/csimigration"
```
vpnachev, line 27:
```suggestion

	"github.com/gardener/gardener/extensions/pkg/controller/csimigration"
```

pkg/controller/worker/machines.go
```diff
@@ -168,19 +173,27 @@ func (w *workerDelegate) generateMachineConfig() error {
...
+				Maximum:        worker.DistributeOverZones(zoneIdx, pool.Maximum, zoneLen),
+				MaxSurge:       worker.DistributePositiveIntOrPercent(zoneIdx, pool.MaxSurge, zoneLen, pool.Maximum),
+				MaxUnavailable: worker.DistributePositiveIntOrPercent(zoneIdx, pool.MaxUnavailable, zoneLen, pool.Minimum),
+				// TODO: remove when AWS CSI driver stops using the aws csi topology key - https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/899
+				// add aws csi driver topology label if its not specified
+				Labels: func() map[string]string {
+					if !csiEnabled {
+						return pool.Labels
+					}
+					return utils.MergeStringMaps(pool.Labels, map[string]string{awsCSIDriverTopologyKey: zone})
```
vpnachev, line 195:
```suggestion
				Labels: func() map[string]string {
					if !csiEnabled {
						return pool.Labels
					}
					// TODO: remove the csi topology label when AWS CSI driver stops using the aws csi topology key - https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/899
					// add aws csi driver topology label if its not specified
					return utils.MergeStringMaps(pool.Labels, map[string]string{awsCSIDriverTopologyKey: zone})
```

vpnachev (changes requested):

rfranzke (commented):
Looks great now, just a unit test is missing as requested by @vpnachev

prashanth26:
/title  Setting CSI AWS topology label on machine deployment creation to allow scale-up

prashanth26 (approved):
Thanks for the changes and tests @himanshu-kun . 
/lgtm
/squash

rfranzke (approved):
/lgtm

vpnachev (approved):
/lgtm

vpnachev:
@prashanth26 @himanshu-kun please open a cherry-pick PR to version 1.25 with this change. 

You can see more details in https://github.com/gardener/gardener/blob/master/docs/development/process.md#cherry-picks and get some inspirations how to open the PR.

prashanth26:
> @prashanth26 @himanshu-kun please open a cherry-pick PR to version 1.25 with this change.
> 
> You can see more details in https://github.com/gardener/gardener/blob/master/docs/development/process.md#cherry-picks and get some inspirations how to open the PR.

@himanshu-kun get this done. Thanks.

PR #372, opened 2021-07-09
Document limitation about encryption with customer managed CMKs

ialidzhikov:
/area documentation

**Release note**:

```other operator
NONE
```

### DIFF
--- a/docs/usage-as-end-user.md
+++ b/docs/usage-as-end-user.md
@@ -271,6 +271,8 @@ spec:
         encrypted: true
 ```
 
+> Note: The AWS extension does not support EBS volume (root & data volumes) encryption with [customer managed CMK](https://docs.aws.amazon.com/kms/latest/developerguide/concepts.html#customer-cmk). Support for [customer managed CMK](https://docs.aws.amazon.com/kms/latest/developerguide/concepts.html#customer-cmk) is out of scope for now. Only [AWS managed CMK](https://docs.aws.amazon.com/kms/latest/developerguide/concepts.html#aws-managed-cmk) is supported.
+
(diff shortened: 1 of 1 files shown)


ialidzhikov:
To be honest, I don't have idea about KMS concepts. @prashanth26 , I tried to document the existing limitation (the status quo) based on your comment/analysis on existing issue. Let me know if I got something wrong or something needs to be adapted.

docs/usage-as-end-user.md
```diff
@@ -271,6 +271,8 @@ spec:
         encrypted: true
 ```
 
+> Note: The AWS extension does not support EBS volume (root & data volumes) encryption with [customer managed CMK](https://docs.aws.amazon.com/kms/latest/developerguide/concepts.html#customer-cmk). Only [AWS managed CMK](https://docs.aws.amazon.com/kms/latest/developerguide/concepts.html#aws-managed-cmk) is supported.
```
prashanth26, line 274:
```suggestion
> Note: The AWS extension does not support EBS volume (root & data volumes) encryption with [customer managed CMK](https://docs.aws.amazon.com/kms/latest/developerguide/concepts.html#customer-cmk). Only [AWS managed CMK](https://docs.aws.amazon.com/kms/latest/developerguide/concepts.html#aws-managed-cmk) is supported. Supporting customer-managed CMK would require extensive enhancements with gardener architecture by supporting storing/retrieving/rotation of these keys while managing such disks.
```

prashanth26 (commented):
@ialidzhikov - We had internally discussed this topic once in the past however haven't gone into the full details of it. However, it looks to me like an epic in itself and would require more time to implement something like this. cc: @amshuman-kr

prashanth26 (approved):

PR #379, opened 2021-07-16
Add dnsrecord integration test

stoyanr:

/area control-plane
/kind test
/platform aws

Adds an integration test for the `DNSRecord` controller.

Fixes #

**Release note**:

```other operator
NONE
```

### DIFF
--- a/.ci/testruns/default/templates/testrun.yaml
+++ b/.ci/testruns/default/templates/testrun.yaml
@@ -28,6 +28,18 @@ spec:
       secretKeyRef:
         name: shoot-operator-aws
         key: secretAccessKey
+  - name: DNS_ACCESS_KEY_ID
+    type: env
+    valueFrom:
+      secretKeyRef:
+        name: route53-operator-aws
+        key: accessKeyID
+  - name: DNS_SECRET_ACCESS_KEY
+    type: env
+    valueFrom:
+      secretKeyRef:
+        name: route53-operator-aws
+        key: secretAccessKey
 
   testflow:
   - name: infrastructure-test
--- a/.ci/testruns/default/templates/testrun.yaml
+++ b/.ci/testruns/default/templates/testrun.yaml
@@ -44,3 +44,10 @@ spec:
       - name: REGION
         type: env
         value: eu-west-1
+  - name: dnsrecord-test
+    definition:
+      name: dnsrecord-test
+      config:
+      - name: USE_EXISTING_CLUSTER
+        type: env
+        value: "true"
--- a/test/integration/dnsrecord/dnsrecord_test.go
+++ b/test/integration/dnsrecord/dnsrecord_test.go
@@ -1,0 +1,488 @@
+// Copyright (c) 2021 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package dnsrecord_test
+
+import (
+	"context"
+	"flag"
+	"fmt"
+	"path/filepath"
+	"strings"
+	"time"
+
+	"github.com/aws/aws-sdk-go/service/route53"
+	awsinstall "github.com/gardener/gardener-extension-provider-aws/pkg/apis/aws/install"
+	"github.com/gardener/gardener-extension-provider-aws/pkg/aws"
+	awsclient "github.com/gardener/gardener-extension-provider-aws/pkg/aws/client"
+	dnsrecordctrl "github.com/gardener/gardener-extension-provider-aws/pkg/controller/dnsrecord"
+	v1beta1constants "github.com/gardener/gardener/pkg/apis/core/v1beta1/constants"
+
+	extensionsv1alpha1 "github.com/gardener/gardener/pkg/apis/extensions/v1alpha1"
+	"github.com/gardener/gardener/pkg/extensions"
+	gardenerutils "github.com/gardener/gardener/pkg/utils"
+	"github.com/gardener/gardener/test/framework"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	. "github.com/onsi/gomega/gstruct"
+	"github.com/sirupsen/logrus"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/pointer"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/envtest"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+)
+
+var (
+	accessKeyID     = flag.String("access-key-id", "", "AWS access key id")
+	secretAccessKey = flag.String("secret-access-key", "", "AWS secret access key")
+)
+
+func validateFlags() {
+	if len(*accessKeyID) == 0 {
+		panic("need an AWS access key id")
+	}
+	if len(*secretAccessKey) == 0 {
+		panic("need an AWS secret access key")
+	}
+}
+
+var _ = Describe("DNSRecord tests", func() {
+	var (
+		ctx = context.Background()
+
+		logger    *logrus.Entry
+		awsClient *awsclient.Client
+		testEnv   *envtest.Environment
+		mgrCancel context.CancelFunc
+		c         client.Client
+
+		testName = fmt.Sprintf("aws-dnsrecord-it--%s", randomString())
+		zoneName = testName + ".gardener.cloud"
+		zoneID   string
+
+		namespace = &corev1.Namespace{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: testName,
+			},
+		}
+		secret = &corev1.Secret{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "dnsrecord",
+				Namespace: testName,
+			},
+			Data: map[string][]byte{
+				aws.AccessKeyID:     []byte(*accessKeyID),
+				aws.SecretAccessKey: []byte(*secretAccessKey),
+			},
+		}
+		cluster = &extensionsv1alpha1.Cluster{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: testName,
+			},
+		}
+	)
+
+	BeforeSuite(func() {
+		repoRoot := filepath.Join("..", "..", "..")
+
+		// enable manager logs
+		logf.SetLogger(zap.New(zap.UseDevMode(true), zap.WriteTo(GinkgoWriter)))
+
+		log := logrus.New()
+		log.SetOutput(GinkgoWriter)
+		logger = logrus.NewEntry(log)
+
+		By("starting test environment")
+		testEnv = &envtest.Environment{
+			CRDInstallOptions: envtest.CRDInstallOptions{
+				Paths: []string{
+					filepath.Join(repoRoot, "example", "20-crd-dnsrecord.yaml"),
+					filepath.Join(repoRoot, "example", "20-crd-cluster.yaml"),
+				},
+			},
+		}
+
+		cfg, err := testEnv.Start()
+		Expect(err).ToNot(HaveOccurred())
+		Expect(cfg).ToNot(BeNil())
+
+		By("setting up manager")
+		mgr, err := manager.New(cfg, manager.Options{
+			MetricsBindAddress: "0",
+		})
+		Expect(err).ToNot(HaveOccurred())
+
+		Expect(extensionsv1alpha1.AddToScheme(mgr.GetScheme())).To(Succeed())
+		Expect(awsinstall.AddToScheme(mgr.GetScheme())).To(Succeed())
+
+		Expect(dnsrecordctrl.AddToManager(mgr)).To(Succeed())
+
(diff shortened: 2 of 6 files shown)


test/integration/dnsrecord/dnsrecord_test.go
```diff
@@ -0,0 +1,294 @@
...
+			Expect(client.IgnoreNotFound(c.Delete(ctx, cluster))).To(Succeed())
+			Expect(client.IgnoreNotFound(c.Delete(ctx, secret))).To(Succeed())
+			Expect(client.IgnoreNotFound(c.Delete(ctx, namespace))).To(Succeed())
+		})
+
+		By("creating dnsrecord")
+		dns := newDNSRecord(testName)
+		Expect(c.Create(ctx, dns)).To(Succeed())
+		framework.AddCleanupAction(func() {
+			By("deleting dnsrecord")
```
timebertt, line 193: Deletion logic should be executed in a `defer` block. Then in a separate cleanup action, the client should be used to clean up any orphaned resources. This avoid the need for manual cleanup, see https://github.com/gardener/terraformer/pull/57

test/integration/dnsrecord/dnsrecord_test.go
```diff
@@ -0,0 +1,294 @@
...
+		// enable manager logs
+		logf.SetLogger(zap.New(zap.UseDevMode(true), zap.WriteTo(GinkgoWriter)))
+
+		log := logrus.New()
+		log.SetOutput(GinkgoWriter)
+		logger = logrus.NewEntry(log)
+
+		By("starting test environment")
+		testEnv = &envtest.Environment{
+			UseExistingCluster: pointer.BoolPtr(true),
```
timebertt, line 116: This test could also run against a temp envtest cluster and we don't necessarily need an existing cluster, right?

test/integration/dnsrecord/dnsrecord_test.go
```diff
@@ -0,0 +1,294 @@
...
+		framework.AddCleanupAction(func() {
+			By("deleting dnsrecord")
+			Expect(client.IgnoreNotFound(c.Delete(ctx, dns))).To(Succeed())
+
+			By("waiting until dnsrecord is deleted")
+			Expect(extensions.WaitUntilExtensionObjectDeleted(
+				ctx,
+				c,
+				logger,
+				dns.DeepCopy(),
```
timebertt, line 201:
```suggestion
				dns,
```

test/integration/dnsrecord/dnsrecord_test.go
```diff
@@ -0,0 +1,294 @@
...
+	})
+})
+
+func newDNSRecord(namespace string) *extensionsv1alpha1.DNSRecord {
+	return &extensionsv1alpha1.DNSRecord{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "dnsrecord",
+			Namespace: namespace,
+		},
+		Spec: extensionsv1alpha1.DNSRecordSpec{
```
timebertt, line 241: Should we add another test case with `spec.zone` set?

test/integration/dnsrecord/dnsrecord_test.go
```diff
@@ -0,0 +1,294 @@
...
+		Expect(extensionsv1alpha1.AddToScheme(mgr.GetScheme())).To(Succeed())
+		Expect(awsinstall.AddToScheme(mgr.GetScheme())).To(Succeed())
+
+		Expect(dnsrecordctrl.AddToManager(mgr)).To(Succeed())
+
+		var mgrContext context.Context
+		mgrContext, mgrCancel = context.WithCancel(ctx)
+
+		By("starting manager")
+		go func() {
```
timebertt, line 144: See doc string of `GinkgoRecover`

```suggestion
		go func() {
		     defer GinkgoRecover()
```

timebertt (commented):
Some early feedback :)

test/integration/dnsrecord/dnsrecord_test.go
stoyanr: Without `DeepCopy()` the test fails. The method resets the resource and the subsequent `verify` doesn't have the right values, so it fails. Could perhaps be avoided in other ways, but the `DeepCopy` here is perhaps the most convenient one.

test/integration/dnsrecord/dnsrecord_test.go
stoyanr: This is still not finished (so `Draft`), I plan to add not one, but multiple "interesting" test cases.

test/integration/dnsrecord/dnsrecord_test.go
timebertt: Ah, got it :)

test/integration/dnsrecord/dnsrecord_test.go
stoyanr: Not sure, will try it out. But what's the problem to use an existing cluster? This can be really any cluster, including our Nodeless dev setup, and in CI such a cluster is already created anyway for the infra tests. The bastion tests use an existing cluster as well.

test/integration/dnsrecord/dnsrecord_test.go
stoyanr: Yes, I had similar thoughts but didn't come to it yet. This is still not finished (so `Draft`). BTW, this is all copy pasted from the other integration tests (bastion in this case, but infra is the same), so ideally it should be fixed there as well.

test/integration/dnsrecord/dnsrecord_test.go
stoyanr: Thanks, will take a look.

test/integration/dnsrecord/dnsrecord_test.go
timebertt: Yes, this should be fixed. Added a project card (https://github.com/orgs/gardener/projects/6#card-51424446) a long time ago, but didn't come to it yet.

test/integration/dnsrecord/dnsrecord_test.go
timebertt: There is no problem with it. But generally it would be possible. One can configure it via the env var `USE_EXISTING_CLUSTER`.
We could leave it off for the local case, so that one doesn't have to think about starting a cluster / pointing `KUBECONFIG` to it. In CI / TM we could specify `USE_EXISTING_CLUSTER`, if we see a benefit in it.

Just explaining the options... Feel free to choose what makes sense from your point of view.

stoyanr: I tried to set the above to `false`, but I got:
```
unable to start the controlplane	{"tries": 0, "error": "fork/exec /usr/local/kubebuilder/bin/etcd: no such file or directory"}
```
I guess I need to install `kubebuilder`, but then it should also be installed via `install requirements`, and this would introduce an important difference with `install requirements` for other provider extensions. And I remember from previous experience with SEN that installing / updating kubebuilder automatically wasn't a great experience.
Are you sure we really need this, or perhaps I am missing something?

stoyanr: In any case, you are right that this line should be removed, since there is the `USE_EXISTING_CLUSTER` variable that could be just set to true (and I can confirm this works as expected), I am just not really sure if I wanted to to use an existing cluster locally what are my options - just install kubebuilder locally (but not via `make install-requirements`)? Is this documented somewhere?

test/integration/dnsrecord/dnsrecord_test.go
stoyanr: Done

test/integration/dnsrecord/dnsrecord_test.go
stoyanr: Added a few more test cases that cover this and other interesting situations.

test/integration/dnsrecord/dnsrecord_test.go
stoyanr: Removed the line as suggested.

test/integration/dnsrecord/dnsrecord_test.go
stoyanr: Done

stoyanr:
The new integration test fails with a permission error. Trying to fix this, meanwhile I consider the PR as ready for review.
/invite @timebertt

.ci/testruns/default/templates/testrun.yaml
```diff
@@ -44,3 +44,10 @@ spec:
       - name: REGION
         type: env
         value: eu-west-1
+  - name: dnsrecord-test
+    definition:
+      name: dnsrecord-test
+      config:
+      - name: USE_EXISTING_CLUSTER
```
timebertt, line 51: Should this be moved to `.test-defs/dnsrecord-test.yaml`?

test/integration/dnsrecord/dnsrecord_test.go
timebertt: We don't need to install kubebuilder, `setup-envtest` is enough, see [Makefile](https://github.com/gardener/gardener/blob/3806079d4e35dca8064e39ce0e69a0912aa02778/Makefile#L177) and [`tools.go`](https://github.com/gardener/gardener/blob/3806079d4e35dca8064e39ce0e69a0912aa02778/hack/tools.go#L28) in g/g.
`setup-envtest` can then be used to pull envtest binaries (e.g. etcd). It's already done in [`test-integration.sh`](https://github.com/gardener/gardener/blob/master/hack/test-integration.sh), so you can basically replace the `go test` call in the Makefile here with the vendored hack script and then everything should directly work.

timebertt (commented):
This test is very nice! Well done :)
Just a few comments from my side. Haven't tried it so far because of the elevated permissions needed for it.

test/integration/dnsrecord/dnsrecord_test.go
timebertt: So basically something like this: https://github.com/gardener/gardener-extension-provider-aws/commit/0601dd8915c252efb9ad80bb322295195c7f171f

test/integration/dnsrecord/dnsrecord_test.go
```diff
@@ -0,0 +1,488 @@
...
+		validateFlags()
+
+		awsClient, err = awsclient.NewClient(*accessKeyID, *secretAccessKey, aws.DefaultDNSRegion)
+		Expect(err).NotTo(HaveOccurred())
+
+		By("setting up shoot environment")
+		setupShootEnvironment(ctx, c, namespace, secret, cluster)
+
+		By("creating AWS DNS hosted zone")
+		zoneID = createDNSHostedZone(ctx, awsClient, zoneName)
```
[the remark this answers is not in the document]
stoyanr: Because otherwise I would have to do either of the following:
* Expose one of our existing hosted zone names / IDs in the repo (not really sensitive perhaps, but still internal information), and I didn't want to do it. Also, the account that is configured to perform the integration test would still not have permissions to use my hardcoded zone.
* Introduce an additional TM configuration / secret (with help from the CI team, they would need to change something on their side) to host the zone name / ID. That would perhaps be possible, but I didn't want to do it, creating a temporary zone for the test is simpler from my perspective.

test/integration/dnsrecord/dnsrecord_test.go
```diff
@@ -0,0 +1,488 @@
...
+		ctx = context.Background()
+
+		logger    *logrus.Entry
+		awsClient *awsclient.Client
+		testEnv   *envtest.Environment
+		mgrCancel context.CancelFunc
+		c         client.Client
+
+		testName = fmt.Sprintf("aws-dnsrecord-it--%s", randomString())
+		zoneName = testName + ".gardener.cloud"
```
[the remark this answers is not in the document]
stoyanr: There is absolutely no issue with hardcoding it. You can create the zone "<id>.gardener.cloud" in any account, even duplicated names would be ok (although since the test ID is generated, no duplication is possible in this case). The records created are of course not resolvable, but our integration test doesn't really care. There is no real need to use an own domain / hosted zone, it simply doesn't matter.

.ci/testruns/default/templates/testrun.yaml
stoyanr: I am not really sure if it's even possible to specify an env variable in a `TestDefinition`. Even if it is, in g/g or extension repos this is right now never done. Instead, `TestDefinitions` only specify arguments but actual env variables are always specified in `Testruns`. Why should we deviate from this model here?

test/integration/dnsrecord/dnsrecord_test.go
stoyanr: Ah, got it, thanks!

stoyanr:
@timebertt The `dnsrecord-test` succeeded now.

test/integration/dnsrecord/dnsrecord_test.go
stoyanr: Added the changes you suggested in a separate commit to this PR. I can confirm everything works correctly now both with and without `USE_EXISTING_CLUSTER=true`.
/squash

.ci/testruns/default/templates/testrun.yaml
timebertt: My thinking behind the question was, that the `--kubeconfig` flag is also configured in the `TestDefinition`, which is basically where we want to say: "hey test, please use this existing cluster". So FMPOV, this is configuration that belongs together.
Thinking about it now, I guess both testrun and testdef are not used independently. So probably there's currently no value in separating cleanly here.

test/integration/dnsrecord/dnsrecord_test.go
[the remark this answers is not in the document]
timebertt: Ah got it. I wasn't aware, that you can create hosted zones with freely chosen names anywhere.
But now that you say it, it makes perfect sense 👍 
Then we can keep it of course.

test/integration/dnsrecord/dnsrecord_test.go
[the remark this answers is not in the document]
timebertt: Ok, you're right.
After sleeping over it and reading your reasoning, creating a temporary zone indeed sounds like a good idea 😉

timebertt (approved):
Love it :)
Executed the test locally and it worked great!

/lgtm

stoyanr:
@timebertt @MartinWeindel I have introduced a new credentials secret for the `dnsrecord-test` to make sure the user has the appropriate `route53` permissions without changing the permissions of `shoot-operator-aws`. I consider the test finished and ready to be merged.
It would be great to get an approval from you so that this PR could be merged.

MartinWeindel (approved):
/lgtm

timebertt (approved):
Thank you!
/lgtm

PR #386, opened 2021-08-05
Enable `srcAndDstChecksEnabled` field from k8s >= v1.22

AxiomSamarth:

/area networking
/kind enhancement
/platform aws

This PR introduces a new field in the `providerSpec` of `MachineClass` for AWS called `srcAndDstChecksEnabled ` which is `true` by default and when set to `false` will disable the source destination check on the AWS Instance (NAT instance in this case). This field shall be used by default from K8s >= v1.22

Fixes [#36](https://github.com/gardener/machine-controller-manager-provider-aws/issues/36)

**Release note**:

```feature developer
Introducing a new field in the `providerSpec` of `MachineClass` for AWS called `srcAndDstChecksEnabled ` which is `true` by default and when set to `false` will disable the source destination check on the AWS Instance 
```

### DIFF
--- a/pkg/controller/worker/machines.go
+++ b/pkg/controller/worker/machines.go
@@ -206,7 +206,7 @@ func (w *workerDelegate) generateMachineConfig() error {
 			machineClassSpec["secret"].(map[string]interface{})["labels"] = map[string]string{v1beta1constants.GardenerPurpose: genericworkeractuator.GardenPurposeMachineClass}
 
 			// if the shoot K8s version is greater than or equal to 1.22 then
-			// enable source destination check on AWS instances by default
+			// disable source destination check on AWS instances by default
 			shootK8sVersion := w.cluster.Shoot.Spec.Kubernetes.Version
 
 			k8sVersionGreaterOrEqualThan122, err := versionutils.CompareVersions(shootK8sVersion, ">=", "1.22")
--- a/pkg/controller/worker/machines.go
+++ b/pkg/controller/worker/machines.go
@@ -32,6 +32,7 @@ import (
 	extensionsv1alpha1 "github.com/gardener/gardener/pkg/apis/extensions/v1alpha1"
 	"github.com/gardener/gardener/pkg/client/kubernetes"
 	"github.com/gardener/gardener/pkg/utils"
+	versionutils "github.com/gardener/gardener/pkg/utils/version"
 	machinev1alpha1 "github.com/gardener/machine-controller-manager/pkg/apis/machine/v1alpha1"
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
@@ -203,6 +204,19 @@ func (w *workerDelegate) generateMachineConfig() error {
 			machineClassSpec["name"] = className
 			machineClassSpec["labels"] = map[string]string{corev1.LabelZoneFailureDomain: zone}
 			machineClassSpec["secret"].(map[string]interface{})["labels"] = map[string]string{v1beta1constants.GardenerPurpose: genericworkeractuator.GardenPurposeMachineClass}
+
+			// if the shoot K8s version is greater than or equal to 1.22 then
+			// enable source destination check on AWS instances by default
+			shootK8sVersion := w.cluster.Shoot.Spec.Kubernetes.Version
+
+			k8sVersionGreaterOrEqualThan122, err := versionutils.CompareVersions(shootK8sVersion, ">=", "1.22")
+			if err != nil {
+				return err
+			}
+
+			if k8sVersionGreaterOrEqualThan122 {
+				machineClassSpec["srcAndDstChecksEnabled"] = false
+			}
 
 			machineClasses = append(machineClasses, machineClassSpec)
 		}


prashanth26:
@AxiomSamarth - We would also have to release the new MCM provider AWS version.

@DockToFuture - There would also be calico changes required right? I am guessing some flags in the configuration of the calico?

charts/internal/machineclass/templates/machineclass.yaml
```diff
@@ -33,7 +33,7 @@ providerSpec:
 {{- if $machineClass.iamInstanceProfile.arn }}
     arn: {{ $machineClass.iamInstanceProfile.arn }}
 {{- end }}
-  keyName: {{ $machineClass.keyName }}
+  keyName: {{ $machineClass.keyName }}
```
AxiomSamarth, line 36: Remove trailing spaces.

AxiomSamarth:
/invite @DockToFuture @ScheererJ 

Is it okay to set the value of `srcAndDstChecksEnabled` to true by default if shoot k8s version is >= 1.22 or should it be set based on the boolean value from the network object [here](https://github.com/gardener/gardener-extension-networking-calico/blob/4ad6c899bfcb0742c5065e8b8ee4d559bb9cd8fc/example/20-network.yaml#L56)?

DockToFuture:
> srcAndDstChecksEnabled



> /invite @DockToFuture @ScheererJ
> 
> Is it okay to set the value of `srcAndDstChecksEnabled` to true by default if shoot k8s version is >= 1.22 or should it be set based on the boolean value from the network object [here](https://github.com/gardener/gardener-extension-networking-calico/blob/4ad6c899bfcb0742c5065e8b8ee4d559bb9cd8fc/example/20-network.yaml#L56)?

It should either be disabled starting from k8s version >= 1.22 which would makes things easier or it should be set based on the boolean value from the network object. For the latter case I will introduce the boolean field `spec.providerConfig.srcAndDstChecks.enabled` in the network resource.
```

  providerConfig:
    apiVersion: calico.networking.extensions.gardener.cloud/v1alpha1
    kind: NetworkConfig
    srcAndDstChecks:
      enabled: false
```

prashanth26:
Depends on https://github.com/gardener/gardener-extension-provider-aws/pull/387.

pkg/controller/worker/machines.go
```diff
@@ -204,6 +205,21 @@ func (w *workerDelegate) generateMachineConfig() error {
...
+				return errors.Wrapf(err, "Invalid version string '%s'", w.cluster.Shoot.Spec.Kubernetes.Version)
+			}
+			supportedK8sVersion, err := semver.NewVersion("1.22")
+			if err != nil {
+				return errors.Wrapf(err, "Invalid version string 1.22")
+			}
+
+			if shootK8sVersion.Compare(supportedK8sVersion) >= 0 {
+				machineClassSpec["srcAndDstChecksEnabled"] = false
+			}
```
rfranzke, line 221: FYI: You could use https://github.com/gardener/gardener/blob/6cdb8dc690c2464558c228105852eb3ac7c2c05a/pkg/utils/version/version.go#L28-L35 which would make this a bit shorter, I guess :)

AxiomSamarth: Thank you, @rfranzke! Version utils is pretty handy.

rfranzke (dismissed):
/lgtm code-wise
/needs second-opinion

prashanth26:
Should we enable this setting by default on the CNIs? Or will the CNI support both ways? I am guessing it would be the latter. In that case, it should be fine. 

cc: @DockToFuture for opinion.

pkg/controller/worker/machines.go
```diff
@@ -204,6 +205,19 @@ func (w *workerDelegate) generateMachineConfig() error {
...
+			// enable source destination check on AWS instances by default
+			shootK8sVersion := w.cluster.Shoot.Spec.Kubernetes.Version
+
+			k8sVersionGreaterOrEqualThan122, err := versionutils.CompareVersions(shootK8sVersion, ">=", "1.22")
+			if err != nil {
+				return err
+			}
+
+			if k8sVersionGreaterOrEqualThan122 {
+				machineClassSpec["srcAndDstChecksEnabled"] = false
```
vpnachev, line 218:
```golang
			// if the shoot K8s version is greater than or equal to 1.22 then
			// enable source destination check on AWS instances by default
```

Shouldn't it be ?
```suggestion
				machineClassSpec["srcAndDstChecksEnabled"] = true
```

DockToFuture:
> Should we enable this setting by default on the CNIs
What do you mean with `Should we enable this setting by default on the CNIs`?
It depends on the networking mode which is used, with an overlay we can use both with native infrastructure routing, we need to disable the checks otherwise it won't work.

pkg/controller/worker/machines.go
DockToFuture: They should be disabled by default, they already are enabled by default.

AxiomSamarth: Ah, my bad! The code is right. The comment is wrong. I will update it.

prashanth26:
> @prashanth26 What do you mean with `Should we enable this setting by default on the CNIs`?
> It depends on the networking mode which is used, with an overlay we can use both with native infrastructure routing, we need to disable the checks otherwise it won't work.

My question was that with current changes the check would be disabled by default with `1.22`.  And with this change, I hope both modes will be supported by the CNI. I guess it shouldn't affect the working of CNI in either configuration i.e. with/without IP-in-IP encapsulation?

rfranzke:
cc @AxiomSamarth @DockToFuture

AxiomSamarth:
@rfranzke I believe implementation is in place.

@DockToFuture Any more comments?

DockToFuture:
> was that with current changes the check would be disabled by default with `1.22`. And with this change, I hope both modes will be supported by the CNI. I guess it shouldn't affect the working of CNI in either configuration i.e. with/without IP-in-IP encapsulation right? It should work both ways?

yes it will work in both modes with disabled checks

rfranzke:
Thanks! @prashanth26 any further comments?

prashanth26 (approved):
/lgtm 
Thanks. no further comments.

ialidzhikov:
/hold
as there is already ongoing release validation

ialidzhikov:
/unhold
as v1.28.0 was released

rfranzke:
cc @timuthy

PR #405, opened 2021-09-01
Update kubernetes-csi/external-snapshotter to v4.2.1

yuvalman:
**How to categorize this PR?**
/area storage
/kind enhancement
/platform gcp

**What this PR does / why we need it**:

Upgrade external-snapshotter to v4 is a must for being aligned with external-provisioner v3.0.0:
https://github.com/kubernetes-csi/external-provisioner/blob/release-3.0/CHANGELOG/CHANGELOG-3.0.md#no-really-you-must-read-this-before-you-upgrade 

**Which issue(s) this PR fixes**:
We have a csi-provision driver that fails because of this issue:
https://github.com/kubernetes-csi/external-provisioner/issues/582

**Special notes for your reviewer**:

**Release note**:
```other operator
The following images are updated:
- k8s.gcr.io/sig-storage/csi-snapshotter: v3.0.3 -> v4.2.1
- k8s.gcr.io/sig-storage/snapshot-controller: v3.0.3 -> v4.2.1

Volume snapshots CRDs version was updated to v1

```

PR #366, opened 2021-06-27
Update kubernetes-csi/external-snapshotter

ialidzhikov:
/area storage
/kind enhancement
/platform aws

**Special notes for your reviewer**:
external-snapshotter@v3 also features a new validating webhook component that provides tightened validation on snapshot objects. It is recommended to install this webhook component, see https://github.com/kubernetes-csi/external-snapshotter/tree/v3.0.3#validating-webhook. For now I didn't introduce the new validating webhook component but I think it is something that we want to introduce in general. Ref to the KEP - https://github.com/kubernetes/enhancements/tree/master/keps/sig-storage/1900-volume-snapshot-validation-webhook.

**Release note**:

```feature user
The following images are updated (see [CHANGELOG](https://github.com/kubernetes-csi/external-snapshotter/blob/v3.0.3/CHANGELOG/CHANGELOG-3.0.md) for more details):
- k8s.gcr.io/sig-storage/csi-snapshotter: v2.1.5 -> v3.0.3
- k8s.gcr.io/sig-storage/snapshot-controller: v2.1.5 -> v3.0.3
```

PR #406, opened 2021-09-02
Update kubernetes-csi/external-provisioner to v3.0.0

yuvalman:
/area storage
/kind enhancement
/platform aws

Upgrade external-provisioner to v3  for being aligned with this pr: https://github.com/gardener/gardener-extension-provider-aws/pull/405

We have a csi-provision driver that fails because of this issue:
kubernetes-csi/external-provisioner#582

**Release note**:
```other operator
The following images are updated:
- k8s.gcr.io/sig-storage/csi-provisioner v2.1.2 -> v3.0.0

```

### DIFF
--- a/charts/images.yaml
+++ b/charts/images.yaml
@@ -53,7 +53,13 @@ images:
 - name: csi-provisioner
   sourceRepository: github.com/kubernetes-csi/external-provisioner
   repository: k8s.gcr.io/sig-storage/csi-provisioner
+  tag: "v2.1.2"
+  targetVersion: "< 1.20"
+- name: csi-provisioner
+  sourceRepository: github.com/kubernetes-csi/external-provisioner
+  repository: k8s.gcr.io/sig-storage/csi-provisioner
   tag: "v3.0.0"
+  targetVersion: ">= 1.20"
 - name: csi-attacher
   sourceRepository: github.com/kubernetes-csi/external-attacher
   repository: k8s.gcr.io/sig-storage/csi-attacher
--- a/charts/images.yaml
+++ b/charts/images.yaml
@@ -53,7 +53,7 @@ images:
 - name: csi-provisioner
   sourceRepository: github.com/kubernetes-csi/external-provisioner
   repository: k8s.gcr.io/sig-storage/csi-provisioner
-  tag: "v2.1.2"
+  tag: "v3.0.0"
 - name: csi-attacher
   sourceRepository: github.com/kubernetes-csi/external-attacher
   repository: k8s.gcr.io/sig-storage/csi-attacher


charts/images.yaml
```diff
@@ -53,7 +53,7 @@ images:
 - name: csi-provisioner
   sourceRepository: github.com/kubernetes-csi/external-provisioner
   repository: k8s.gcr.io/sig-storage/csi-provisioner
-  tag: "v2.1.2"
+  tag: "v3.0.0"
```
ialidzhikov, line 56: First, I have the same comment such as https://github.com/gardener/gardener-extension-provider-gcp/pull/321/files#r704090886:

> Update external-snapshotter dependency to v4, enabling use of the volume snapshot v1 client, which requires minimum Kubernetes version 1.20 (#650, @xinyanw409)

Second, wrt to updating to `external-snapshotter@v4` I have the following concern. In https://github.com/gardener/gardener-extension-provider-aws/pull/366#issue-930985490 I described that `external-snapshotter@v3` features a new validating webhook component. We currently don't install this new validating webhook component. The note from the external-snapshotter REAME.md is exactly this one:

> :warning: **WARNING**: Cluster admins choosing not to install the webhook server and participate in the phased release process can cause future problems when upgrading from `v1beta1` to `v1` volumesnapshot API, if there are currently persisted objects which fail the new stricter validation. Potential impacts include being unable to delete invalid snapshot objects.

`external-snapshotter@v4` is exactly the version that upgrades the API from `v1beta1` to `v1`. So before upgrading to `external-snapshotter@v4` we have to decide what we do with the external-snapshotter validating webhook component - do we install it; do we proceed without it (highly not recommended by the upstream project) and what would be the impact/consequences then? Without clarifying these questions, there is no point to update to the affected version of the sidecars.

/cc @rfranzke 
/hold

ialidzhikov (changes requested):

charts/images.yaml
rfranzke: Good points @ialidzhikov, I'd vote to follow the suggestion of the community and install this webhook to make sure that future problems are prevented/minimized.

yuvalman: Hi,
I added the fix regarding targetVersion.

If you decide to install the webhook, will it be ok that I'll create the separate pr for it?

yuvalman: And if yes,
We have an architecture question:
Should the validation webhook be configured in the seed cluster or in the shoot cluster?

rfranzke: The webhook server should run in the seed cluster targeting the shoot (similar to how the csi-controllers are ran). Maybe it might even make sense to put it as a sidecar container into the csi-snapshot-controller pod, not sure what the community recommends here.

yuvalman: @rfranzke,
Do you know where the resources (Deployment/Service/etc.) of the csi-controllers are configured in the seed cluster? In which repo can I find these resources?
Also, can I get access to seed cluster to see how the resources are implemented in the cluster?

rfranzke: https://github.com/gardener/gardener-extension-provider-aws/tree/master/charts/internal/seed-controlplane/charts/csi-driver-controller/templates

> Also, can I get access to seed cluster to see how the resources are implemented in the cluster?

I don't get this question, you can run Gardener with this extension and your own seeds. There are no "central seed clusters" one could get access to.

PR #424, opened 2021-10-08
Vendor gardener/gardener@1.33.0

voelzmo:

/area open-source
/kind enhancement
/platform aws

Re-vendor gardener/gardener to v1.33.0

**Release note**:

```other operator
NONE
```

### DIFF
--- a/test/integration/infrastructure/infrastructure_test.go
+++ b/test/integration/infrastructure/infrastructure_test.go
@@ -298,20 +298,7 @@ var _ = Describe("Infrastructure tests", func() {
 			Expect(c.Create(ctx, namespace)).To(Succeed())
 
 			By("create cluster")
-			shoot := gardencorev1beta1.Shoot{
-				TypeMeta: metav1.TypeMeta{
-					APIVersion: gardencorev1beta1.SchemeGroupVersion.String(),
-					Kind:       "Shoot",
-				},
-				ObjectMeta: metav1.ObjectMeta{Name: "testShoot"},
-			}
-			shootJson, _ := json.Marshal(shoot)
-			cluster = &extensionsv1alpha1.Cluster{
-				ObjectMeta: metav1.ObjectMeta{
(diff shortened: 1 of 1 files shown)


ialidzhikov:
@voelzmo , there are integration test failures. Can you please check?

voelzmo:
> Do I get it correctly, that the watchdog still won't be started during the integration tests, because the owner `DNSRecord` is optional? We just need to add `cluster.shoot.name` and the `DNSRecord` CRD here to allow the controller to decide not to use the watchdog, correct?

My understanding is that we're executing this code https://github.com/gardener/gardener-extension-provider-aws/blob/1189030d537cb698383985c0bf150cb04e2fc112/vendor/github.com/gardener/gardener/extensions/pkg/controller/infrastructure/reconciler.go#L108-L120 which seems to start a watchdog process if it is not started yet. Is this undesired? Should this be feature-gated somehow?

stoyanr:
> We just need to add cluster.shoot.name and the DNSRecord CRD here to allow the controller to decide not to use the watchdog, correct?

Yes, correct, see also below.

> which seems to start a watchdog process if it is not started yet

As @timebertt  pointed out, the watchdog will only be started if there is indeed an owner `DNSRecord` resource in the shoot namespace. If there is no such record, no watchdog will be started, and everything will work as before the watchdogs were introduced. This was done like this because the owner DNS records themselves are only created if the `UseDNSRecords` feature gate (still alpha) is `true`.

ialidzhikov:
> I would propose to skip the watchdog creation for the integration tests for now.

My assumption was that this point is clarified. The tests simply adds the infrastructure controller to a controller-runtime manager. I don't see the watchdog creation to be configurable, so I guess the current revision is the best thing we can do on the integration tests.

> So just to summarize: we're keeping everything as-is in the PR and merge it,

Let's wait until for the @timebertt and @stoyanr 's review/approval.

PR #425, opened 2021-10-10
Add client-side provider rate limiters to `DNSRecord` controller

stoyanr:

/area networking
/kind enhancement
/platform aws

Adds client-side provider rate limiters (per access key id) to the `DNSRecord` controller so the controller can throttle its requests to the provider to avoid exceeding the actual provider rate limits. The throttling can be configured via the `--dnsrecord-provider-client-qps` and `--dnsrecord-provider-client-burst` options.

Fixes #

* Perhaps the `DNSRecordControllerOptions` and `DNSRecordControllerConfig` types should be moved to g/g, since they may be applicable to `DNSRecord` controllers for other providers as well. 
* An earlier mechanism to mitigate exceeding the rate limit by ensuring that retries on provider errors happen after 30 seconds have been removed, since the new rate limiting mechanism can achieve similar results more elegantly.

**Release note**:

```feature operator
The `DNSRecord` controller will now throttle its requests to the provider to avoid exceeding the provider rate limits. The throttling can be configured via the `--dnsrecord-provider-client-qps` and `--dnsrecord-provider-client-burst` options.
```

### DIFF
--- a/pkg/aws/client/client.go
+++ b/pkg/aws/client/client.go
@@ -36,7 +36,9 @@ import (
 	"github.com/aws/aws-sdk-go/service/s3/s3iface"
 	"github.com/aws/aws-sdk-go/service/sts"
 	"github.com/aws/aws-sdk-go/service/sts/stsiface"
+	"github.com/go-logr/logr"
 	"golang.org/x/time/rate"
+	"sigs.k8s.io/controller-runtime/pkg/log"
 )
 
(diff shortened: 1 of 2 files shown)


stoyanr:
@timebertt As agreed, moved the rate limiter to the AWS client (in a new commit). Unfortunately, this means that your last 2 recommendations can't be easily addressed (unless I spent some additional effort to add unit tests for the client, or at least for the route53 methods). Hopefully this is ok.

PR #439, opened 2021-11-04
Upgrade terraformer to v2.13.0, a version that uses terraform 0.13

stoyanr:

/area control-plane
/kind enhancement
/platform aws

Upgrades `terraformer` image to v2.13.0, a version that uses `terraform` 0.13.

Fixes #

**Release Notes**:

``` other operator github.com/gardener/terraformer #105 @stoyanr
terraform has been upgraded to 0.13.7
```

### DIFF
--- a/charts/images.yaml
+++ b/charts/images.yaml
@@ -2,7 +2,7 @@ images:
 - name: terraformer
   sourceRepository: github.com/gardener/terraformer
   repository: eu.gcr.io/gardener-project/gardener/terraformer-aws
-  tag: "v2.12.1"
+  tag: "v2.13.1"
 
 - name: cloud-controller-manager
   sourceRepository: github.com/kubernetes/kubernetes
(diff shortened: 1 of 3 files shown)


rfranzke:
Does this also work for existing shoots? https://www.terraform.io/upgrade-guides/0-13.html#in-house-providers talks about some state migration and a new `terraform state replace-provider` command that should be called?

stoyanr:
> Does this also work for existing shoots? https://www.terraform.io/upgrade-guides/0-13.html#in-house-providers talks about some state migration and a new terraform state replace-provider command that should be called?

I am currently in the process of testing and fixing this (hence `Draft`). As a matter of fact, it doesn't work even with the non-inhouse providers, the `terraform state replace-provider` command is needed there as well.

ialidzhikov:
/hold
as we want to tackle few existing terraform issue before the update to terraform@0.13

stoyanr:
/invite @ialidzhikov @dkistner 

The new terraform version has already been tested with AWS, see https://github.com/gardener/terraformer/pull/105 for more details.

ialidzhikov:
/unhold

As the issues mentioned in https://github.com/gardener/gardener-extension-provider-aws/pull/439#issuecomment-974007554 are now mitigated/fixed.

dkistner (approved):
/lgtm

PR #453, opened 2021-11-26
Enable volume encryption by default

dkistner:

/area security
/kind enhancement
/platform aws

Gardener managed storage classes will now produce encrypted volumes by default.
This will only apply for newly created volumes.
Old unencrypted volumes created via the Gardener managed storage classes will not be touched.

**Release note**:

```breaking user
The Gardener managed storage classes for AWS Shoot clusters will now produce encrypted volumes by default. Existing volumes will not be changed, only new volumes will be encrypted by default. In case non-encrypted volumes are still required a storage class needs to be deployed by the user.
```

/invite @ialidzhikov


ialidzhikov (dismissed):
/lgtm
/needs rebase

kon-angelo (approved):
/lgtm

ialidzhikov (approved):
/lgtm

PR #458, opened 2021-12-07
Make route53 rate limiter wait timeout configurable

stoyanr:

/area robustness
/kind enhancement
/platform aws

* Makes the route53 rate limiter wait timeout that was previously hardcoded to 10s configurable via the `--dnsrecord-provider-client-wait-timeout` command line option. The "correct" value of this timeout actually depends on the configured provider client QPS and max concurrent reconciles (workers), and therefore should be configurable.
* Ensures that route53 throttling errors and rate limiter wait errors are reported with a fixed requeue interval (hardcoded to 30s) in order to avoid retries with backoff in such cases that may lead to longer reconciliation times when many dnsrecords are reconciled at the same time.

Fixes #

* The QPS should be configured to ensure that no or few actual rate limit errors are encountered if all (or a predefined desired number of) seeds are reconciling dnsrecords at the same time. For example, with server side QPS of 5 (route53 default) and 25 seeds, a reasonable value of QPS would be 5 / 25 = 0.2 for all seeds (so 0.4 for half of the seeds, etc.)
* The wait timeout should be configured to ensure that a certain percentage of workers are actually blocked and waiting, as opposed to reporting rate limit wait errors. The higher this number the better in terms of throughput when many dnsrecords of the same account are being reconciled at the same time, but a certain percentage of workers should always be free to ensure that dnsrecords of other accounts can also be reconciled. For example, to block 50% out of 100 workers with QPS = 0.2, the wait timeout should be set to (100 / 0.2)*0.5 = 250 seconds.

**Release note**:

```other operator
It is now possible to configure the route53 rate limiter wait timeout via the `--dnsrecord-provider-client-wait-timeout` command line option.
```


rfranzke (approved):
/lgtm
/needs rebase

stoyanr:
Rebased on latest `master`

timebertt (approved):
/lgtm

PR #463, opened 2021-12-16
Revendor `gardener/gardener@v1.38` and update Golang version to `1.17`

rfranzke:

/area open-source
/kind enhancement
/platform aws

Revendor `gardener/gardener@v1.38` and update Golang version to `1.17`.

ℹ️ The `branch.cfg` definition is still using Golang `1.16`, hence, the `verify` step fails. Once this PR is approved, we can adapt the `branch.cfg` (also update to `1.17`) and then re-trigger the CI pipeline for this PR.

Part of https://github.com/gardener/gardener/issues/4866

**Release note**:

```noteworthy developer
The Golang version has been updated to `1.17.5`.
```
```noteworthy user
Since go1.17 both `net.ParseIP` and `net.ParseCIDR` reject leading zeros in the dot-decimal notation of IPv4 addresses. With the update to go1.17, admission-aws now rejects Shoot objects with CIDR ranges that have such leading zeros in the dot-decimal notation. Before updating to this version of admission-aws, make sure that there are no Shoot objects with leading zeros in the dot-decimal notation of an IPv4 address. For reference: https://nvd.nist.gov/vuln/detail/CVE-2021-29923
```


Dockerfile
```diff
@@ -1,12 +1,12 @@
 ############# builder
-FROM golang:1.16.3 AS builder
+FROM golang:1.17.5 AS builder
```
ialidzhikov, line 2: You could add the following release note of type `breaking user`:

```
Since go1.17 both `net.ParseIP` and `net.ParseCIDR` reject leading zeros in the dot-decimal notation of IPv4 addresses. With the update to go1.17, admission-aws now rejects Shoot objects with CIDR ranges that have such leading zeros in the dot-decimal notation. Before updating to this version of admission-aws, make sure that there are no Shoot objects with leading zeros in the dot-decimal notation of an IPv4 address. For reference: https://nvd.nist.gov/vuln/detail/CVE-2021-29923
```

It also make sense to confirm that we don't have Shoots in our landscapes with such IPv4 addresses in their InfrastructureConfig.

go.mod
```diff
@@ -8,8 +8,8 @@ require (
 	github.com/ahmetb/gen-crd-api-reference-docs v0.2.0
 	github.com/aws/aws-sdk-go v1.38.19
 	github.com/coreos/go-systemd/v22 v22.3.2
-	github.com/gardener/etcd-druid v0.6.1-0.20211022081522-071746e9d0bd
-	github.com/gardener/gardener v1.36.0
+	github.com/gardener/etcd-druid v0.7.0
+	github.com/gardener/gardener v1.38.0
```
ialidzhikov, line 12: Optional

```suggestion
	github.com/gardener/gardener v1.38.1
```

Dockerfile
```diff
@@ -1,12 +1,12 @@
...
 
 WORKDIR /go/src/github.com/gardener/gardener-extension-provider-aws
 COPY . .
 RUN make install
 
 ############# base
-FROM alpine:3.13.2 AS base
+FROM alpine:3.15.0 AS base
```
ialidzhikov, line 9: Months ago (July this year) with @vpnachev we experienced issues with `alpine@3.14` where 3.14 was working only with latest versions of runc (`>=v1.0.0-rc93`). The runc version in the concourse workers was older one. And the runc version in garden-linux@184.0 was an older one. Are these issues now resolved and safe to update to `alpine@3.15.0`?

/cc @AndreasBurger

rfranzke: Can you share more information what these problems were (is there an issue or something about this)?

Dockerfile
rfranzke: I've added the release note. cc @dkistner @kon-angelo for the rollout

go.mod
rfranzke: Is there something relevant in it? Otherwise, I'd propose to wait for the 1.39 revendoring PR.

ialidzhikov: There is https://github.com/gardener/gardener/pull/5194. But of course such version bump (from v1.38.0 to v1.38.1) can be done in separate PR, that's why I also wrote it as optional. Let's then consider it for another PR.

Dockerfile
ialidzhikov: > Can you share more information what these problems were (is there an issue or something about this)?

I forwarded you the corresponding Slack conversation back then. Unfortunately, I don't recall all of the details. 
If we have components that are already using `alpine@3.15` for some time without any issues (wrt to concourse and garden-linux), then this change is fine for me. 👍

ialidzhikov: external-dns-management was recently updated to `alpine@3.15` (ref https://github.com/gardener/external-dns-management/pull/233). Looks like no issues occurred during its release and no issues so far on landscapes. Then I guess it is fine to proceed with `alpine@3.15`.

rfranzke: Hm, the PR was only merged yesterday, so I guess we won't have much experience yet. I tend towards reverting the alpine changes for now asp art of this PR. WDYT?

ialidzhikov: Okay, we can delay the alpine version update and do it when we have more experience with it (we will monitor external-dns-management).

rfranzke: 👍🏻 Thanks :)

ialidzhikov (approved):
/lgtm
/squash

dkistner (approved):
/lgtm

kon-angelo (approved):
/lgtm

PR #486, opened 2022-01-28
Fix instance terminated check

petersutter:

/area quality
/kind bug
/platform aws

This PR fixes an issue / edge case where the bastion instance is not considered terminated when it is already cleaned up. This causes the bastion resource to never be cleaned up and must be cleaned up by hand:
- delete the [bastion security group](https://github.com/gardener/gardener-extension-provider-aws/blob/be02e4cfa70af128a225491e0aed3458e75c99c9/pkg/controller/bastion/actuator_delete.go#L162-L183)
- remove the finalizer on the bastion extension resource on the seed
- confirm that also the bastion resource on the garden cluster is removed

Alternatively, run the `gardener-extension-provider-aws` that includes this fix which will then proceed with the cleanup.

Fixes #

**Release note**:

```bufix operator
Fixed an edge case where the bastion instance is not considered terminated in case it is already cleaned up
```

### DIFF
--- a/pkg/controller/bastion/actuator_delete.go
+++ b/pkg/controller/bastion/actuator_delete.go
@@ -122,13 +122,13 @@ func instanceIsTerminated(ctx context.Context, awsClient *awsclient.Client, opt 
 
 	for _, reservation := range instances.Reservations {
 		for _, instance := range reservation.Instances {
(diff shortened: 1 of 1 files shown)


dkistner (approved):
/lgtm

PR #491, opened 2022-02-14
Classify duplicate and overlapping zones as configuration errors

plkokanov:

/area control-plane
/area ops-productivity
/kind enhancement
/platform aws

This PR classifies DNS errors that have "duplicate zones" or "overlapping zones" in their error message as ERR_CONFIGURATION_PROBLEM

Fixes https://github.com/gardener/gardener/issues/4844

I found out the following error codes that seem describe the duplicate and overlapping zones error but I'm currently testing if that is the case:

https://github.com/gardener/gardener-extension-provider-aws/blob/40348fe6cac55f97bb0051622a4bbd6cdfeb9bee/vendor/github.com/aws/aws-sdk-go/service/route53/errors.go#L14
https://github.com/gardener/gardener-extension-provider-aws/blob/40348fe6cac55f97bb0051622a4bbd6cdfeb9bee/vendor/github.com/aws/aws-sdk-go/service/route53/errors.go#L112

Additionally, they look like possible codes that could be caused by a configuration problem, so perhaps it makes sense to include them anyway.

**Release note**:

```other operator
Error messages containing `duplicate zones` and `overlapping zones` in their description that can happen when creating DNSRecords are now classified as ERR_CONFIGURATION_PROBLEM.
```

### DIFF
--- a/pkg/aws/client/client_route53.go
+++ b/pkg/aws/client/client_route53.go
@@ -17,7 +17,6 @@ package client
 import (
 	"context"
 	"fmt"
-	"regexp"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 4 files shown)


plkokanov:
Moving this classification to g/g as the error is actually for DNSProviders and not DNSRecords

PR #502, opened 2022-02-21
Switch to `gp3` for etcd storage class

dkistner:
/kind enhancement
/platform aws

Switch etcd storage class to use `gp3` disk type as they offer higher iops.
This will affect only newly created etcd disk.

cc @dguendisch

**Release note**:

```other operator
The etcd storage class is now using `gp3` disk type instead of `gp2` as this offers higher iops capabilities. This wil affect only newly created etcd disks.
```


dguendisch:
I just tried it out with this storageclass:
```
allowVolumeExpansion: true
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: test
parameters:
  encrypted: "true"
  type: gp2
provisioner: kubernetes.io/aws-ebs
reclaimPolicy: Delete
volumeBindingMode: Immediate
```

Upon replacing gp2 to gp3 in the spec and reapplying I get:
```
k apply -f sc.yaml
The StorageClass "test" is invalid: parameters: Forbidden: updates to parameters are forbidden.
```

Seems it can't be changed this way 😞

ialidzhikov:
> Seems it can't be changed this way 😞

The StorageClass has the annotation `resources.gardener.cloud/delete-on-invalid-update: "true"`. GRM has to take care to delete the StorageClass and recreate it.

dguendisch:
> GRM has to take care to delete the StorageClass and recreate it.

Indeed, this way it works, thanks for the pointer!

dguendisch (approved):
/lgtm

kon-angelo (approved):
/lgtm

PR #506, opened 2022-02-25
Add object selector for provider type in the `gardener-extension-admission-aws` webhook

shafeeqes:

/area robustness
/kind enhancement
/platform aws

This PR adds an object selector for provider type label in the `gardener-extension-admission-aws` webhook. This helps in filtering out only resources having provider type `aws` for validation,

Fixes part of gardener/gardener#4932

Related to gardener/gardener#5472

**Release note**:

```other operator
The `gardener-extension-admission-aws` webhook now contains an object selector for provider type label. Please make sure you are running `gardener@v1.42` or later before enabling this.
```

### DIFF
--- a/charts/gardener-extension-admission-aws/charts/application/templates/validatingwebhook-validator.yaml
+++ b/charts/gardener-extension-admission-aws/charts/application/templates/validatingwebhook-validator.yaml
@@ -18,7 +18,10 @@ webhooks:
     - shoots
   failurePolicy: Fail
   objectSelector:
-    {{- toYaml (default (dict) .Values.global.webhookConfig.objectSelector) | nindent 4 }}
+    {{- if .Values.global.webhookConfig.useObjectSelector }}
+    matchLabels:
+      provider.extensions.gardener.cloud/aws: "true"
+    {{- end }}
   namespaceSelector: {}
   sideEffects: None
   admissionReviewVersions:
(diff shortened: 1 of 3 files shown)


charts/gardener-extension-admission-aws/charts/application/templates/validatingwebhook-validator.yaml
```diff
@@ -17,7 +17,9 @@ webhooks:
     - cloudprofiles
     - shoots
   failurePolicy: Fail
-  objectSelector: {}
+  objectSelector:
```
rfranzke, line 20: It should be configurable via the Helm chart values whether the object selector should be used. Otherwise, this might not be backwards-compatible if the Gardener version is still old (not adding the labels) while the admission webhooks already relys on them.

Same for other extensions, of course.

example/40-validatingwebhookconfiguration.yaml
```diff
@@ -17,7 +17,9 @@ webhooks:
     - shoots
     - cloudprofiles
   failurePolicy: Fail
-  objectSelector: {}
+  objectSelector:
```
rfranzke, line 20: Here it is fine to rely on it, I guess. Maybe you can add a comment that at least Gardener v1.42 is needed.

charts/gardener-extension-admission-aws/values.yaml
```diff
@@ -32,6 +32,9 @@ global:
         -----BEGIN RSA PRIVATE KEY-----
         ...
         -----END RSA PRIVATE KEY-----
+    # objectSelector:
+    #   matchLabels:
+    #     provider.extensions.gardener.cloud/aws: "true"
```
rfranzke, line 37: What about a simple flag `useObjectSelector` which is `true` or `false`? Then operators would not need to specify them here.

rfranzke (dismissed):
/lgtm

rfranzke (approved):
/lgtm

kon-angelo (approved):
/lgtm

PR #507, opened 2022-02-25
Add snaphot validation webhook

### LINKED ISSUE #489: Install `Snapshot Validation Webhook`  for `external-snapshotter@v3`

/area storage
/kind enhancement
/platform aws

Install `Snapshot Validation Webhook` for `external-snapshotter@v3`.

**Why is this needed**:
With `external-snapshotter@v4` the Snapshot APIs have been moved to V1. A new snapshot validation webhook was introduced with the release of `external-snapshotter@v3`. For upgrading external-snapshotter to v4 as recommended by the community 
> Before upgrading to v1 CRDs and upgrading snapshot controller and validation webhook to v4.0, the user should upgrade to the snapshot controller 3.0.0 and higher (v3.0.3 is the latest recommended version for 3.0.x) and install the corresponding validation webhook so that all existing invalid objects will be labeled and can be easily identified and deleted.

we need to install `Snapshot Validation Webhook` for `external-snapshotter@v3`.

Please see this [KEP](https://github.com/kubernetes/enhancements/tree/master/keps/sig-storage/1900-volume-snapshot-validation-webhook) for full context on why a webhook is needed.

acumino:

/area open-source
/kind enhancement
/platform aws

This PR adds the webhook to provide additional validation to volume snapshot objects i.e. `VolumeSnapshot` and `VolumeSnapshotContent`.

Fixes #489

**Release note**:

```feature user
The provider-aws extension now installs the external-snapshotter's validating webhook server for VolumeSnapshot and VolumeSnapshotContent objects. For more details check the corresponding [KEP](https://github.com/kubernetes/enhancements/tree/master/keps/sig-storage/1900-volume-snapshot-validation-webhook#kep-1900-add-additional-validation-to-volume-snapshot-objects).
```

### DIFF
--- a/charts/internal/seed-controlplane/charts/csi-driver-controller/templates/csi-snapshot-validation-webhook-deployment.yaml
+++ b/charts/internal/seed-controlplane/charts/csi-driver-controller/templates/csi-snapshot-validation-webhook-deployment.yaml
@@ -4,7 +4,7 @@ metadata:
   name: csi-snapshot-validation
   namespace: {{ .Release.Namespace }}
   labels:
-    app: snapshot-validationxs
+    app: snapshot-validation
 spec:
   replicas: {{ .Values.csiSnapshotValidationWebhook.replicas }}
   selector:
--- a/charts/internal/seed-controlplane/charts/csi-driver-controller/templates/csi-snapshot-validation-webhook-deployment.yaml
+++ b/charts/internal/seed-controlplane/charts/csi-driver-controller/templates/csi-snapshot-validation-webhook-deployment.yaml
@@ -4,8 +4,7 @@ metadata:
   name: csi-snapshot-validation
   namespace: {{ .Release.Namespace }}
   labels:
-    app: snapshot-validation
-    role: webhook
+    app: snapshot-validationxs
 spec:
   replicas: {{ .Values.csiSnapshotValidationWebhook.replicas }}
   selector:
--- a/charts/internal/seed-controlplane/charts/csi-driver-controller/templates/csi-snapshot-validation-webhook-deployment.yaml
+++ b/charts/internal/seed-controlplane/charts/csi-driver-controller/templates/csi-snapshot-validation-webhook-deployment.yaml
@@ -1,0 +1,42 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: csi-snapshot-validation
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app: snapshot-validation
+    role: webhook
+spec:
+  replicas: {{ .Values.csiSnapshotValidationWebhook.replicas }}
+  selector:
+    matchLabels:
+      app: snapshot-validation
+  template:
+    metadata:
+      annotations:
+{{- if .Values.csiSnapshotValidationWebhook.podAnnotations }}
+{{ toYaml .Values.csiSnapshotValidationWebhook.podAnnotations | indent 8 }}
+{{- end }}
+      labels:
+        app: snapshot-validation
+        networking.gardener.cloud/from-shoot-apiserver: allowed
+    spec:
+      containers:
+      - name: csi-snapshot-validation
+        image: {{ index .Values.images "csi-snapshot-validation-webhook" }}
+        imagePullPolicy: IfNotPresent
+        args: ['--tls-cert-file=/etc/csi-snapshot-validation/csi-snapshot-validation.crt', '--tls-private-key-file=etc/csi-snapshot-validation/csi-snapshot-validation.key']
+        ports:
+        - containerPort: 443
+{{- if .Values.csiSnapshotValidationWebhook.resources }}
+        resources:
+{{ toYaml .Values.csiSnapshotValidationWebhook.resources | indent 10 }}
+{{- end }}
+        volumeMounts:
+          - name: csi-snapshot-validation
+            mountPath: /etc/csi-snapshot-validation
+            readOnly: true
+      volumes:
+        - name: csi-snapshot-validation
+          secret:
+            secretName: csi-snapshot-validation
--- a/charts/internal/seed-controlplane/charts/csi-driver-controller/templates/csi-snapshot-validation-webhook-vpa.yaml
+++ b/charts/internal/seed-controlplane/charts/csi-driver-controller/templates/csi-snapshot-validation-webhook-vpa.yaml
@@ -1,0 +1,17 @@
+apiVersion: autoscaling.k8s.io/v1beta2
+kind: VerticalPodAutoscaler
+metadata:
+  name: csi-snapshot-webhook-vpa
+  namespace: {{ .Release.Namespace }}
+spec:
+  resourcePolicy:
+    containerPolicies:
+    - containerName: csi-snapshot-validation
+      minAllowed:
+        memory: 25Mi
+  targetRef:
+    apiVersion: apps/v1
+    kind: Deployment
+    name: csi-snapshot-validation
+  updatePolicy:
+    updateMode: Auto
--- a/pkg/controller/controlplane/valuesprovider.go
+++ b/pkg/controller/controlplane/valuesprovider.go
@@ -453,7 +453,7 @@ func (vp *valuesProvider) GetControlPlaneShootChartValues(
 	cluster *extensionscontroller.Cluster,
 	_ map[string]string,
 ) (map[string]interface{}, error) {
-	return getControlPlaneShootChartValues(ctx, cluster, vp.Client(), vp.useTokenRequestor, vp.useProjectedTokenMount)
+	return getControlPlaneShootChartValues(ctx, cluster, cp, vp.Client(), vp.useTokenRequestor, vp.useProjectedTokenMount)
 }
 
 // GetControlPlaneShootCRDsChartValues returns the values for the control plane shoot CRDs chart applied by the generic actuator.
@@ -669,6 +669,7 @@ func getCSIControllerChartValues(
 func getControlPlaneShootChartValues(
 	ctx context.Context,
 	cluster *extensionscontroller.Cluster,
+	cp *extensionsv1alpha1.ControlPlane,
 	client client.Client,
 	useTokenRequestor bool,
 	useProjectedTokenMount bool,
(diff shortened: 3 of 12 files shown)


rfranzke (commented):
Looks great!

acumino:
@rfranzke Please check if there is any specific annotation needed for components, Maybe I missed it because of unawareness.

rfranzke (commented):
We probably should also health-check the new deployment, so add it here: https://github.com/gardener/gardener-extension-provider-aws/blob/master/pkg/controller/healthcheck/add.go (only add it if the `Deployment exists` - otherwise, the health check will fail until the `Shoot` gets reconciled for the first time after the new AWS extension has been deployed).

acumino:
> (only add it if the `Deployment exists` - otherwise, the health check will fail until the `Shoot` gets reconciled for the first time after the new AWS extension has been deployed).

IIUC I need to add a predicate that check if deployment is there or not. If that right why we are nit doing same with the other also [here](https://github.com/gardener/gardener-extension-provider-aws/blob/55b98dbc3b29d191fef4df4eb518ad7c159b986f/pkg/controller/healthcheck/add.go#L66-L92).

pkg/controller/controlplane/valuesprovider.go
```diff
@@ -655,10 +683,20 @@ func getControlPlaneShootChartValues(
 		return nil, err
 	}
 
+	// get the ca.crt for caBundle of the snapshot-validation webhook
+	secret := &corev1.Secret{}
+	if err := client.Get(ctx, kutil.Key(cluster.ObjectMeta.Name, string(secrets.CACert)), secret); err != nil {
```
ialidzhikov, line 688: I guess this is not a proper usage of `secrets.CACert` - to my understanding it rather denotes a CertType in `CertificateSecretConfig` that later gets generated. We rather want to use `SecretNameCACluster` that denotes the `ca` Secret name.

```suggestion
	if err := client.Get(ctx, kutil.Key(cluster.ObjectMeta.Name, v1beta1constants.SecretNameCACluster), secret); err != nil {
```

pkg/controller/controlplane/valuesprovider_test.go
```diff
@@ -276,13 +279,28 @@ var _ = Describe("ValuesProvider", func() {
...
 	})
 
 	Describe("#GetControlPlaneShootChartValues", func() {
+		BeforeEach(func() {
+			c = mockclient.NewMockClient(ctrl)
+
+			err := vp.(inject.Client).InjectClient(c)
+			Expect(err).NotTo(HaveOccurred())
+
+			c.EXPECT().Get(ctx, kutil.Key(cp.ClusterName, string(secrets.CACert)), gomock.AssignableToTypeOf(&corev1.Secret{}))
```
ialidzhikov, line 300:
```suggestion
			c.EXPECT().Get(ctx, kutil.Key(cp.ClusterName, v1beta1constants.SecretNameCACluster, gomock.AssignableToTypeOf(&corev1.Secret{}))
```

ialidzhikov, line 300: Is it possible to add a test case to covert the case when the GET request for the secret fails with an error?

charts/internal/seed-controlplane/charts/csi-driver-controller/templates/csi-snapshot-validation-webhook-deployment.yaml
```diff
@@ -0,0 +1,41 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: csi-snapshot-validation
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app: snapshot-validationxs
```
rfranzke, line 7: typo?

charts/internal/seed-controlplane/charts/csi-driver-controller/templates/csi-snapshot-validation-webhook-vpa.yaml
```diff
@@ -0,0 +1,17 @@
...
+kind: VerticalPodAutoscaler
+metadata:
+  name: csi-snapshot-webhook-vpa
+  namespace: {{ .Release.Namespace }}
+spec:
+  resourcePolicy:
+    containerPolicies:
+    - containerName: csi-snapshot-validation
+      minAllowed:
+        memory: 25Mi
```
rfranzke, line 11: What about some CPU min allowed value?

pkg/controller/healthcheck/add.go
```diff
@@ -78,6 +78,11 @@ func RegisterHealthChecks(mgr manager.Manager, opts healthcheck.DefaultAddArgs)
 				HealthCheck:   general.NewSeedDeploymentHealthChecker(aws.CSISnapshotControllerName),
 				PreCheckFunc:  csiEnabledPreCheckFunc,
 			},
+			{
+				ConditionType: string(gardencorev1beta1.ShootControlPlaneHealthy),
+				HealthCheck:   general.NewSeedDeploymentHealthChecker(aws.CSISnapshotValidation),
+				PreCheckFunc:  csiEnabledPreCheckFunc,
```
rfranzke, line 84: As said, this check will fail the entire time until the `Shoot` got reconciled (which might take up to `24h`). So, you should add an additional pre-check for whether the `Deployment` exists in the shoot namespace in the seed.

charts/internal/seed-controlplane/charts/csi-driver-controller/templates/csi-snapshot-validation-webhook-vpa.yaml
acumino: For now I am following the same behaviour as snapshot-controller.

pkg/controller/healthcheck/add.go
acumino: It can't be implemented using PreCheckFunc because we can't get client. Please check [here](https://github.com/gardener/gardener-extension-provider-aws/blob/55b98dbc3b29d191fef4df4eb518ad7c159b986f/vendor/github.com/gardener/gardener/extensions/pkg/controller/healthcheck/actuator.go#L58).

rfranzke: You can use `mgr.GetClient()` to get it. But I see that we'd still lack the `context`. Potentially the signature of the `PreCheckFunc` can be improved?

acumino: Yes that again make it two step process i.e. first make changes in g/g and then in extensions, also with time in future i.e. with next release this precheck need to be removed otherwise [this check](https://github.com/gardener/gardener-extension-provider-aws/blob/55b98dbc3b29d191fef4df4eb518ad7c159b986f/vendor/github.com/gardener/gardener/extensions/pkg/controller/healthcheck/general/deployment.go#L98-L103) will not be respected. I will prefer to leave it, WDYT?

rfranzke: Your concerns are valid, but simply keeping it as it is now is no option, I fear. Hundreds, or rather thousands of `Shoot`s (depending on landscape size) will have failing health checks until they get reconciled which might take `24h`. This will not help operators of such landscapes.

Maybe we can be pragmatic and

1. Not add the health check for this new `Deployment` now (add it but comment it)
2. Add a TODO to enable it only in a future release
3. Still improve the `PreCheck` signature in g/g so that we can add such logic in the future in case we need it

acumino: Ok, Thanks for explanation, 
But still we will be required to make changes two time in provider extension i.e add the precheck condition `for whether the Deployment exists` and in the next release remove the precheck condition, is there a possible solution so we can make it one step process?

rfranzke: My proposal (https://github.com/gardener/gardener-extension-provider-aws/pull/507#discussion_r816566940) was to not have the health check now (1) and only enable it later (2) by just removing the commented code. There is no need to introduce the pre-check logic we discussed earlier then because we can assume that the `Deployment` exists in the meantime.
I don't think we can find a one-step approach, but the proposal is pragmatic with minimal impact.

acumino: PTAL 5c899f2d21db9ea4c3e6c96a39ee85c3df8b93de

rfranzke (approved):
/lgtm

pkg/controller/controlplane/valuesprovider.go
```diff
@@ -655,10 +682,20 @@ func getControlPlaneShootChartValues(
 		return nil, err
 	}
 
+	// get the ca.crt for caBundle of the snapshot-validation webhook
+	secret := &corev1.Secret{}
+	if err := client.Get(ctx, kutil.Key(cluster.ObjectMeta.Name, string(v1beta1constants.SecretNameCACluster)), secret); err != nil {
```
ialidzhikov, line 687: Now going 1 more time over this handling I guess it would be more suitable to use the ControlPlane namespace rather than the Cluster name.

```suggestion
	if err := client.Get(ctx, kutil.Key(cp.Namespace, string(v1beta1constants.SecretNameCACluster)), secret); err != nil {
```

pkg/controller/controlplane/valuesprovider.go
```diff
@@ -655,10 +682,20 @@ func getControlPlaneShootChartValues(
...
+		return nil, err
+	}
+
 	csiDriverNodeValues := map[string]interface{}{
 		"enabled":           csiEnabled,
 		"kubernetesVersion": kubernetesVersion,
 		"vpaEnabled":        gardencorev1beta1helper.ShootWantsVerticalPodAutoscaler(cluster.Shoot),
+		"webhookConfig": map[string]interface{}{
+			"url":      "https://" + aws.CSISnapshotValidation + "." + cluster.ObjectMeta.Name + "/volumesnapshot",
```
ialidzhikov, line 696:
```suggestion
			"url":      "https://" + aws.CSISnapshotValidation + "." + cp.Namespace + "/volumesnapshot",
```

pkg/controller/healthcheck/add.go
```diff
@@ -78,6 +79,12 @@ func RegisterHealthChecks(mgr manager.Manager, opts healthcheck.DefaultAddArgs)
 				HealthCheck:   general.NewSeedDeploymentHealthChecker(aws.CSISnapshotControllerName),
 				PreCheckFunc:  csiEnabledPreCheckFunc,
 			},
+			// TODO(acumino): Enable this health check in future release
```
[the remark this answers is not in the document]
acumino: PTAL 5df9a1e50724489f52b30f5286829d5ceb2b2a21

ialidzhikov (approved):
/lgtm

rfranzke (approved):

kon-angelo (approved):
/lgtm

PR #516, opened 2022-03-08
Move shoot webhook startup reconciliation into runnable

rfranzke:

/area usability
/kind bug
/platform aws

This PR moves the shoot webhook reconciliation during startup into a `Runnable` (as suggested in gardener/gardener#5358) and only executes it when the respective instance of the extension was elected as leader.

Part of gardener/gardener#5358

**Release note**:

```bugfix operator
A race condition preventing shoot namespaces from being cleaned up due to orphaned resources has been fixed.
```

### DIFF
--- a/cmd/gardener-extension-provider-aws/app/app.go
+++ b/cmd/gardener-extension-provider-aws/app/app.go
@@ -20,21 +20,6 @@ import (
 	"os"
 	"time"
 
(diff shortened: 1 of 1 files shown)


ialidzhikov (approved):
Thanks for the detailed explanations in https://github.com/gardener/gardener/issues/5358. They were very helpful to understand the motivation for this change.

/lgtm

PR #555, opened 2022-06-02
Add label to nodes indicating whether node local dns is enabled or not.

ScheererJ:

/area networking
/kind enhancement
/platform aws

Add label to nodes indicating whether node local dns is enabled or not.

The label will later be used by a node selector to let node local dns components run until
the corresponding nodes are rolled.
The constant values will be changed to proper constants when the new gardener release is
revendored.

None.

**Release note**:

```other operator
Nodes get a label indicating whether node-local-dns is enabled or not.
```

### DIFF
--- a/pkg/controller/worker/machines.go
+++ b/pkg/controller/worker/machines.go
@@ -28,6 +28,7 @@ import (
 	"github.com/gardener/gardener/extensions/pkg/controller/csimigration"
 	"github.com/gardener/gardener/extensions/pkg/controller/worker"
 	genericworkeractuator "github.com/gardener/gardener/extensions/pkg/controller/worker/genericactuator"
+	gardencorev1beta1constants "github.com/gardener/gardener/pkg/apis/core/v1beta1/constants"
 	v1beta1constants "github.com/gardener/gardener/pkg/apis/core/v1beta1/constants"
 	extensionsv1alpha1 "github.com/gardener/gardener/pkg/apis/extensions/v1alpha1"
 	"github.com/gardener/gardener/pkg/client/kubernetes"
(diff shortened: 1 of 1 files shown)


pkg/controller/worker/machines.go
```diff
@@ -204,7 +206,7 @@ func (w *workerDelegate) generateMachineConfig() error {
 					}
 					// TODO: remove the csi topology label when AWS CSI driver stops using the aws csi topology key - https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/899
 					// add aws csi driver topology label if its not specified
-					return utils.MergeStringMaps(pool.Labels, map[string]string{awsCSIDriverTopologyKey: zone})
+					return utils.MergeStringMaps(pool.Labels, map[string]string{awsCSIDriverTopologyKey: zone, "networking.gardener.cloud/node-local-dns-enabled": nodeLocalDNSLabel})
```
kon-angelo, line 209: If you unconditionally include it in the map then the annotation will be there (even if the value in the kv will be ""). Is that the intended behavior ?

ialidzhikov, line 209: Cann't we add this label from a central place in gardenlet instead of doing this change for each and every provider extension?

ScheererJ:
@kon-angelo / @ialidzhikov : Thanks for the feedback.

I meanwhile learnt how to do this in the gardener core (see https://github.com/gardener/gardener/pull/6057 for details). Therefore, this change becomes obsolete.
Initially, my tests indicated that the labels of the workers are directly applied to the nodes. However, if the worker pool hash changes the labels are not applied to the old instances. This makes it a lot easier to implement in the gardener core.

PR #568, opened 2022-06-30
Vendor gardener/gardener@v1.50.1-0.20220701074220-216e074e49da

acumino:

/area open-source
/kind enhancement
/platform aws

Fixes #

**Release note**:

```other operator
The following dependency is updated:
- github.com/gardener/gardener: v1.48.0 -> v1.50.0
```


kon-angelo (approved):
/lgtm

acumino:
> I think we should rather release 1.50.1 and vendor this tag instead of a "random" commit.

Yes, though I have used the commit from the 1.50 release branch so merging this will not be an issue and we can later vendor 1.50.1 once it is released.

go.mod
```diff
@@ -18,19 +18,19 @@ require (
...
+	golang.org/x/time v0.0.0-20220210224613-90d013bbcef8
 	golang.org/x/tools v0.1.10
-	k8s.io/api v0.23.3
-	k8s.io/apiextensions-apiserver v0.23.3
-	k8s.io/apimachinery v0.23.3
-	k8s.io/autoscaler/vertical-pod-autoscaler v0.10.0
+	k8s.io/api v0.24.2
+	k8s.io/apiextensions-apiserver v0.24.2
+	k8s.io/apimachinery v0.24.2
+	k8s.io/autoscaler/vertical-pod-autoscaler v0.11.0
```
ialidzhikov, line 26: Don't we need to update the replace directive for `k8s.io/autoscaler/vertical-pod-autoscaler` to point to v0.11.0 as well?

go.mod
```diff
@@ -8,8 +8,8 @@ require (
 	github.com/ahmetb/gen-crd-api-reference-docs v0.2.0
 	github.com/aws/aws-sdk-go v1.38.19
 	github.com/coreos/go-systemd/v22 v22.3.2
-	github.com/gardener/etcd-druid v0.8.0
-	github.com/gardener/gardener v1.48.0
+	github.com/gardener/etcd-druid v0.9.0
+	github.com/gardener/gardener v1.50.1-0.20220701074220-216e074e49da
```
ialidzhikov, line 12: https://github.com/gardener/gardener/pull/6059/commits/ccaea580cfb6236a1a11ef7fc164f1b37b60697d makes the metrics bind address configurable. We can expose the metricsPort as a ControllerDeployment value in a similar way like it is done in this commit for provider-local. But if you want let's better do it in a separate follow-up PR.

acumino: Addressed in follow-up PR #570.

go.mod
acumino: Addressed in follow-up PR #570.

PR #571, opened 2022-07-06
Adapt the priority according to the gardener guideline

kris94:

/area quality
/kind enhancement
/platform aws

This PR removes the old `PriorityClass` from the charts and adapt the deployment priorityClassName according to the gardener documentation. Part of [#5634](https://github.com/gardener/gardener/issues/5634)
Fixes #

**Release note**:

```breaking operator
This version of provider-aws requires Gardener v1.50+. 
```


charts/gardener-extension-admission-aws/charts/runtime/templates/deployment.yaml
```diff
@@ -21,6 +21,7 @@ spec:
       labels:
 {{ include "labels" . | indent 8 }}
     spec:
+      priorityClassName: gardener-system-900
```
dkistner, line 24: Shorty question. This component will be deployed into the host cluster and interact with the gardener api-server so it's deployed into a seed or shoot nor interacting with them. So will this priority class also available in the context where this component is deployed?

charts/gardener-extension-provider-aws/templates/daemonset-custom-mtu.yaml
```diff
@@ -19,7 +19,7 @@ spec:
       labels:
         k8s-app: mtu-customizer
     spec:
-      priorityClassName: gardener-extension-provider-aws
+      priorityClassName: gardener-system-90
```
dkistner, line 22: Is it `gardener-system-900` instead of `gardener-system-90`? Same for others below? ref: https://github.com/gardener/gardener/blob/master/docs/development/priority-classes.md

kris94: Yes, its a typo thanks for noticing

charts/gardener-extension-admission-aws/charts/runtime/templates/deployment.yaml
kris94: Yes, PriorityClasses `gardener-shoot-system(600-900)` are deployed to the shoot cluster as well.

dkistner: Well, but the component is not deployed to a Shoot or a Seed. It is deployed to the cluster where the Gardener components run. So will it be ensured that the `priorityClass` you referenced is also there. At least there is probably no gardenlet which could take care.

kris94: Ok, I misunderstood you. So, if it is not part of the Shoot System Components, Seed System Components or Shoot Control Plane Components, we should skip this deployment.

PR #572, opened 2022-07-15
Adds MaxAllowed for VPA objects

dimitar-kostadinov:

/area auto-scaling
/kind enhancement
/platform aws

Limits the range of maximum allowed resource requests to ensure that pods are schedulable
Fixes #

**Release note**:

```other operator
NONE
```

### DIFF
--- a/charts/internal/machine-controller-manager/seed/templates/vpa.yaml
+++ b/charts/internal/machine-controller-manager/seed/templates/vpa.yaml
@@ -17,10 +17,16 @@ spec:
       minAllowed:
         cpu: {{ .Values.resources.mcmProviderAWS.requests.cpu }}
         memory: {{ .Values.resources.mcmProviderAWS.requests.memory }}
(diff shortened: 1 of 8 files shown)


charts/internal/machine-controller-manager/seed/templates/vpa.yaml
```diff
@@ -17,10 +17,16 @@ spec:
       minAllowed:
         cpu: {{ .Values.resources.mcmProviderAWS.requests.cpu }}
         memory: {{ .Values.resources.mcmProviderAWS.requests.memory }}
+      maxAllowed:
```
kon-angelo, line 20: Can you add these to the `values` file and pass them via templates ? (similar to the minAllowed)

dimitar-kostadinov: Done

PR #583, opened 2022-07-26
Remove install-requirement and use tools.mk

acumino:

/area dev-productivity
/kind enhancement
/platform aws

This PR adapts the tool concept according to https://github.com/gardener/gardener/issues/6396 and removes the install-requirement script to install dependencies required for development on the fly and as needed.

Part of https://github.com/gardener/gardener/issues/6396

**Release note**:

```other operator
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -96,15 +96,6 @@ docker-images:
 # Rules for verification, formatting, linting, testing and cleaning #
 #####################################################################
 
-.PHONY: install-requirements
-install-requirements:
-	@go install -mod=vendor $(REPO_ROOT)/vendor/github.com/ahmetb/gen-crd-api-reference-docs
-	@go install -mod=vendor $(REPO_ROOT)/vendor/github.com/golang/mock/mockgen
-	@go install -mod=vendor $(REPO_ROOT)/vendor/golang.org/x/tools/cmd/goimports
-	@go install -mod=vendor sigs.k8s.io/controller-runtime/tools/setup-envtest
-	@go install -mod=vendor sigs.k8s.io/controller-tools/cmd/controller-gen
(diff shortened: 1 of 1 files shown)


rfranzke:
This PR has - again - no reasonable issue description whatsoever. The title states "and use tools.mk" but it's not clear from the changes how this is used and not explained anywhere else.

acumino:
> This PR has - again - no reasonable issue description whatsoever. The title states "and use tools.mk" but it's not clear from the changes how this is used and not explained anywhere else.

Now issue this PR refers to has the required description.

dkistner:
> how this is used and not explained anywhere else.
Actually also not clear to me. Can you explain?

acumino:
> Actually also not clear to me. Can you explain?

In the description, I have added the related issue in gardener/gardener also the issue in gardener/gardener has related PR where the original motivation for change is described.
In general, we are importing `tools.mk`  file in Makefile here https://github.com/gardener/gardener-extension-provider-aws/blob/555c27765ece50ab9f0d50f116e1f0f3e26b962c/Makefile#L41-L46
And tools.mk has desired `make` rules for installing required libraries. In our makefile for make rules we are installing the required libraries like this
https://github.com/gardener/gardener-extension-provider-aws/blob/555c27765ece50ab9f0d50f116e1f0f3e26b962c/Makefile#L127
Her make check requires `goimports` and `golangci_lint` which will be installed just before running the make check rules and make rules for these can be found here in tools.mk file https://github.com/gardener/gardener-extension-provider-aws/blob/555c27765ece50ab9f0d50f116e1f0f3e26b962c/vendor/github.com/gardener/gardener/hack/tools.mk#L110-L113

dkistner (approved):

PR #587, opened 2022-08-08
Don't deploy `PSP`s when `PodSecurityPolicy` plugin is disabled 

shafeeqes:

/area open-source
/kind enhancement
/platform aws

`PodSecurityPolicy` is deprecated and will be disabled in kubernetes v1.25+.
End-users are provided an option to migrate their PSPs before upgrading and disable the `PodSecurityPolicy` admission plugin in the ShootSpec. 
In that case, we should stop deploying our PSPs as well.

- This PR stops deploying PSPs related to `provider-aws` if the plugin is disabled in the Shoot.

Part of https://github.com/gardener/gardener/issues/5250

**Release note**:

```breaking operator
Please make sure you're running gardener@v1.52 or above before upgrading to this version.
```

### DIFF
--- a/charts/internal/shoot-system-components/charts/csi-driver-node/templates/clusterrole-csi-driver.yaml
+++ b/charts/internal/shoot-system-components/charts/csi-driver-node/templates/clusterrole-csi-driver.yaml
@@ -16,7 +16,9 @@ rules:
 - apiGroups: ["storage.k8s.io"]
   resources: ["volumeattachments"]
   verbs: ["get", "list", "watch", "update", "patch"]
+{{- if not .Values.pspDisabled }}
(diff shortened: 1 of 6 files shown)


dkistner (dismissed):

dkistner:
Looks good but some tests need adaption.

dkistner:
Shall the release note not rather `breaking operator` as it is only compatible with a g/g @v1.52 or higher?

shafeeqes:
> Shall the release note not rather `breaking operator` as it is only compatible with a g/g @v1.52 or higher?

Corrected. Thanks!

dkistner (approved):

PR #589, opened 2022-08-16
Create shoot cluster per default without overlay network for kubernetes >= 1.22.

ScheererJ:

/area networking
/kind enhancement
/platform aws

Create shoot cluster per default without overlay network for kubernetes >= 1.22.

Unless explicitly specified, the overlay network is disabled with a mutating
webhook for new clusters. This only works for clusters >= 1.22 due to the
source/destinations checks being disabled only for those clusters (see
https://github.com/gardener/machine-controller-manager-provider-aws/issues/36
for details).

Prerequisite for https://github.com/gardener/backlog/issues/29.

None.

This requires the routes for the per node pod cidrs to be created in the infrastructure as a prerequisite. This could be done by cloud-controller-manager. Unfortunately, aws cloud-controller-manager does not work with multiple route tables, which are used by gardener (one per availability zone + a global one).
Therefore, this pull request is a draft for now.
https://github.com/gardener/gardener-extension-provider-aws/pull/591 is the prerequisite for this change.

**Release note**:

```other operator
All new calico aws shoot clusters with kubernetes >= 1.22 will be created without an overlay if not explicitly specified in the shoot spec.
```


dkistner (approved):
/lgtm

PR #592, opened 2022-08-18
Set seccomp profile to RuntimeDefault for csi-driver-node

dimityrmirchev:

/area security
/kind enhancement
/platform aws

This PR enhances the `securityContext` of the `csi-driver-node` pods by adding a seccomp profile.

Fixes #

**Release note**:

```other operator
The `csi-driver-node` daemonset now have its seccomp profile set to "RuntimeDefault".
```

### DIFF
--- a/charts/internal/shoot-system-components/charts/csi-driver-node/templates/daemonset.yaml
+++ b/charts/internal/shoot-system-components/charts/csi-driver-node/templates/daemonset.yaml
@@ -33,6 +33,10 @@ spec:
         runAsUser: 0
         runAsGroup: 0
         fsGroup: 0
+        {{- if semverCompare ">= 1.19" .Capabilities.KubeVersion.GitVersion }}
+        seccompProfile:
+          type: RuntimeDefault
+        {{- end }}
       containers:
       - name: csi-driver
         image: {{ index .Values.images "csi-driver" }}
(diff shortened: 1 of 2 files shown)


charts/internal/shoot-system-components/charts/csi-driver-node/templates/daemonset.yaml
```diff
@@ -28,6 +28,11 @@ spec:
         operator: Exists
       - effect: NoExecute
         operator: Exists
+      {{- if semverCompare ">= 1.19" .Capabilities.KubeVersion.GitVersion }}
+      securityContext:
+        seccompProfile:
+          type: RuntimeDefault
```
ialidzhikov, line 34: Why we set to `RuntimeDefault` when the CSI drivers run in privileged mode (and will continue to run in privileged mode; there is now way to run them in non-privileged mode) and the actually used profile for privileged containers is Unconfined?

ialidzhikov (changes requested):

charts/internal/shoot-system-components/charts/csi-driver-node/templates/daemonset.yaml
dimityrmirchev: Yes, for the `csi-driver` container this is correct. In the pod definition there are two more containers which do not run in privileged mode (`csi-liveness-probe` and `csi-node-driver-registrar`). They will have their seccomp profile changed Unconfined -> RuntimeDefault. Also specifying the `RuntimeDefault` profile on pod spec level acts as guardrails against:
1. Adding a new container and running it as `Unconfined`
2. Removing `privileged: true` from a container definition and still running as `Unconfined`

ialidzhikov: Ok, thanks for explaining! 👍

ialidzhikov (dismissed):
/lgtm

kon-angelo (approved):
/lgtm

PR #594, opened 2022-08-20
update ebs driver

kon-angelo:

/area control-plane
/kind enhancement
/platform aws

Fixes #

**Release note**:

```other operator
Update EBS CSI driver to v1.11.2
```

### DIFF
--- a/charts/images.yaml
+++ b/charts/images.yaml
@@ -64,7 +64,7 @@ images:
 - name: csi-driver
   sourceRepository: github.com/kubernetes-sigs/aws-ebs-csi-driver
   repository: public.ecr.aws/ebs-csi-driver/aws-ebs-csi-driver
-  tag: "v1.9.0"
+  tag: "v1.11.2"
 - name: csi-provisioner
   sourceRepository: github.com/kubernetes-csi/external-provisioner
   repository: k8s.gcr.io/sig-storage/csi-provisioner


charts/images.yaml
```diff
@@ -64,7 +64,7 @@ images:
 - name: csi-driver
   sourceRepository: github.com/kubernetes-sigs/aws-ebs-csi-driver
   repository: public.ecr.aws/ebs-csi-driver/aws-ebs-csi-driver
-  tag: "v1.9.0"
+  tag: "v1.11.2"
```
ialidzhikov, line 67: In https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/1333 the securityContext in the upstream chart is improved. Can you reflect that changes to our charts?

ialidzhikov (changes requested):

charts/images.yaml
kon-angelo: @ialidzhikov  Can you PTAL ?

charts/internal/shoot-system-components/charts/csi-driver-node/values.yaml
```diff
@@ -37,3 +37,19 @@ resources:
...
 pspDisabled: false
+securityContext:
+  # The node pod must be run as root to bind to the registration/driver sockets
+  runAsNonRoot: false
+  runAsUser: 0
+  runAsGroup: 0
+  fsGroup: 0
+containerSecurityContext:
+  node:
+    readOnlyRootFilesystem: true
```
ialidzhikov, line 48: Can some of these changes be reflected to the PodSecurityPolicy?

For example now we have `readOnlyRootFilesystem=true` for all containers. Can the corresponding field in the PSP be set to `true`? 

https://github.com/gardener/gardener-extension-provider-aws/blob/9a70cad0e269a8e3d307e86d083d419abc325673/charts/internal/shoot-system-components/charts/csi-driver-node/templates/podsecuritypolicy.yaml#L29

charts/internal/seed-controlplane/charts/csi-driver-controller/values.yaml
```diff
@@ -74,3 +74,25 @@ csiSnapshotValidationWebhook:
   secrets:
     server: csi-snapshot-validation-server
 
+containerSecurityContext:
+  provisioner:
+    readOnlyRootFilesystem: true
+    allowPrivilegeEscalation: false
```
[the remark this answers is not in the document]
kon-angelo: > Shouldn't this be rather
This makes sense only if you plan to add more stuff under each container and e.g. move all of https://github.com/gardener/gardener-extension-provider-aws/blob/5e81cc71bdc040d25c7f9442db3444445ce4f1d1/charts/internal/seed-controlplane/charts/csi-driver-controller/values.yaml#L17 under this format too and refactor their usages. With the way the values.yaml file is setup ATM it made more sense to make it like it is.

But for your second question, I would say that we could hardcode it as I do not see that much benefit in having them configurable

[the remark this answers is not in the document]
ialidzhikov: > But for your second question, I would say that we could hardcode it as I do not see that much benefit in having them configurable

Agree. Let's hardcode them for consistency reasons. It seems that currently we mostly make things configurable when we need them to be configurable (values passed from the control plane actuator).

charts/internal/shoot-system-components/charts/csi-driver-node/values.yaml
```diff
@@ -36,4 +36,4 @@ resources:
     limits:
       memory: 300Mi
 
-pspDisabled: false
+pspDisabled: false
```
ialidzhikov, line 39: Nit: In my previous review I commented about the empty trailing lines because there were more than 1. I think it is accepted to leave 1 empty line at the end of the file. You could revert the no-op change to this file.

ialidzhikov (approved):
/lgtm
/squash

PR #602, opened 2022-09-09
Update golang for builder image and linter to 1.19 

MartinWeindel:
- update golang for builder image: `1.18.3` -> `1.19.2`
- update golangci_lint:  `1.45.2` -> `1.49` to prepare switching to golang 1.19 for verify step

**Release note**:

```other operator
Update golang for builder image: `1.18.3` -> `1.19.2`
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,6 +43,8 @@ SECRET_ACCESS_KEY_FILE := .kube-secrets/aws/secret_access_key.secret
 #########################################
 
 TOOLS_DIR := hack/tools
+# temporarily until Gardener version has been revendored >= v1.54
+GOLANGCI_LINT_VERSION := v1.49.0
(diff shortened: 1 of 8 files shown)


Makefile
```diff
@@ -43,6 +43,8 @@ SECRET_ACCESS_KEY_FILE := .kube-secrets/aws/secret_access_key.secret
 #########################################
 
 TOOLS_DIR := hack/tools
+# temporarily until Gardener version has been revendored >= v1.54
```
acumino, line 46: This is not required now, the latest vendored gardener version is 1.56

kon-angelo: Still it doesn't harm to use a later golangci-lint version. We can remove it later.

kon-angelo (approved):
/lgtm

PR #612, opened 2022-09-19
Enable custom aws route controller manager if no provider config is specified.

ScheererJ:

/area networking
/kind bug
/platform aws

Enable custom aws route controller manager if no provider config is specified.

The custom aws route controller manager is essential to the shoot cluster working without overlay network. Therefore, it needs to be enabled regardless if there is a provider configuration or not. Otherwise, shoot clusters will get no overlay, but no pod routes.

None.

**Release note**:

```other operator
Correctly enable aws custom route controller if required to ensure overlay free cluster operation.
```

### DIFF
--- a/pkg/webhook/controlplane/ensurer.go
+++ b/pkg/webhook/controlplane/ensurer.go
@@ -667,25 +667,25 @@ func (c *controlPlaneMutator) Mutate(ctx context.Context, new, old client.Object
 		if greaterEqual122 {
 			switch x.Name {
 			case cluster.Shoot.Name:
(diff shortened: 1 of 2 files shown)


rfranzke (commented):
Could you also add a unit test for this case so that we don't regress it in the future? :)

dkistner (dismissed):

ScheererJ:
> Could you also add a unit test for this case so that we don't regress it in the future? :)

A corresponding test has been added.

rfranzke (approved):
/lgtm

dkistner (approved):

PR #617, opened 2022-09-23
Scale down if custom-route-controller is disabled after creation

MartinWeindel:

/area control-plane
/kind bug
/platform aws

If a shoot is created with `aws-custom-route-controller` enabled and the controller is disabled later by setting `cloudControllerManager.useCustomRouteController=false` in the `ControlPlaneConfig`, the deployment is not removed.
The control plane deployment cannot deal with such cases. Looks like a minor edge case to me, therefore the fix just scales down the deployment.

Additionally a non-relevant typo was fixed in `charts/internal/seed-controlplane/values.yaml` 

**Release note**:

```other operator
NONE
```

### DIFF
--- a/charts/internal/seed-controlplane/values.yaml
+++ b/charts/internal/seed-controlplane/values.yaml
@@ -5,5 +5,5 @@ cloud-controller-manager:
   enabled: true
 csi-driver-controller:
   enabled: false
-aws-custom-routes-controller:
+aws-custom-route-controller:
   enabled: false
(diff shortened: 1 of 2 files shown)


shafeeqes:
Is it not cleaner to not deploy the corresponding charts if the controller is disabled? based on the `enabled` field?

shafeeqes:
Okay, this chart is not part of any managed resource? Then my suggestion won't probably work because the resources won't be cleaned up.

dkistner (approved):
/lgtm

kon-angelo (approved):
/lgtm

ialidzhikov:
FYI, this PR does not fix/mitigate anything - see https://github.com/gardener/gardener-extension-provider-aws/issues/628.

PR #630, opened 2022-10-11
Fix infrastructure tests after gardener v1.56.0

kon-angelo:

/area control-plane
/kind bug
/platform aws

After revendoring of gardener `v1.56.0` the priority class is necessary for the terraformer pod.

/invite @shafeeqes 

Fixes #

**Release note**:

```other operator

```

### DIFF
--- a/test/integration/infrastructure/infrastructure_test.go
+++ b/test/integration/infrastructure/infrastructure_test.go
@@ -32,6 +32,7 @@ import (
 	"github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/aws/aws-sdk-go/service/iam"
 	gardencorev1beta1 "github.com/gardener/gardener/pkg/apis/core/v1beta1"
+	v1beta1constants "github.com/gardener/gardener/pkg/apis/core/v1beta1/constants"
 	gardencorev1beta1helper "github.com/gardener/gardener/pkg/apis/core/v1beta1/helper"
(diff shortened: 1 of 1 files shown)


test/integration/infrastructure/infrastructure_test.go
```diff
@@ -313,6 +316,16 @@ var _ = Describe("Infrastructure tests", func() {
...
+			priorityClass = &schedulingv1.PriorityClass{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: v1beta1constants.PriorityClassNameShootControlPlane300,
+				},
+				Description:   "PriorityClass for Shoot control plane components",
+				GlobalDefault: false,
+				Value:         999998300,
+			}
+			Expect(client.IgnoreAlreadyExists(c.Create(ctx, priorityClass))).To(BeNil())
+
```
shafeeqes, line 328: Shouldn't this be inside the `runTest` function here,
https://github.com/gardener/gardener-extension-provider-aws/blob/5e67cfc2354fdd1017479452589893894f7b4201/test/integration/infrastructure/infrastructure_test.go#L367
I see lot of code is duplicated here which is in runTest, can we reuse them?

dkistner: Can we not simply put this logic into `BeforeSuite`? It's a global/non namespaced resource, which is needed by multiple test cases.

shafeeqes: Yes. Better.

dkistner (approved):

Issue #628, opened 2022-10-04
Scale down of aws-custom-route-controller is not working when it is disabled


ialidzhikov:
**How to categorize this issue?**

/area control-plane
/kind bug
/platform aws

**What happened**:
Shoots created with `provider-aws@v1.38.1` (hence affected by https://github.com/gardener/gardener-extension-provider-aws/issues/619) have their aws-route-controller in `CrashLoopBackOff`:

```
$ k get po -A -l role=aws-custom-route-controller
NAMESPACE                        NAME                                           READY   STATUS             RESTARTS          AGE
<omitted>                        aws-custom-route-controller-56bd5c96dd-lt9b8   0/1     CrashLoopBackOff   112 (2m28s ago)   9h
<omitted>                        aws-custom-route-controller-69b6484dd-6xxjx    0/1     CrashLoopBackOff   253 (58s ago)     21h
<omitted>                        aws-custom-route-controller-5b7fd48466-hsk9w   0/1     CrashLoopBackOff   168 (115s ago)    14h
<omitted>                        aws-custom-route-controller-7d45d99468-xnhgw   0/1     Running            161 (8s ago)      9h
<omitted>                        aws-custom-route-controller-7dbf58fd6-52ltx    0/1     CrashLoopBackOff   33 (2m23s ago)    102m
<omitted>                        aws-custom-route-controller-6cccdd7cdb-87kml   0/1     Running            14 (5m23s ago)    37m
<omitted>                        aws-custom-route-controller-59668966cb-999jp   0/1     CrashLoopBackOff   210 (113s ago)    18h
<omitted>                        aws-custom-route-controller-545f9b976-66vcw    0/1     CrashLoopBackOff   209 (3m1s ago)    18h
<omitted>                        aws-custom-route-controller-57b4fbddf6-8f9jj   0/1     CrashLoopBackOff   211 (2m57s ago)   18h
<omitted>                        aws-custom-route-controller-5bff8bc576-gjtzc   0/1     Running            183 (5m18s ago)   15h
<omitted>                        aws-custom-route-controller-7f5fd4744-r2866    0/1     Running            301 (22s ago)     17h
<omitted>                        aws-custom-route-controller-65866fdccb-c57vg   0/1     CrashLoopBackOff   300 (4m52s ago)   16h
<omitted>                        aws-custom-route-controller-857567754-pspxf    0/1     CrashLoopBackOff   225 (4m42s ago)   12h
[...]

PR #634, opened 2022-10-21
Scale down of aws-custom-route-controller

elchead:

/area TODO
/kind TODO
/platform aws

Fixes https://github.com/gardener/gardener-extension-provider-aws/issues/628

**Release note**:

```other operator

```

### DIFF
--- a/pkg/controller/controlplane/valuesprovider.go
+++ b/pkg/controller/controlplane/valuesprovider.go
@@ -530,10 +530,9 @@ func getCRCChartValues(
 	enabled := cpConfig.CloudControllerManager != nil &&
 		cpConfig.CloudControllerManager.UseCustomRouteController != nil &&
 		*cpConfig.CloudControllerManager.UseCustomRouteController
-
 	values := map[string]interface{}{
 		"enabled":     enabled,
-		"replicas":    extensionscontroller.GetControlPlaneReplicas(cluster, scaledDown || !enabled, 1),
+		"replicas":    extensionscontroller.GetControlPlaneReplicas(cluster, scaledDown, 1),
 		"clusterName": cp.Namespace,
 		"podNetwork":  extensionscontroller.GetPodNetwork(cluster),
 		"podAnnotations": map[string]interface{}{
@@ -543,6 +542,9 @@ func getCRCChartValues(
 			v1beta1constants.LabelPodMaintenanceRestart: "true",
 		},
 		"region": cp.Spec.Region,
+	}
+	if !enabled {
+		values["replicas"] = 0
 	}
 
 	return values, nil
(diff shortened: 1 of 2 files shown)


gardener-robot-ci-3:
Thank you @elchead for your contribution. Before I can start building your PR, a member of the organization must set the required label(s) {'reviewed/ok-to-test'}. Once started, you can check the build status in the PR checks section below.

MartinWeindel (changes requested):
Thanks, looks good to me.
I have only one small change request. There is another test case named "should return correct control plane chart values (k8s >= 1.18)" (valuesprovider_test.go, line 296ff). Can you check here for replicas = 0?

elchead:
@MartinWeindel i am doing this already see l.257 in the definition of `crcChartValues`

DockToFuture:
Didn't know that there was a PR open on this, just discovered it. Did the same thing in my code to remove the overlay but also checked if the `UseCustomRouteController` field is nil and then disabled the awsRouteController. This will keep it disabled for existing clusters as long as they specify no overlay.
```
	awsCustomRouteControllerEnabled := false
	replicas := extensionscontroller.GetControlPlaneReplicas(cluster, scaledDown, 1)
	if cpConfig.CloudControllerManager != nil &&
		cpConfig.CloudControllerManager.UseCustomRouteController != nil {
		awsCustomRouteControllerEnabled = true
	}
	if awsCustomRouteControllerEnabled && !*cpConfig.CloudControllerManager.UseCustomRouteController {
		replicas = 0
	}

	values := map[string]interface{}{
		"enabled":     awsCustomRouteControllerEnabled,
		"replicas":    replicas,
...
```
Do you prefer to enable the aws route controller always or only for clusters which have the UseCustomRouteController field specified?

kon-angelo:
@DockToFuture AFAIK, the problem is that without setting the `enabled` field the replica update won't be propagated because the chart will not be applied. 

Ideally, we would be able to discern if the chart had already been applied at least once (hence there was a CRC running at some point) and only then apply the replicas, but I think this complicates the flow. Since we want to support disabling CRC and effectively setting replicas to 0 (since we don't delete the deployment), why not have only this case to care about.

DockToFuture:
> @DockToFuture AFAIK, the problem is that without setting the `enabled` field the replica update won't be propagated because the chart will not be applied.

Sure without setting the enabled field the chart won't be applied. If we want to keep it like this im fine with it. Can we get it merged then I will rebase my PR on this.

DockToFuture (approved):
/lgtm

kon-angelo (approved):
/lgtm

MartinWeindel (approved):
/lgtm

PR #644, opened 2022-11-11
Add validation for `domain-name` of custom dhcp options used in VPCs.

### LINKED ISSUE #615: Add validation for `domain-name` of custom dhcp options used in VPCs

/area ops-productivity
/area quality
/kind bug
/platform aws

Extend the infrastructure config validator so that it also checks if a VPC's [dhcp options set](https://docs.aws.amazon.com/vpc/latest/userguide/VPC_DHCP_Options.html) `domain-name` key has the correct value corresponding to the region where the nodes are created. The correct value for the `domain-name` for the `us-east-1` is ec2.internal 

**Why is this needed**:
When users specify a VPC ID in the shoot spec, the referenced VPC can use custom dhcp options. The `domain-name` value inside these options could be missing which leads to the following problem: https://github.com/kubernetes/cloud-provider-aws/issues/384

An incorrect dhcp options would look like this:
```
{
    "DhcpOptions": [
        {
            "DhcpConfigurations": [
                {
                    "Key": "domain-name-servers",
                    "Values": [
                        {
                            "Value": "AmazonProvidedDNS"
                        }
                    ]
                }
            ],
            "DhcpOptionsId": "dopt-546be43d",
            "OwnerId": "008447767675",
            "Tags": []
        }
    ]
}
```
A correct one would look like this:
```
{
    "DhcpOptions": [
        {
            "DhcpConfigurations": [
                {
                    "Key": "domain-name",
                    "Values": [
                        {
                            "Value": "us-west-2.compute.internal"
                        }
                    ]
                },
                {
                    "Key": "domain-name-servers",
                    "Values": [
                        {
                            "Value": "AmazonProvidedDNS"
                        }
                    ]
                }
            ],
            "DhcpOptionsId": "dopt-546be43d",
            "OwnerId": "008447767675",
            "Tags": []
        }
    ]
}
```

Additionally, when setting the `domain-name` key field to some non-default value (e.g. `compute.custom`) a similar problem appears - the nodes will be created in the cluster with names like `ip-10-180-25-59.custom.domain`, however the `Private IP DNS` name in the instance resource on AWS side will still be with a default value - `ip-10-180-29-19.eu-west-2.compute.internal` and the machines will again fail to join the cluster. 

The dhcp options id can be retrieved from the VPC's `DhcpOptionsId` field.

The validation that we introduce could be only temporary depending on whether/how/if https://github.com/kubernetes/cloud-provider-aws/issues/384 will be fixed

AleksandarSavchev:

/area ops-productivity
/area quality
/kind bug
/platform aws

This PR adds validation for `domain-name` of custom dhcp options used in VPCs.

Fixes #615

**Release note**:

```bugfix operator
Added validation for `domain-name` of custom dhcp options used in VPCs
```

### DIFF
--- a/pkg/aws/client/client.go
+++ b/pkg/aws/client/client.go
@@ -16,6 +16,7 @@ package client
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"time"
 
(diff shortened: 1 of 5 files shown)


pkg/controller/infrastructure/configvalidator_test.go
```diff
@@ -188,6 +198,62 @@ var _ = Describe("ConfigValidator", func() {
...
 
+		Describe("validate DHCP options", func() {
+
+			BeforeEach(func() {
+				awsClient.EXPECT().GetVPCAttribute(ctx, vpcID, "enableDnsSupport").Return(true, nil)
+				awsClient.EXPECT().GetVPCAttribute(ctx, vpcID, "enableDnsHostnames").Return(true, nil)
+				awsClient.EXPECT().GetVPCInternetGateway(ctx, vpcID).Return(vpcID, nil)
+			})
+
+			It("should allows VPC with correctly configurated DHCP options", func() {
```
plkokanov, line 209: Can you add a tests that cover the case when the region is `us-east-1`?
Additionally, it would be better to use a DescribeTable() here to avoid code repetition.

plkokanov, line 209:
```suggestion
			It("should allow VPC with correctly configurated DHCP options", func() {
```

plkokanov (changes requested):

pkg/aws/client/client.go
```diff
@@ -222,6 +222,58 @@ func (c *Client) GetVPCAttribute(ctx context.Context, vpcID string, attribute st
...
+	describeDhcpOptionsOutput, err := c.EC2.DescribeDhcpOptionsWithContext(ctx, describeDhcpOptionsInput)
+	if err != nil {
+		return nil, err
+	}
+
+	result := make(map[string]string)
+	if len(describeDhcpOptionsOutput.DhcpOptions) > 0 {
+		for _, dhcpConfiguration := range describeDhcpOptionsOutput.DhcpOptions[0].DhcpConfigurations {
+			if dhcpConfiguration.Key != nil && *dhcpConfiguration.Key == "domain-name" && dhcpConfiguration.Values != nil {
+				if len(dhcpConfiguration.Values) > 0 && dhcpConfiguration.Values[0].Value != nil {
```
plkokanov, line 268: You could use a single if statement here to avoid nesting.
```suggestion
				if dhcpConfiguration.Key != nil && *dhcpConfiguration.Key == "domain-name" && len(dhcpConfiguration.Values) > 0 &&  dhcpConfiguration.Values[0].Value != nil {
```

plkokanov (changes requested):
Just one minor nit left.

plkokanov (approved):

kon-angelo (approved):
/lgtm

PR #650, opened 2022-11-22
Use provider specific regex to parse error codes from error

acumino:

/area open-source
/kind enhancement
/platform aws

This PR removes extension dependency on g/g to extract error codes from errors and uses regex defined in the extension repo. Later extensions can be adapted to map provider-specific error codes to gardener error codes.

Part of https://github.com/gardener/gardener/issues/5444

Follow Up of - https://github.com/gardener/gardener-extension-provider-aws/pull/629

**Release note**:

```other operator
Provider-specific error codes are now detected/parsed on provider-extension side.
```


pkg/apis/aws/helper/errors.go
```diff
@@ -0,0 +1,101 @@
...
+		}
+	}
+
+	// compute error code list based on code string set
+	var out []gardencorev1beta1.ErrorCode
+	for _, c := range codes.List() {
+		out = append(out, gardencorev1beta1.ErrorCode(c))
+	}
+	return out
+}
```
rfranzke, line 101: Why do we still have these functions in the light of https://github.com/gardener/gardener/pull/6912?

acumino: It is draft I have not pushed recent changes, I was meanwhile looking into something else. I will push the change with the addressing review from the last PR so overall review effort is less.

rfranzke: What is the value of such draft PR when it is not adapted at all? Typically, draft PRs can be used to get early feedback, but in this case this was not even your intention. Anyways ...

rfranzke (dismissed):
/lgtm

pkg/apis/aws/helper/error_codes.go
```diff
@@ -0,0 +1,46 @@
...
+	quotaExceededRegexp                 = regexp.MustCompile(`(?i)((?:^|[^t]|(?:[^s]|^)t|(?:[^e]|^)st|(?:[^u]|^)est|(?:[^q]|^)uest|(?:[^e]|^)quest|(?:[^r]|^)equest)LimitExceeded|Quotas|Quota.*exceeded|exceeded quota|Quota has been met|QUOTA_EXCEEDED|Maximum number of ports exceeded|ZONE_RESOURCE_POOL_EXHAUSTED_WITH_DETAILS|VolumeSizeExceedsAvailableQuota)`)
+	rateLimitsExceededRegexp            = regexp.MustCompile(`(?i)(RequestLimitExceeded|Throttling|Too many requests)`)
+	dependenciesRegexp                  = regexp.MustCompile(`(?i)(PendingVerification|Access Not Configured|accessNotConfigured|DependencyViolation|OptInRequired|DeleteConflict|Conflict|inactive billing state|ReadOnlyDisabledSubscription|is already being used|InUseSubnetCannotBeDeleted|VnetInUse|InUseRouteTableCannotBeDeleted|timeout while waiting for state to become|InvalidCidrBlock|already busy for|InsufficientFreeAddressesInSubnet|InternalServerError|internalerror|internal server error|A resource with the ID|VnetAddressSpaceCannotChangeDueToPeerings|InternalBillingError|There are not enough hosts available)`)
```
ialidzhikov, line 28: Is the list of error codes filtered to contain only AWS specific error strings? Because I see non-AWS error strings.

ialidzhikov: `InvalidSubscriptionId` - subscription is Azure specific term, hence it could only occur for Azure Shoots
`SERVICE_ACCOUNT_ACCESS_DENIED` - checking https://github.com/gardener/gardener/issues/5364, it could occur only for GCP Shoots
`ZONE_RESOURCE_POOL_EXHAUSTED_WITH_DETAILS` - checking our internal ops-guide, it is GCP specific
`VolumeSizeExceedsAvailableQuota` - checking our internal ops-guide, it is OpenStack specific
`ReadOnlyDisabledSubscription` - it could occur only for Azure as subscription is Azure specific term
`VnetInUse` - vnet is only Azure specific term
`VnetAddressSpaceCannotChangeDueToPeerings` - vnet is only Azure specific term
`SkuNotAvailable` - SKU is Azure specific term

These are the ones that I noticed at first look.
You can check the error messages using:
- the git history - maybe the PR that added the error string reveals hints on which infra it occurred
- the internal ops-guide
- the Shoots on our landscapes

There are also a lot gardener specific ones such as `KubeletHasInsufficientMemory`, `KubeletHasDiskPressure`, `KubeletHasInsufficientPID`. What we do about them - do we duplicate them on every extension?

acumino: I have updated the list.
>There are also a lot gardener specific ones such as KubeletHasInsufficientMemory, KubeletHasDiskPressure, KubeletHasInsufficientPID. What we do about them - do we duplicate them on every extension?

Yes, Otherwise there will be a need to handle it again in gardener/gardener.

ialidzhikov (approved):
Duplicating the approval from https://github.com/gardener/gardener-extension-provider-aws/pull/650#pullrequestreview-1202551643 to not be the guy that blocks this PR. I have no capacity to check this PR further.

pkg/apis/aws/helper/error_codes.go
kon-angelo: But doesn't it make sense to handle the kubelet errors centrally rather than in the extensions ? Does it really help if we duplicate them now ?

acumino: I have dropped kublet specific error codes. PTAL.

kon-angelo (approved):
/lgtm

PR #657, opened 2022-11-25
Do not fail a request for a Shoot from the core.gardener.cloud/v1alpha1 API

ialidzhikov:

/area quality
/kind regression
/kind bug
/platform aws

After https://github.com/gardener/gardener-extension-provider-aws/pull/635 CREATE/UPDATE requests for Shoots from the core.gardener.cloud/v1alpha1 API fail with:
```
$ k create -f ~/foo.yaml
Error from server: error when creating "foo.yaml": admission webhook "mutation.aws.provider.extensions.gardener.cloud" denied the request: unexpected request kind core.gardener.cloud/v1alpha1, Kind=Shoot
```

This PR removes the v1alpha1 from the mutatingwebhook spec and relies on the `matchPolicy: Equivalent` field of the webhook that will take care to convert the v1alpha1 obj to v1beta1. Ref https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#matching-requests-matchpolicy

Similar to https://github.com/gardener/gardener-extension-provider-alicloud/pull/341

This PR additionally removes RBAC rule for `shoots` as admission-aws does not get/list/watch Shoots.

See above

**Release note**:

```bugfix operator
An issue causing admission-aws to fail CREATE/UPDATE request to the core.gardener.cloud/v1alpha1 API is now fixed.
```

### DIFF
--- a/charts/gardener-extension-admission-aws/charts/application/templates/mutatingwebhook-mutator.yaml
+++ b/charts/gardener-extension-admission-aws/charts/application/templates/mutatingwebhook-mutator.yaml
@@ -8,7 +8,6 @@ webhooks:
   - apiGroups:
     - "core.gardener.cloud"
     apiVersions:
-    - v1alpha1
     - v1beta1
     operations:
     - CREATE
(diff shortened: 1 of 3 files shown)


kon-angelo (approved):
/lgtm

Issue #659, opened 2022-11-25
admission-aws mutates migrating/restoring Shoots -> leads to a deadlock on Shoot migration/restore


ialidzhikov:
**How to categorize this issue?**

/area control-plane-migration
/kind bug
/platform aws

**What happened**:
After https://github.com/gardener/gardener-extension-provider-aws/pull/635 (and similar PR on other providers) we have ~20 Shoots on various providers failing to restore with:
```
task "Waiting until shoot infrastructure has been reconciled" failed: shoots.core.gardener.cloud "foo" is forbidden: cannot change shoot spec during Restore operation that is in state Processing
```

After investigation I see that the reason for this is the new mutator introduced in https://github.com/gardener/gardener-extension-provider-aws/pull/635. The mutator tries to add the following part
```yaml
spec:
  networking:
    providerConfig:
      backend: null
```

to the already restoring Shoot. This is a Shoot spec change and Shoot spec changes are forbidden by the gardener-apiserver during migrate/restore.

**What you expected to happen**:
admission-aws should not try to set `spec.networking.providerConfig` for migration/restoring Shoot as Shoot spec updates are forbidden during restore/migrate. Such action leads to a deadlock and Shoot cannot be migrated/restored and gets stuck in this state.

**How to reproduce it (as minimally and precisely as possible)**:
1. Trigger control plane migration.

2. Deploy `provider-aws@v1.40.0`

3. Make sure restore operation fails with:
```
task "Waiting until shoot infrastructure has been reconciled" failed: shoots.core.gardener.cloud "tmn33-zb3-hs-mgr" is forbidden: cannot change shoot spec during Restore operation that is in state Processing
```

**Anything else we need to know?**:
N/A

**Environment**:

- Gardener version (if relevant):
- Extension version: v1.40.0
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
- Others:

ialidzhikov:
cc @DockToFuture @plkokanov

PR #661, opened 2022-11-28
mutating webhook respects shoot deletion, restore and migration

DockToFuture:

/area networking
/kind bug
/platform aws

Mutating webhook respects shoot deletion, restore and migration.Before those operations failed under certain conditions. E.g.: 
```
task "Waiting until shoot infrastructure has been reconciled" failed: shoots.core.gardener.cloud "tmn33-zb3-hs-mgr" is forbidden: cannot change shoot spec during Restore operation that is in state Processing
```

Fixes https://github.com/gardener/gardener-extension-provider-aws/issues/659

**Release note**:

```other operator
Mutating webhook respects shoot deletion, restore and migration. Before those operations failed under certain conditions.
```

### DIFF
--- a/pkg/admission/mutator/mutator_suite_test.go
+++ b/pkg/admission/mutator/mutator_suite_test.go
@@ -1,0 +1,27 @@
+// Copyright (c) 2022 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package mutator_test
+
+import (
+	"testing"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+func TestMutator(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Shoot Mutator Suite")
+}
--- a/pkg/admission/mutator/shoot.go
+++ b/pkg/admission/mutator/shoot.go
@@ -55,8 +55,30 @@ func (s *shoot) Mutate(ctx context.Context, new, old client.Object) error {
 		return fmt.Errorf("wrong object type %T", new)
 	}
 
-	// source/destination checks are only disabled for kubernetes >= 1.22
-	// see https://github.com/gardener/machine-controller-manager-provider-aws/issues/36 for details
+	// Skip if shoot is in restore or migration phase
+	if wasShootRescheduledToNewSeed(shoot) {
+		return nil
+	}
+
+	var oldShoot *gardencorev1beta1.Shoot
+	if old != nil {
+		oldShoot, ok = old.(*gardencorev1beta1.Shoot)
+		if !ok {
+			return fmt.Errorf("wrong object type %T", old)
+		}
+	}
+
+	if oldShoot != nil && isShootInMigrationOrRestorePhase(shoot) {
+		return nil
+	}
+
+	// Skip if shoot is in deletion phase
+	if shoot.DeletionTimestamp != nil || oldShoot != nil && oldShoot.DeletionTimestamp != nil {
+		return nil
+	}
+
+	// Source/destination checks are only disabled for kubernetes >= 1.22
+	// See https://github.com/gardener/machine-controller-manager-provider-aws/issues/36 for details
 	greaterEqual122, err := versionutils.CompareVersions(shoot.Spec.Kubernetes.Version, ">=", "1.22")
 	if err != nil {
 		return err
@@ -70,18 +92,11 @@ func (s *shoot) Mutate(ctx context.Context, new, old client.Object) error {
 		return err
 	}
 
-	if old == nil && networkConfig.Overlay == nil {
+	if oldShoot == nil && networkConfig.Overlay == nil {
 		networkConfig.Overlay = overlay
 	}
 
-	if old != nil && networkConfig.Overlay == nil {
-		oldShoot, ok := old.(*gardencorev1beta1.Shoot)
-		if !ok {
-			return fmt.Errorf("wrong object type %T", old)
-		}
-		if oldShoot.DeletionTimestamp != nil {
-			return nil
-		}
+	if oldShoot != nil && networkConfig.Overlay == nil {
 		oldNetworkConfig, err := s.decodeNetworkingConfig(oldShoot.Spec.Networking.ProviderConfig)
 		if err != nil {
 			return err
@@ -90,6 +105,7 @@ func (s *shoot) Mutate(ctx context.Context, new, old client.Object) error {
 			networkConfig.Overlay = oldNetworkConfig.Overlay
 		}
 	}
+
 	shoot.Spec.Networking.ProviderConfig = &runtime.RawExtension{
 		Object: networkConfig,
 	}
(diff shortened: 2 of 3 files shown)


pkg/admission/mutator/shoot.go
```diff
@@ -55,6 +56,28 @@ func (s *shoot) Mutate(ctx context.Context, new, old client.Object) error {
 		return fmt.Errorf("wrong object type %T", new)
 	}
 
+	// skip if shoot is in restore or migration phase
+	if wasShootRescheduledToNewSeed(shoot) {
+		return nil
+	}
```
ialidzhikov, line 62: Q: Why do we need to check for this? Wrt to CP migration, I think this admission should only be interested in the Shoot status to check whether the lastOperation is a restore/migrate in progress.

pkg/admission/mutator/shoot.go
```diff
@@ -55,6 +56,28 @@ func (s *shoot) Mutate(ctx context.Context, new, old client.Object) error {
...
+
+	var oldShoot *gardencorev1beta1.Shoot
+	if old != nil {
+		oldShoot, ok = old.(*gardencorev1beta1.Shoot)
+		if !ok {
+			return fmt.Errorf("wrong object type %T", old)
+		}
+	}
+
+	if oldShoot != nil && isShootInMigrationOrRestorePhase(shoot) && !reflect.DeepEqual(oldShoot.Spec, shoot.Spec) {
```
ialidzhikov, line 72: Q: Why do we need `!reflect.DeepEqual(oldShoot.Spec, shoot.Spec)`? Copy-paste from gardener-apiserver?

pkg/admission/mutator/shoot.go
```diff
@@ -55,6 +56,28 @@ func (s *shoot) Mutate(ctx context.Context, new, old client.Object) error {
 		return fmt.Errorf("wrong object type %T", new)
 	}
 
+	// skip if shoot is in restore or migration phase
+	if wasShootRescheduledToNewSeed(shoot) {
```
ialidzhikov, line 60: Can you add unit tests for the newly introduced checks? It seems that there are no unit tests for this webhook.

pkg/admission/mutator/shoot.go
```diff
@@ -55,6 +56,28 @@ func (s *shoot) Mutate(ctx context.Context, new, old client.Object) error {
...
+		if !ok {
+			return fmt.Errorf("wrong object type %T", old)
+		}
+	}
+
+	if oldShoot != nil && isShootInMigrationOrRestorePhase(shoot) && !reflect.DeepEqual(oldShoot.Spec, shoot.Spec) {
+		return nil
+	}
+
+	// skip if shoot is in deletion phase
```
ialidzhikov, line 76:
```suggestion
	// Skip if shoot is in deletion phase
```

pkg/admission/mutator/shoot.go
```diff
@@ -70,18 +93,12 @@ func (s *shoot) Mutate(ctx context.Context, new, old client.Object) error {
...
-	if old != nil && networkConfig.Overlay == nil {
-		oldShoot, ok := old.(*gardencorev1beta1.Shoot)
-		if !ok {
-			return fmt.Errorf("wrong object type %T", old)
-		}
-		if oldShoot.DeletionTimestamp != nil {
-			return nil
-		}
+	if oldShoot != nil && networkConfig.Overlay == nil {
+
```
ialidzhikov, line 101:
```suggestion
```

ialidzhikov (changes requested):

pkg/admission/mutator/shoot.go
ialidzhikov: Ah, I now see that gardener-apiserver forbids updates in that case - https://github.com/gardener/gardener/blob/bb5f23257344259673ba5987cbec728460c3e062/plugin/pkg/shoot/validator/admission.go#L485-L487. My bad.

pkg/admission/mutator/shoot.go
```diff
@@ -38,6 +38,7 @@ func NewShootMutator() extensionswebhook.Mutator {
 
 type shoot struct {
 	decoder runtime.Decoder
+	client  client.Client
```
ialidzhikov, line 41: Why the client is introduced? It is unused.

```suggestion
```

pkg/admission/mutator/shoot.go
```diff
@@ -46,6 +47,12 @@ func (s *shoot) InjectScheme(scheme *runtime.Scheme) error {
 	return nil
 }
 
+// InjectClient injects the given client into the validator.
+func (s *shoot) InjectClient(client client.Client) error {
+	s.client = client
+	return nil
+}
```
ialidzhikov, line 54: Same as above, `client` is unused and we don't need this func.

```suggestion
```

pkg/admission/mutator/mutator_suite_test.go
```diff
@@ -0,0 +1,27 @@
...
+package mutator_test
+
+import (
+	"testing"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+func TestValidator(t *testing.T) {
```
ialidzhikov, line 24:
```suggestion
func TestShootMutator(t *testing.T) {
```

pkg/admission/mutator/mutator_suite_test.go
```diff
@@ -0,0 +1,27 @@
...
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Mutator Suite")
```
ialidzhikov, line 26:
```suggestion
	RunSpecs(t, "Shoot Mutator Suite")
```

pkg/admission/mutator/shoot_test.go
```diff
@@ -0,0 +1,133 @@
...
+				Expect(err).NotTo(HaveOccurred())
+			})
+		})
+	})
+})
+
+func encode(obj runtime.Object) []byte {
+	data, _ := json.Marshal(obj)
+	return data
+}
```
ialidzhikov, line 133: Seems to be unused.

```suggestion
```

pkg/admission/mutator/shoot_test.go
```diff
@@ -0,0 +1,133 @@
...
+					LastUpdateTime: metav1.Time{Time: metav1.Now().Add(time.Second * -1000)},
+					Progress:       0,
+					Type:           gardencorev1beta1.LastOperationTypeReconcile,
+					State:          gardencorev1beta1.LastOperationStateProcessing,
+				}
+				shoot.Status.SeedName = pointer.String("gcp")
+				err := shootMutator.Mutate(ctx, shoot, nil)
+				Expect(err).NotTo(HaveOccurred())
+			})
+			It("should return nil when shoot is in migration or restore phase", func() {
```
ialidzhikov, line 109: Nit: you could leave 1 empty lines between the `It`s to increase the readability.

```suggestion
			})

			It("should return nil when shoot is in migration or restore phase", func() {
```

pkg/admission/mutator/shoot_test.go
```diff
@@ -0,0 +1,133 @@
...
+				shoot.Status.LastOperation = &gardencorev1beta1.LastOperation{
+					Description:    "test",
+					LastUpdateTime: metav1.Time{Time: metav1.Now().Add(time.Second * -1000)},
+					Progress:       0,
+					Type:           gardencorev1beta1.LastOperationTypeReconcile,
+					State:          gardencorev1beta1.LastOperationStateProcessing,
+				}
+				shoot.Status.SeedName = pointer.String("gcp")
+				err := shootMutator.Mutate(ctx, shoot, nil)
+				Expect(err).NotTo(HaveOccurred())
```
ialidzhikov, line 107: For testing Validators it is usually enough to check the error. But for Mutators, the mutator could actually return nil and still mutate the obj, right? Is it possible to add expectation that the given obj was not mutated? Otherwise we could have passing unit tests but mutations could still happen.
This comment applies also to the other tests.

pkg/admission/mutator/shoot_test.go
```diff
@@ -0,0 +1,133 @@
...
+var _ = Describe("Shoot mutator", func() {
+
+	Describe("#Mutate", func() {
+		const namespace = "garden-dev"
+
+		var (
+			shootMutator extensionswebhook.Mutator
+
+			ctrl  *gomock.Controller
+			c     *mockclient.MockClient
```
ialidzhikov, line 47: These should not be needed after you drop the unused client from the mutator.

pkg/admission/mutator/shoot_test.go
```diff
@@ -0,0 +1,133 @@
...
+					Region: "us-west",
+					Networking: gardencorev1beta1.Networking{
+						Nodes: pointer.String("10.250.0.0/16"),
+						Type:  "calico",
+					},
+				},
+			}
+		})
+
+		Context("Mutate shoot", func() {
```
ialidzhikov, line 96: I would appreciate if you can put a TODO somewhere in this file such as
```
// TODO(DockToFuture): Add tests for all possible cases for the Mutator.
```

and follow up to add tests for all possible cases.

pkg/admission/mutator/shoot_test.go
```diff
@@ -0,0 +1,133 @@
...
+						Workers: []gardencorev1beta1.Worker{
+							{
+								Name: "worker-1",
+								Volume: &gardencorev1beta1.Volume{
+									VolumeSize: "50Gi",
+									Type:       pointer.String(gp2type),
+								},
+								Zones: []string{"zone1"},
+							},
+						},
```
ialidzhikov, line 85: I believe you can drop the workers field as the mutator is not using it, hence it is not needed to specify it.

kon-angelo (approved):
/lgtm

kon-angelo (dismissed):
/lgtm

pkg/admission/mutator/shoot_test.go
ialidzhikov: @DockToFuture , I don't see this particular comment applied. Do you have a reason to not apply it?

ialidzhikov (approved):
/lgtm

Thanks for applying my suggestions!

PR #673, opened 2022-12-16
Check for duplicate zone names in infrastructureconfig

### LINKED ISSUE #643: Validation should fail if multiple zones with the same AZ are specified
Disallow specifying multiple zones which reference the same AZ under `infrastructureConfig.networks.zones` in the Shoot specification.

/area usability
/kind bug
/platform aws

**What happened**:
Currently, it is possible for a Shootowner to specify the following configuration:
```
    infrastructureConfig:
      kind: InfrastructureConfig
      networks:
        vpc:
          cidr: 10.250.0.0/16
        zones:
          - internal: 10.250.112.0/22
            name: eu-central-1a
            public: 10.250.96.0/22
            workers: 10.250.0.0/19
          - internal: 10.250.144.0/22
            name: eu-central-1a
            public: 10.250.128.0/22
            workers: 10.250.32.0/19
          - internal: 10.250.176.0/22
            name: eu-central-1a
            public: 10.250.160.0/22
            workers: 10.250.64.0/19
```

Note: all 3 `zones` are in the same AZ.
3 subnets are created in `eu-central-1a`. The ordering of `infrastructureStatus.vpc.subnets` is not deterministic and this causes `worker.spec.infrastructureProviderStatus.vpc.subnets` to have different subnets at the top of the list. The [extension-code](https://github.com/gardener/gardener-extension-provider-aws/blob/e1c6c7550b15d20c870310734f94091780b36c44/pkg/controller/worker/machines.go#L141) always picks the first one with matching zone (and purpose). The machineClass then will be updated to reference the subnet.

The above is confusing as:
- the subnet in the MachineClass might not correlate to the actual subnet where the node is in
- Whatever subnet is on top will be selected which seems random and unexpected

Note: A change of the subnet in the MachineClass does not cause a new MachineClass object to be created and therefore it also does not cause a node-rollout.

The configuration to not be allowed.

I created a test cluster with the above configuration of `zones` and then I triggered maintenance 8 times and the subnet referenced in the MachineClass changed 3 times.

Because the above described effects are confusing and it is not clear what behaviour is actually expected if multiple zones in the same AZ are specified I would vote for disallowing that via validation in the extension.

A separate issue would be to keep the ordering in the infrastructureStatus stable as it seems unnecessary to switch this ordering at all.

MartinWeindel:

/area control-plane
/kind enhancement
/platform aws

The validation of the `InfrastructureConfig` has been extended by a duplicate name check in `spec.network.zones`

Fixes #643 

**Release note**:

```other operator
Check for duplicate zone names in infrastructureconfig
```

### DIFF
--- a/pkg/apis/aws/validation/infrastructure.go
+++ b/pkg/apis/aws/validation/infrastructure.go
@@ -56,6 +56,7 @@ func validateInfrastructureConfigZones(oldInfra, infra *apisaws.InfrastructureCo
 		awsZones.Insert(awsZone.Name)
 	}
 
+	usedZones := sets.NewString()
 	for i, zone := range infra.Networks.Zones {
 		if oldInfra != nil && len(oldInfra.Networks.Zones) > i && oldInfra.Networks.Zones[i] == zone {
 			continue
@@ -64,6 +65,10 @@ func validateInfrastructureConfigZones(oldInfra, infra *apisaws.InfrastructureCo
 		if !awsZones.Has(zone.Name) {
 			allErrs = append(allErrs, field.NotSupported(fldPath.Child("zones").Index(i).Child("name"), zone.Name, awsZones.UnsortedList()))
 		}
+		if usedZones.Has(zone.Name) {
+			allErrs = append(allErrs, field.Duplicate(fldPath.Child("zones").Index(i).Child("name"), zone.Name))
+		}
+		usedZones.Insert(zone.Name)
 	}
 
 	return allErrs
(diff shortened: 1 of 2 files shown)


rfranzke (commented):
Same like https://github.com/gardener/gardener/pull/7221#discussion_r1050756415?

MartinWeindel:
> Same like https://github.com/gardener/gardener/pull/7221#discussion_r1050756415?

The check is only performed if the zone is added or updated. See line
https://github.com/gardener/gardener-extension-provider-aws/blob/master/pkg/apis/aws/validation/infrastructure.go#L60

That should be good enough.

dkistner (approved):
/lgtm

kon-angelo (approved):
/lgtm

Issue #577, opened 2022-07-20
Consider switching to the out-of-tree AWS cloud-controller-manager


ialidzhikov:
**How to categorize this issue?**

/area control-plane
/kind enhancement
/platform aws

**What would you like to be added**:
Since some time https://github.com/kubernetes/cloud-provider-aws has releases (that are not alpha or beta) and pushes container images.

We can check the compatibility with the in-tree CCM and, if possible, switch to the out-of-tree AWS CCM. We could do this by deploying the OOT CCM staring given K8s version (for example K8s 1.25).

**Why is this needed**:
To get rid of our in-tree CCM fork (https://github.com/gardener/cloud-provider-aws) and all the maintenance effort related to it.

dkistner:
Yes I fully agree. Finally it seems that the is also a published AWS ccm that we can consume. Adopting it with `v1.25` sounds reasonable for me (after successful validation of course).

PR #682, opened 2023-01-17
add upstream ccm image support

kon-angelo:

/area control-plane
/kind enhancement
/platform aws

Fixes https://github.com/gardener/gardener-extension-provider-aws/issues/577

**Release note**:

```other operator
Switch cloud-controller-manager image to upstream
```


MartinWeindel (approved):
/lgtm

pkg/controller/controlplane/valuesprovider.go
```diff
@@ -182,14 +183,6 @@ var (
 		Name: "shoot-system-components",
 		Path: filepath.Join(aws.InternalChartsPath, "shoot-system-components"),
 		SubCharts: []*chart.Chart{
-			{
```
dkistner, line 185: Can you elaborate why the rbac is not needed anymore? Even for k8s < 1.25?

kon-angelo: CCM is being deployed with cluster-admin https://github.com/gardener/gardener-extension-provider-aws/blob/master/charts/internal/shoot-system-components/charts/cloud-controller-manager/templates/rbac-cloud-controller-manager.yaml RBAC so it has all permissions anyway. 

We also don't deploy any separate node-controller so I don't see where these RBACs are used. Are you aware of any specific system usage that I didn't consider ?

kon-angelo: For a more detailed answer: the RBACs were needed in the past as the controllers use their own service account. Because of https://github.com/kubernetes/cloud-provider/blob/279df9502932e9784fe3a1ac562cbef38e35cbb2/app/controllermanager.go#L404-L406 they now use the node-controller SA which is already deployed as part of the KCM deployment. Only if we find out that the current node-controller RBACs are somehow insufficient we would need to add additional bindings. Even then the target of the binding would be different. So far, I didn't encounter a problem with this change. The same will be for other providers that deploy these RBACs. 

Thanks @rfranzke

charts/images.yaml
```diff
@@ -127,7 +127,7 @@ images:
       availability_requirement: 'low'
 - name: cloud-controller-manager
   sourceRepository: github.com/gardener/cloud-provider-aws
```
ialidzhikov, line 129:
```suggestion
  sourceRepository: github.com/kubernetes/cloud-provider-aws
```

MartinWeindel (approved):
/lgtm

PR #683, opened 2023-01-18
Add `cloudprovider` mutating webhook to ensure the `cloudprovider` secret has the `credentialsFile` field present.

AleksandarSavchev:

/area security
/area compliance
/kind enhancement
/platform aws

This PR adds a new mutating webhook which creates a new field `credentialsFile` for the `cloudprovider` secret, if the field does not already exist. The field is used for authentication purposes in `cloud-controler-manager` and `csi-driver-controller` deployments, replacing the old authentication method which used environment variables.
This PR is in sync with Kubernetes' STIG. See rule [242415](https://www.stigviewer.com/stig/kubernetes/2022-09-13/finding/V-242415) for more details.

Fixes #

cc @dimityrmirchev 

**Release note**:

```other operator
Added `cloudprovider` webhook part of `gardener-extension-provider-aws` which ensures that the `cloudprovider` secret has the `credentialsFile` field present. The format of this field can be found in [this documentation](https://docs.aws.amazon.com/sdkref/latest/guide/file-format.html).
```
```other operator
The `cloud-controler-manager` and `csi-driver-controller` deployments now use an AWS credentials file for authentication.
```

### DIFF
--- a/charts/internal/seed-controlplane/charts/csi-driver-controller/values.yaml
+++ b/charts/internal/seed-controlplane/charts/csi-driver-controller/values.yaml
@@ -101,5 +101,3 @@ csiSnapshotValidationWebhook:
   secrets:
     server: csi-snapshot-validation-server
   topologyAwareRoutingEnabled: false
-
-useCredentialsFile: false
--- a/charts/internal/seed-controlplane/charts/csi-driver-controller/values.yaml
+++ b/charts/internal/seed-controlplane/charts/csi-driver-controller/values.yaml
@@ -101,3 +101,5 @@ csiSnapshotValidationWebhook:
   secrets:
     server: csi-snapshot-validation-server
   topologyAwareRoutingEnabled: false
+
+useCredentialsFile: false
--- a/pkg/aws/types.go
+++ b/pkg/aws/types.go
@@ -65,6 +65,8 @@ const (
 	AccessKeyID = "accessKeyID"
 	// SecretAccessKey is a constant for the key in a cloud provider secret and backup secret that holds the AWS secret access key.
 	SecretAccessKey = "secretAccessKey"
+	// SharedCredentialsFile is a constant for the key in cloud provider secret that holds the AWS credentials file.
+	SharedCredentialsFile = "cloudprovider.credentials"
 	// Region is a constant for the key in a backup secret that holds the AWS region.
 	Region = "region"
 	// DNSAccessKeyID is a constant for the key in a DNS secret that holds the AWS access key id.
--- a/pkg/aws/types.go
+++ b/pkg/aws/types.go
@@ -66,7 +66,7 @@ const (
 	// SecretAccessKey is a constant for the key in a cloud provider secret and backup secret that holds the AWS secret access key.
 	SecretAccessKey = "secretAccessKey"
 	// SharedCredentialsFile is a constant for the key in cloud provider secret that holds the AWS credentials file.
-	SharedCredentialsFile = "cloudprovider.credentials"
+	SharedCredentialsFile = "credentialsFile"
 	// Region is a constant for the key in a backup secret that holds the AWS region.
 	Region = "region"
 	// DNSAccessKeyID is a constant for the key in a DNS secret that holds the AWS access key id.
--- a/pkg/controller/controlplane/valuesprovider.go
+++ b/pkg/controller/controlplane/valuesprovider.go
@@ -317,7 +317,7 @@ func (vp *valuesProvider) GetControlPlaneChartValues(
 	var useCredentialsFile bool
 	cloudprovider := &corev1.Secret{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      "cloudprovider",
+			Name:      v1beta1constants.SecretNameCloudProvider,
 			Namespace: cp.Namespace,
 		},
 	}
--- a/pkg/controller/controlplane/valuesprovider.go
+++ b/pkg/controller/controlplane/valuesprovider.go
@@ -37,14 +37,12 @@ import (
 	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
-	networkingv1 "k8s.io/api/networking/v1"
 	policyv1beta1 "k8s.io/api/policy/v1beta1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	storagev1 "k8s.io/api/storage/v1"
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	autoscalingv1 "k8s.io/autoscaler/vertical-pod-autoscaler/pkg/apis/autoscaling.k8s.io/v1"
-	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	apisaws "github.com/gardener/gardener-extension-provider-aws/pkg/apis/aws"
 	"github.com/gardener/gardener-extension-provider-aws/pkg/apis/aws/helper"
@@ -299,7 +297,7 @@ func (vp *valuesProvider) GetConfigChartValues(
 
 // GetControlPlaneChartValues returns the values for the control plane chart applied by the generic actuator.
 func (vp *valuesProvider) GetControlPlaneChartValues(
-	ctx context.Context,
+	_ context.Context,
 	cp *extensionsv1alpha1.ControlPlane,
 	cluster *extensionscontroller.Cluster,
 	secretsReader secretsmanager.Reader,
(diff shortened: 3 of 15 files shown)


pkg/aws/types.go
```diff
@@ -65,6 +65,8 @@ const (
 	AccessKeyID = "accessKeyID"
 	// SecretAccessKey is a constant for the key in a cloud provider secret and backup secret that holds the AWS secret access key.
 	SecretAccessKey = "secretAccessKey"
+	// SharedCredentialsFile is a constant for the key in cloud provider secret that holds the AWS credentials file.
+	SharedCredentialsFile = "cloudprovider.credentials"
```
rfranzke, line 69: The secret is already called `cloudprovider`. What is the rationale of duplicating this into this key? Just call it `credentials`?

AleksandarSavchev: Agreed, I have chosen a poor name. Maybe I can better name it `credentialsFile` since we already have credentials in the secret?

rfranzke: Sounds good :)

pkg/webhook/cloudprovider/ensurer.go
```diff
@@ -0,0 +1,80 @@
...
+
+// InjectScheme injects the given scheme into the decoder of the ensurer.
+func (e *ensurer) InjectScheme(_ *runtime.Scheme) error {
+	return nil
+}
+
+// EnsureCloudProviderSecret ensures that cloudprovider secret contain
+// the shared credentials file (if not present).
+func (e *ensurer) EnsureCloudProviderSecret(ctx context.Context, _ gcontext.GardenContext, new, _ *corev1.Secret) error {
+	if !hasSecretKey(new, aws.AccessKeyID) {
```
dimityrmirchev, line 57:
```suggestion
	if _, ok := new.Data[aws.AccessKeyID]; !ok {
```

pkg/controller/controlplane/valuesprovider.go
```diff
@@ -322,7 +323,21 @@ func (vp *valuesProvider) GetControlPlaneChartValues(
 		}
 	}
 
-	return getControlPlaneChartValues(cpConfig, cp, cluster, secretsReader, checksums, scaledDown)
+	var useCredentialsFile bool
+	cloudprovider := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "cloudprovider",
```
dimityrmirchev, line 329: We can reuse the name from [here](https://github.com/gardener/gardener/blob/a38354a6c3a0ae734d243553d14bed26c2fcbd0d/pkg/apis/core/v1beta1/constants/types_constants.go#L53)

dimityrmirchev (changes requested):
Thanks! I have left some initial suggestions and comments.

vendor/github.com/gardener/gardener/extensions/pkg/webhook/cloudprovider/cloudprovider.go
```diff
@@ -0,0 +1,86 @@
+// Copyright (c) 2021 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```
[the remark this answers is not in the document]
kon-angelo: This is a vendored file so I wouldn't suggest this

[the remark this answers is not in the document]
dimityrmirchev: Thanks, I did not see this at first.

pkg/webhook/cloudprovider/ensurer.go
```diff
@@ -0,0 +1,70 @@
...
+
+// EnsureCloudProviderSecret ensures that cloudprovider secret contains
+// the shared credentials file.
+func (e *ensurer) EnsureCloudProviderSecret(ctx context.Context, _ gcontext.GardenContext, new, _ *corev1.Secret) error {
+	if _, ok := new.Data[aws.AccessKeyID]; !ok {
+		return fmt.Errorf("could not mutate cloudprovider secret as %q field is missing", aws.AccessKeyID)
+	}
+	if _, ok := new.Data[aws.SecretAccessKey]; !ok {
+		return fmt.Errorf("could not mutate cloudprovider secret as %q field is missing", aws.SecretAccessKey)
+	}
```
kon-angelo, line 62: I am assuming that means you went with @dimityrmirchev first suggestion to not allow users to specify their own credentials file so that you always overwrite it right ?

AleksandarSavchev: Yes, users will not be able to specify credential files and the secret will remain the same. If you think it's better I can make it the other way and change the validating web hook as well.

dimityrmirchev: Now I am wondering if we always overwrite it do we really need the `useCredentialsFile` [boolean](https://github.com/gardener/gardener-extension-provider-aws/pull/683/files#diff-6a010d50f9d04c84f1da3b131822d1b10d39acac136bc54ab0b32b68a3698eaeR24)? I guess this will always be true

AleksandarSavchev: We would still need `useCredentialsFile` since our webhook would not mutate the `cloudprovider` secret unless it has been updated by the user.

AleksandarSavchev: After a joint review with @dimityrmirchev we discovered that in fact `useCredentialsFile` is not needed. During shoot reconciliation the `cloudprovider` secret is updated and the new `credentialsFile` field is always added.

pkg/webhook/cloudprovider/ensurer_test.go
```diff
@@ -0,0 +1,106 @@
...
+
+		err := ensurer.(inject.Scheme).InjectScheme(scheme)
+		Expect(err).NotTo(HaveOccurred())
+	})
+
+	Describe("#EnsureCloudProviderSecret", func() {
+		It("should fail as no accessKeyID is present", func() {
+			delete(secret.Data, aws.AccessKeyID)
+			err := ensurer.EnsureCloudProviderSecret(ctx, gctx, secret, nil)
+			Expect(err).To(HaveOccurred())
```
dimityrmirchev, line 73: Can we also check the error so we ensure that the returned err is the one we expect?

pkg/webhook/cloudprovider/ensurer_test.go
```diff
@@ -0,0 +1,106 @@
...
+		})
+		It("should fail as no secretAccessKey is present", func() {
+			delete(secret.Data, aws.SecretAccessKey)
+			err := ensurer.EnsureCloudProviderSecret(ctx, gctx, secret, nil)
+			Expect(err).To(HaveOccurred())
```
dimityrmirchev, line 78: Same here.

pkg/webhook/cloudprovider/ensurer_test.go
```diff
@@ -0,0 +1,106 @@
...
+		ensurer = NewEnsurer(logger)
+
+		err := ensurer.(inject.Scheme).InjectScheme(scheme)
+		Expect(err).NotTo(HaveOccurred())
+	})
+
+	Describe("#EnsureCloudProviderSecret", func() {
+		It("should fail as no accessKeyID is present", func() {
+			delete(secret.Data, aws.AccessKeyID)
+			err := ensurer.EnsureCloudProviderSecret(ctx, gctx, secret, nil)
```
dimityrmirchev, line 72: `gctx` is ignored in `EnsureCloudProviderSecret` so you can just pass `nil` I think.

pkg/webhook/cloudprovider/ensurer.go
```diff
@@ -0,0 +1,70 @@
...
+
+	e.logger.V(5).Info("mutate cloudprovider secret", "namespace", new.Namespace, "name", new.Name)
+	new.Data[aws.SharedCredentialsFile] = []byte("[default]\n" +
+		fmt.Sprintf("aws_access_key_id=%s\n", string(new.Data[aws.AccessKeyID])) +
+		fmt.Sprintf("aws_secret_access_key=%s", string(new.Data[aws.SecretAccessKey])))
```
dimityrmirchev, line 67: Nit: I think that it is nicer to put the closing bracket on a new line 
```suggestion
		fmt.Sprintf("aws_secret_access_key=%s", string(new.Data[aws.SecretAccessKey])),
		)
```

charts/internal/seed-controlplane/charts/csi-driver-controller/values.yaml
```diff
@@ -100,4 +100,3 @@ csiSnapshotValidationWebhook:
       memory: 200Mi
   secrets:
     server: csi-snapshot-validation-server
-
```
dimityrmirchev, line 103: Nit: Please add the newline at the end of the file

dimityrmirchev (changes requested):
Some final comments from my side :)

charts/internal/seed-controlplane/charts/csi-driver-controller/values.yaml
AleksandarSavchev: There were 2 new lines here. Should I revert them?

dimityrmirchev: I think that a single newline at the end of the file is enough

dimityrmirchev: > There were 2 new lines here. Should I revert them?

I see... Github tends to omit the newlines at the end and hence I got confused. You change is OK

dimityrmirchev (dismissed):
Thanks for addressing the review comments!
/lgtm

JordanJordanov:
@rfranzke @kon-angelo Do you have any more comments here or we can proceed and merge the PR?

kon-angelo (dismissed):
/lgtm

kon-angelo (approved):
/lgtm

PR #687, opened 2023-01-30
Allow to expand node CIDR

kon-angelo:

/area control-plane
/kind enhancement
/platform aws

This PR: 
- [x] adds an admission check to check that the nodes CIDR block can only be expanded when a gardener-managed VPC is used.
- [x] configvalidator check during runtime to check if the nodes CIDR is within the VPC range in "bring-your-own-VPC" scenarios.

Fixes #https://github.com/gardener/gardener/issues/4834

Generally for https://github.com/gardener/gardener/issues/4834 the extension is already having sufficient checks to ensure that the `vpc > nodes > worker subnets`. This PR only enhances this checks for runtime. In addition, the nodes CIDR is mutable in BYO-VPC cases to allow users to correct mistakes.

**Release note**:

```other operator

```

### DIFF
--- a/pkg/admission/validator/shoot.go
+++ b/pkg/admission/validator/shoot.go
@@ -186,6 +186,10 @@ func (s *shoot) validateShootUpdate(ctx context.Context, oldShoot, shoot *core.S
 		return err
 	}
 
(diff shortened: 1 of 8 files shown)


MartinWeindel (approved):
/lgtm

kon-angelo:
closing to fix issues that occured with the configvalidator

PR #692, opened 2023-01-31
purge remains of landscaper migration

ccwienk:
purge remains of landscaper-migration

not relevant for release-notes.


MartinWeindel (approved):
/lgtm

Issue #694, opened 2023-02-03
SSH to machines of type amd64 does not work


DockToFuture:
**How to categorize this issue?**

/area ops-productivity
/kind bug
/platform aws

**What happened**:
I wanted to ssh to a machine of architecture type `amd64` (c3.2xlarge) with gardenctl and currently it's not possible.
The CR shows the following error:

```
status:
  lastError:
    description: "Error reconciling Bastion: failed to ensure bastion instance: failed
      to run instance: InvalidParameterValue: The architecture 'x86_64,i386' of the
      specified instance type does not match the architecture 'arm64' of the specified
      AMI. Specify an instance type and an AMI that have matching architectures, and
      try again. You can use 'describe-instance-types' or 'describe-images' to discover
      the architecture of the instance type or AMI.\n\tstatus code: 400, request id:
      c1b2abb2-cf06-4024-8ac6-9cc980537b13"
    lastUpdateTime: "2023-02-03T10:21:30Z"
  lastOperation:
    description: "Error reconciling Bastion: failed to ensure bastion instance: failed
      to run instance: InvalidParameterValue: The architecture 'x86_64,i386' of the
      specified instance type does not match the architecture 'arm64' of the specified
      AMI. Specify an instance type and an AMI that have matching architectures, and
      try again. You can use 'describe-instance-types' or 'describe-images' to discover
      the architecture of the instance type or AMI.\n\tstatus code: 400, request id:
      c1b2abb2-cf06-4024-8ac6-9cc980537b13"
```

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:

- Gardener version (if relevant):
- Extension version:
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
- Others:

tedteng:
I like to take a look 
/assign

tedteng:
identify the issue due to the latest AMI Image `gardenlinux-aws-gardener_prod-arm64` base on Architecture "arm64" not matching with the instance type t2.nano. will look for a suitable instance type to match the Architecture "arm64"

PR #700, opened 2023-02-09
fix arm64 type for bastion instance

tedteng:

/area ops-productivity
/kind task
/platform aws

As AMI Image changed to arm64 processor Architecture `gardenlinux-aws-gardener_prod-arm64` , the original `t2.nano` with x86_64 is not supported. add a new instance type `t4g.nano` with arm64 architecture

Fixes https://github.com/gardener/gardener-extension-provider-aws/issues/694

**Release note**:

```other operator
add an extra t4g.nano(arm64) processor architecture bastion instance to support for linux/arm64
```

### DIFF
--- a/pkg/controller/bastion/options.go
+++ b/pkg/controller/bastion/options.go
@@ -86,7 +86,7 @@ func DetermineOptions(ctx context.Context, bastion *extensionsv1alpha1.Bastion, 
 		return nil, fmt.Errorf("failed to determine OS image for bastion host: %w", err)
 	}
 
-	instanceType, err := determineInstanceType(ctx, awsClient)
+	instanceType, err := determineInstanceType(ctx, imageID, awsClient)
 	if err != nil {
 		return nil, fmt.Errorf("failed to determine instance type: %w", err)
 	}
@@ -163,7 +163,15 @@ func determineImageID(shoot *gardencorev1beta1.Shoot, providerConfig *awsv1alpha
 	return "", fmt.Errorf("found no suitable AMI for machines in region %q", shoot.Spec.Region)
 }
 
-func determineInstanceType(ctx context.Context, awsClient *awsclient.Client) (string, error) {
+func determineInstanceType(ctx context.Context, imageID string, awsClient *awsclient.Client) (string, error) {
+	var preferredType string
+	imageInfo, err := getImagesInfo(ctx, imageID, awsClient)
+	if err != nil {
+		return "", err
+	}
+
+	imageArchitecture := imageInfo.Architecture
+
 	offerings, err := awsClient.EC2.DescribeInstanceTypeOfferingsWithContext(ctx, &ec2.DescribeInstanceTypeOfferingsInput{})
 	if err != nil {
 		return "", fmt.Errorf("failed to list instance types: %w", err)
@@ -174,9 +182,15 @@ func determineInstanceType(ctx context.Context, awsClient *awsclient.Client) (st
 		types[i] = *offering.InstanceType
 	}
 
-	// prefer t2.nano
+	switch *imageArchitecture {
+	case "amd64":
+		preferredType = "t2.nano"
+	case "arm64":
+		preferredType = "t4g.nano"
+	}
+
 	for _, t := range types {
-		if t == "t2.nano" {
+		if t == preferredType {
 			return t, nil
 		}
 	}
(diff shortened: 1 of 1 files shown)


kon-angelo (approved):
/lgtm

petersutter:
Hi Ted,
there are some things that can be improved in my opinion, however I'm not firm with the aws api so take it with a grant of salt :)
- consider ALL images of the ami, not just the first
- use the supported-architecture filter when searching for instance types
- pass the t* types list to the `DescribeInstanceTypesWithContext` function, instead of calling it for every single one in a loop
- it uses `x86_64` instead of `amd64` as architecture. This was a bug in your code in my opinion. Fixing this we can also get rid of the unnecessary `convertName` mapping function
- also you had no fallback-preferredType (no `default` in your switch) in case the architecture is unknown

Here is my suggestion (not tested!). Please note that it makes use of the new generic implementation of `k8s.io/apimachinery/pkg/util/sets` `sets.New`. For this you have to revendor `k8s.io/apimachinery` >= v0.26
```go
func getImages(ctx context.Context, ami string, awsClient *awsclient.Client) ([]*ec2.Image, error) {
	imageInfo, err := awsClient.EC2.DescribeImagesWithContext(ctx, &ec2.DescribeImagesInput{
		ImageIds: []*string{
			aws.String(ami),
		},
	})
	if err != nil {
		return nil, fmt.Errorf("failed to get Images Info: %w", err)
	}

	if len(imageInfo.Images) == 0 {
		return nil, fmt.Errorf("images info not found: %w", err)
	}

	return imageInfo.Images, nil
}

func determineInstanceType(ctx context.Context, imageID string, awsClient *awsclient.Client) (string, error) {
	images, err := getImages(ctx, imageID, awsClient)
	if err != nil {
		return "", err
	}

	imageArchitectures := sets.New[*string]()
	for _, image := range images {
		if image.Architecture == nil {
			continue
		}
		imageArchitectures.Insert(image.Architecture)
	}

	offerings, err := awsClient.EC2.DescribeInstanceTypeOfferingsWithContext(ctx, &ec2.DescribeInstanceTypeOfferingsInput{})
	if err != nil {
		return "", fmt.Errorf("failed to list instance types: %w", err)
	}

	tTypes := make([]*string, len(offerings.InstanceTypeOfferings))
	for i, offering := range offerings.InstanceTypeOfferings {
		// we are only interested in instance types starting with `t`
		if offering.InstanceType == nil || !strings.HasPrefix(*offering.InstanceType, "t") {
			continue
		}
		tTypes[i] = offering.InstanceType
	}

	preferredTypes := sets.New[string]()
	for imageArchitecture, _ := range imageArchitectures {
		switch *imageArchitecture {
		case "arm64":
			preferredTypes.Insert("t4g.nano")
		case "x86_64":
		default:
			preferredTypes.Insert("t2.nano")
		}
	}

	for _, t := range tTypes {
		if _, ok := preferredTypes[*t]; ok {
			return *t, nil
		}
	}

	instanceTypeOutput, err := awsClient.EC2.DescribeInstanceTypesWithContext(ctx, &ec2.DescribeInstanceTypesInput{
		Filters: []*ec2.Filter{
			{
				Name:   aws.String("processor-info.supported-architecture"),
				Values: imageArchitectures.UnsortedList(),
			},
		},
		InstanceTypes: tTypes,
	})
	if err != nil {
		return "", fmt.Errorf("failed to describe instance types %v: %w", tTypes, err)
	}

	if len(instanceTypeOutput.InstanceTypes) == 0 {
		return "", fmt.Errorf("no t.* instance type available")
	}

	// fallback to the first (hopefully smallest) other general purpose instance type
	return *instanceTypeOutput.InstanceTypes[0].InstanceType, nil
}
```

kon-angelo:
@tedteng I reverted the PR because the bastion tests fail. Please open a new PR with the bastion test fixes

tedteng:
Thanks Peter ,  just some thoughts 
```
func getImages(ctx context.Context, ami string, awsClient *awsclient.Client) ([]*ec2.Image, error) {
```
an node ami(eg ami-1234) passed in `getImages` which get from cloudProfileConfig. then return full image info. I noticed from aws side the`imageInfo.Images` field is slices type ([]*ec2.Image). I guess that the reason this method return `[]*ec2.Image` . I don't know why is slices, I guess it may use for ami alias mapping to one ami.  then I think maybe just return `*ec2.Image`, in this way we don't need to introduce `sets` ,`imageArchitectures := sets.New[*string]()` and the relevant code . 

image

petersutter:
Hi Ted, I agree.

As we pass the image ID and only one to `DescribeImagesWithContext` we should only expect up to one image to be returned as the images ids are unique. 
-> We do not need to consider multiple images in the response and can revert the behaviour of `getImages` to your implementation
I ~~crossed~~ this point out in my list above. Thanks

tedteng:
New PR is ready in https://github.com/gardener/gardener-extension-provider-aws/pull/707 and some logic changed per feedback

PR #705, opened 2023-02-16
Revert "fix arm64 type for bastion instance"

kon-angelo:
Reverts gardener/gardener-extension-provider-aws#700

Because bastion tests failed


MartinWeindel (approved):
/lgtm

Issue #468, opened 2021-12-29
allow customizing instance metadata options


tareksha:
/area control-plane
/kind enhancement
/platform aws

**What would you like to be added**:

I would like to have the ability to customize ec2 instance metadata options at worker level. Most importantly I want to modify the option `HttpPutResponseHopLimit` which is very useful for using the EC2 IAM Roles feature from containerized applications on top of ec2 instances:
https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/iam-roles-for-amazon-ec2.html
The security-credentials endpoint limits the number of hops at IP level to 1 by default. This makes it inaccessible from inside containers because their network traffic goes through an additional hop. To make the endpoint accessible from inside containerized applications the metadata option `HttpPutResponseHopLimit` must be changed to 2 (or higher).

I suggest adding a new section to `WorkerConfig`:
```yaml
apiVersion: aws.provider.extensions.gardener.cloud/v1alpha1
kind: WorkerConfig
...
instanceMetadataOptions:
  httpPutResponseHopLimit: 2
```

Full list of options:
https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_ModifyInstanceMetadataOptions.html

**Possible alternative**

If we do not want a full-fledged feature for customizing the options maybe we can change the hop limit to 2 by default. The instances are created for the purpose of running containerized applications anyway. EKS did this:
https://aws.amazon.com/about-aws/whats-new/2020/08/amazon-eks-supports-ec2-instance-metadata-service-v2/
> Now, newly launched and any updated EKS managed node groups will be configured with a metadata token response hop limit set to 2.

**Why is this needed**:

Allow using IAM Roles for EC2 instances in gardener clusters conveniently.

etiennnr:
Related old PR : https://github.com/gardener/machine-controller-manager-provider-aws/pull/61

shenningz:
[...]

PR #709, opened 2023-02-22
add instance metadatav2 support

kon-angelo:

/area control-plane
/kind enhancement
/platform aws

Adds support for IMDSv2. 

Fixes #https://github.com/gardener/gardener-extension-provider-aws/issues/468
Depends on https://github.com/gardener/machine-controller-manager-provider-aws/pull/105

+ Completely disabling access to the metadata service is currently not supported because nodes cannot be created without it. I am not aware of where the problem lies.
+ The IMDS both v1 and v2 are available on the host node by default.
+ By enabling both `httpPutResponseHopLimit >=2` and `httpTokens = "required"` we can enforce usage of IMDSv2.
+ By enabling only `httpTokens = "required"` we basically disallow access to both services from inside containers. However IMDSv2 and hence access to the metadata server will still be possible from the Host Network.

**Release note**:

```other operator
Add support for IMDSv2
```


MartinWeindel (approved):
/lgtm

himanshu-kun:
points discussed in internal grooming:
- default should be `httpTokens: required , hopLimit: 2 , httpEndpoint: enabled`  (this mandatorily enables IMDSv2 for new instances)
- the default can be overriden in case they want more time with IMDSv1

@kon-angelo we will need a meeting with you over this, and would provide more inputs after reviewing the PR.

kon-angelo:
@himanshu-kun 

> default should be httpTokens: required , hopLimit: 2 , httpEndpoint: enabled (this mandatorily enables IMDSv2 for new instances)

I disagree with this. this will break any and all users that are currently relying on IMDSv1. Why go forward with a breaking change ? You could argue at most that the IMDSv2 should be enabled by default  (`hopLimit: 2 , httpEndpoint: enabled`) but I would not go forward and disable v1 yet.

dkistner:
Actually I agree with @kon-angelo. We should preserve the behaviour for a while. 
In my opinion it is good strategy to allow now optional opt in for v2 per worker pool and later (after a proper announcement and time for adoption by users) to switch the default from v1 to v2.

pkg/controller/worker/machines.go
```diff
@@ -163,7 +165,8 @@ func (w *workerDelegate) generateMachineConfig() error {
 				"secret": map[string]interface{}{
 					"cloudConfig": string(pool.UserData),
 				},
-				"blockDevices": blockDevices,
+				"blockDevices":            blockDevices,
+				"instanceMetadataOptions": instanceMetadataOptions,
```
dkistner, line 169: Nit: we can call `computeInstanceMetadata(workerConfig)` directly here.

pkg/apis/aws/validation/worker.go
```diff
@@ -120,3 +122,29 @@ func validateVolumeConfig(volume *apisaws.Volume, volumeType string, fldPath *fi
...
+	if md == nil {
+		return allErrs
+	}
+
+	if !md.EnableInstanceMetadataV2 && md.HTTPPutResponseHopLimit != nil {
+		allErrs = append(allErrs, field.Invalid(fldPath.Child("HTTPPutResponseHopLimit"), *md.HTTPPutResponseHopLimit, "enableInstanceMetadataV2 must be set to specify this field"))
+		return allErrs
+	}
+
+	if md.EnableInstanceMetadataV2 && md.HTTPPutResponseHopLimit == nil {
```
dkistner, line 137: Can this case ever happen as `HTTPPutResponseHopLimit` is defaulted to 2?

pkg/controller/worker/machines.go
kon-angelo: The only reason I didn't do it like this is because this line (`169`) is part of the per-zone loop. The metadata are common per worker group and not needed to be called per zone. 

It is a minuscule optimization that saves a function call but if you think that it improves readability we can follow your suggestion.

pkg/apis/aws/validation/worker.go
kon-angelo: Correct it cannot happen as it is (due to the defaulter). This is a bit of "defensive programming".

pkg/controller/worker/machines.go
dkistner: Good point. No calculate the options for metadata service once per workerpool is sufficient.

pkg/apis/aws/v1alpha1/defaults.go
```diff
@@ -55,3 +55,10 @@ func SetDefaults_MachineImage(obj *MachineImage) {
 		obj.Architecture = pointer.String(v1beta1constants.ArchitectureAMD64)
 	}
 }
+
+// SetDefaults_InstanceMetadata sets the default response hop limit for the IMDS.
+func SetDefaults_InstanceMetadata(obj *InstanceMetadata) {
+	if obj.EnableInstanceMetadataV2 && obj.HTTPPutResponseHopLimit == nil {
+		obj.HTTPPutResponseHopLimit = pointer.Int64(2)
+	}
```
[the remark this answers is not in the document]
himanshu-kun: After the internal call, we have decided to only expose `httpTokens` and `httpPutResponseHopLimit` to user

himanshu-kun (changes requested):
Thanks for the PR!
I have updated the review comments after the internal discussion.

pkg/apis/aws/v1alpha1/defaults.go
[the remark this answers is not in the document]
himanshu-kun: Keeping @unmarshall's comment in mind https://github.com/gardener/machine-controller-manager-provider-aws/pull/105#discussion_r1122564757
maybe we should enhance the `mcm-provider-aws` `InstanceMetadataOptions` struct so that we could re-use it here? your call though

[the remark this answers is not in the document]
kon-angelo: I don't think we should reuse it. For starters it is currently not possible to create a machine with IMDS completely disabled. At most you can disable it from inside containers/pods (which is handled by `httpTokens`). Therefore reusing the struct and would also expose `httpEndpoint` which would cause more headaches. 

It would also make updates to mcm-aws harder if we start coupling the structs (because it would require also updates to provider-aws) etc. Let each component define its interface even if for convenience here it's almost the same.

charts/internal/machineclass/values.yaml
```diff
@@ -39,3 +39,7 @@ machineClasses:
     # iops: 100
     # throughput: 125 #(only for gp3)
     # snapshotID: snap-12345
+#  instanceMetadata:
+#    httpEndpoint: "disabled"
+#    httpTokens: "required"
+#    httpPutResponseHopLimit: 2
```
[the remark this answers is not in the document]
kon-angelo: These will remain commented and are only there for example's grace but I agree we can populate them with the defaults :)

pkg/apis/aws/types_worker.go
```diff
@@ -118,3 +120,14 @@ const (
...
+
+// InstanceMetadata contains configuration for controlling access to the metadata API.
+type InstanceMetadata struct {
+	// EnableInstanceMetadataV2 when set will allow access to IMDSv2 API from containerized applications.
+	EnableInstanceMetadataV2 bool
+	// DisableInstanceMetadataV1 will disable access to IMDSv1.
+	DisableInstanceMetadataV1 bool
+	// HTTPPutResponseHopLimit is the response hop limit for instance metadata requests.
+	// Valid values are between 1 and 64. Defaults to 2 if not specified by user.
+	HTTPPutResponseHopLimit *int64
```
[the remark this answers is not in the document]
kon-angelo: The initial idea was to keep the provider-aws focused on customer scenarios. Say this issue is a good example: https://github.com/gardener/machine-controller-manager-provider-aws/issues/99 asking to enable v1 only. If you see the screenshot provided in the comment AWS also doesn't not expose the internals directly but only the logical operations.

In general the scenarios supported are the following:
- enable v2
- disable v1
- disable both for containers (which is something on k8s level only and not reflected in AWS documentation)

There are some benefits to decoupling the APIs (logical operation vs implementation via "hops) but we can go with the simplest route which mirroring the internal opts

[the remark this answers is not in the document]
unmarshall: My 2 cents:

- WorkerConfig is AWS specific struct. Therefore using provider specific configuration names makes it easy to correlate.
- If AWS tomorrow decides to remove support for v1 completely then this struct has to be changed again.

pkg/controller/worker/machines.go
```diff
@@ -376,3 +379,20 @@ func computeIAMInstanceProfile(workerConfig *awsapi.WorkerConfig, infrastructure
 
 	return nil, fmt.Errorf("unable to compute IAM instance profile configuration")
 }
+
+func computeInstanceMetadata(workerConfig *awsapi.WorkerConfig) map[string]interface{} {
+	res := make(map[string]interface{})
```
himanshu-kun, line 384: we should default `httpPutResponseHopLimit` to 2 here.

himanshu-kun (changes requested):
could you also update the `special notes for reviewer`

docs/usage-as-end-user.md
```diff
@@ -295,6 +295,9 @@ dataVolumes:
   snapshotID: snap-1234
 iamInstanceProfile: # (specify either ARN or name)
   name: my-profile
+instanceMetadataOptions:
+  httpTokens: required
```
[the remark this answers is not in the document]
kon-angelo: I think we should use an example of how people would like to use it which is the required value. The "optional" is the equivalent of a nil value anyway.

pkg/controller/worker/machines.go
kon-angelo: If we default it to 2, then we change the default behavior to enable IMDSv2  by default and this is not what we should do for now. We can in the future consider this with proper "breaking" announcements, but let's not change the defaults in this iteration.

himanshu-kun: okay , would adding a `TODO` here make sense ? because in the docString of `HTTPResponseHopLimit` its written that the hop limit is 2, so someone reading the code might get confused.

kon-angelo: This was a problem with the documentation I that I forgot to remove. I removed the mention of the defaulting - thank you for this catch.

The defaulting only makes sense with the old set of flags that I had exposed.

himanshu-kun (approved):
looks great ! Thanks for the effort
/lgtm

pkg/apis/aws/types_worker.go
```diff
@@ -117,3 +119,24 @@ const (
...
+	// HTTPTokensRequired is a constant for requiring the use of tokens to access IMDS. Effectively disables access via
+	// the IMDSv1 endpoints.
+	HTTPTokensRequired HTTPTokensValue = "required"
+	// HTTPTokensOptional that makes the use of tokens for IMDS optional. Effectively allows access via both IMDSv1 and
+	// IMDSv2 endpoints.
+	HTTPTokensOptional HTTPTokensValue = "optional"
+)
+
+// InstanceMetadataOptions contains configuration for controlling access to the metadata API.
+type InstanceMetadataOptions struct {
```
unmarshall, line 136: To completely disable IMDS access AWS provides `HttpEndpoint=disabled`. Would it make sense to have that explicitly instead of depending upon the combination of `HttpTokens = required` and `HttpPutResponseHopLimit = 1 or not set`? From the consumer's perspective it becomes simpler to disable via one field.

pkg/apis/aws/types_worker.go
```diff
@@ -117,3 +119,24 @@ const (
...
+	// HTTPTokens enforces the use of metadata v2 API.
+	HTTPTokens *HTTPTokensValue
+	// HTTPPutResponseHopLimit is the response hop limit for instance metadata requests.
+	// Valid values are between 1 and 64.
+	HTTPPutResponseHopLimit *int64
```
unmarshall, line 141: If you have a separate field for disabling IMDS as commented above then you could not make it a pointer and default it to 1. Even aws does the same (https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-launchtemplate-launchtemplatedata-metadataoptions.html)

pkg/apis/aws/types_worker.go
unmarshall, line 136: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-launchtemplate-launchtemplatedata-metadataoptions.html also provides another option to enable/disable access to instance tags via `InstanceMetadataTags`. By default it is disabled. Should this also be configurable?

kon-angelo: I wrote about this in the special notes:

> Completely disabling access to the metadata service is currently not supported because nodes cannot be created without it. I am not aware of where the problem lies.

If I pass `HttpEndpoint=disabled` to MCM, then the machine cannot come up. So you need to have access to some form of IMDS on the host at the moment. The combination of `httpTokens` and `HttpPutResponseHopLimit` allows you to have access from the host (and therefore the machine can come up), but not from inside containers.

kon-angelo: This wouldn't work at the moment because the MCM side would need to be adapted as well.

pkg/apis/aws/types_worker.go
kon-angelo: The idea behind exposing pointers is that when they are nil you allow AWS to default them. Case in point, if AWS would change their defaults to `2` then we wouldn't need to follow up with a mirroring change on our side, we simply don't assume what the default value is and don't pass this option if we don't have to. 

Our validation also would prevent you from using any value that doesn't fall in the `1-64` range anyway.

pkg/apis/aws/types_worker.go
kon-angelo: But as with many of these requests, we add features as requested. The main focus has been controlling the access to IMDS for this one. 

We can go back and add the flags to MCM and then the provider extension in a separate PR when needed (e.g. like the ipv6 flag), unless you think that they have to be included in this PR.

unmarshall: We can always enhance the struct with additional fields at a later point in time.

pkg/apis/aws/types_worker.go
unmarshall: > Case in point, if AWS would change their defaults to 2 then we wouldn't need to follow up with a mirroring change on our side, we simply don't assume what the default value is and don't pass this option if we don't have to.

Ok would it make sense to add this as part of the documentation that if you choose not to specify any value then provider defaults will be used (to make it clear that gardener does not default them to anything). This means that the consumer will have to be aware of the provider defaults which should ideally not be a problem.

pkg/apis/aws/types_worker.go
unmarshall: > If I pass HttpEndpoint=disabled to MCM, then the machine cannot come up. So you need to have access to some form of IMDS on the host at the moment. 

@himanshu-kun is that the case? I do not see any code in the provider during creation of the VM which mandates this. We can try and check if this is true.

kon-angelo: @unmarshall it would be nice if you independently validate that. My impression was that it should work without IMDS but somehow in practice it didn't. I would be happy to add the `httpEndpoint` once this point is clarified.

unmarshall (approved):
/lgtm

MartinWeindel (approved):
/lgtm

Issue #713, opened 2023-03-03
Find a more "sustainable" way to get the image AMIs for the test


tedteng:
**How to categorize this issue?**

/area ops-productivity
/kind enhancement
/platform aws

**What would you like to be added**:
To avoid image AMIs expired or out of stack which blocks the test, find a  "sustainable" way for test image AMIs 
**Why is this needed**:
per comments, https://github.com/gardener/gardener-extension-provider-aws/pull/712#issuecomment-1452528784, find a more "sustainable" way to get the image AMIs for the test to avoid image AMI expired or out of stock

PR #715, opened 2023-03-06
add getImageAMI for sustainable test

tedteng:

/area ops-productivity
/kind enhancement
/platform aws

add a "sustainable" way to get available image AMIs for the test 

Fixes https://github.com/gardener/gardener-extension-provider-aws/issues/713

**Release note**:

```other operator
add a sustainable way to get available image AMIs for the test
```
/invite @kon-angelo

### DIFF
--- a/test/integration/bastion/bastion_test.go
+++ b/test/integration/bastion/bastion_test.go
@@ -182,7 +182,7 @@ var _ = BeforeSuite(func() {
 	awsClient, err = awsclient.NewClient(*accessKeyID, *secretAccessKey, *region)
 	Expect(err).NotTo(HaveOccurred())
 
-	imageAMI := getImageIAM(ctx, bastionAMI, awsClient)
+	imageAMI := getImageAMI(ctx, bastionAMI, awsClient)
 	amiID := determineBastionImage(ctx, imageAMI, awsClient)
 	extensionscluster, corecluster = newCluster(namespaceName, amiID)
 })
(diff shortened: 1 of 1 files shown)


kon-angelo:
I am not sure about the particulars of the image, but should we maybe try to use the `server` type images ? I think this is not guaranteed as it. Maybe with an small adaptation to the glob pattern ?

tedteng:
done change to `ubuntu-*server*` type image
/test

tedteng:
@kon-angelo I did some testing after changing to `ubuntu-*server*` type image, however, it bring one issue let me whether still need more effort to resolve this by using `server` type image, or just keep the current solution. 
`error reconciling Bastion: failed to ensure bastion instance: failed to run instance: OptInRequired: In order to use this AWS Marketplace product you need to accept terms and subscribe. To do so please visit https://aws.amazon.com/marketplace/pp?sku=f165x13dfhzd2xbgnkvs6s5tr\n\tstatus code: 401, `

tedteng:
just check `ubuntu-jammy*`, the full image name `ubuntu/images/hvm-ssd/ubuntu-jammy-22.04-amd64-server-20230208`, is this the `server` type image you mentioned?

tedteng:
@kon-angelo Can this PR be processed?

tedteng:
@kon-angelo Can we process this PR? Thanks

PR #726, opened 2023-03-24
fix issue with dup zone validation

kon-angelo:

/area control-plane
/kind bug
/platform aws

Fixes #

**Release note**:

```other operator
Fix an issue that allows duplicate zones in the infrastructure.
```


MartinWeindel (approved):
/lgtm

ialidzhikov:
Isn't this a breaking change? What happens to Shoots that already have duplicate zones in their infrastructureConfig? Do we have such Shoots on our landscapes?

kon-angelo:
@ialidzhikov  It shouldn't be a breaking change. For existing zones (pre-update), it just adds them to the set, but it should disallow you from adding even more duplicate zones. Before this change you could theoretically add already duplicated zones ad-infinitum.

ialidzhikov:
Ah, I see. It is only for new zones that are being added. You are right. Thanks for explaining!

PR #738, opened 2023-04-19
Block public access to S3 bucket as per standard practices

shreyas-s-rao:

/area backup
/area security
/area compliance
/kind task
/platform aws

This PR configures the newly created S3 buckets to be blocked from public access as per standard practices. This is achieved by putting a public-access-block on the created bucket, as per [AWS documentation](https://docs.aws.amazon.com/cli/latest/reference/s3api/put-public-access-block.html).

Fixes #

cc @vlerenc 

**Release note**:

```other operator
Block public access for newly created S3 buckets.
```

### DIFF
--- a/pkg/aws/client/client.go
+++ b/pkg/aws/client/client.go
@@ -377,6 +377,42 @@ func (c *Client) CreateBucketIfNotExists(ctx context.Context, bucket, region str
 		return err
 	}
 
+	// Set bucket policy to deny non-HTTPS requests
+	bucketPolicy := map[string]interface{}{
+		"Version": "2012-10-17",
+		"Statement": []map[string]interface{}{
+			{
+				"Effect":    "Deny",
+				"Principal": "*",
+				"Action":    "s3:*",
+				"Resource": []string{
+					fmt.Sprintf("arn:aws:s3:::%s", bucket),
+					fmt.Sprintf("arn:aws:s3:::%s/*", bucket),
+				},
+				"Condition": map[string]interface{}{
+					"Bool": map[string]string{
+						"aws:SecureTransport": "false",
+					},
+					"NumericLessThan": map[string]string{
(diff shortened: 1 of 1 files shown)


pkg/aws/client/client.go
```diff
@@ -364,6 +364,19 @@ func (c *Client) CreateBucketIfNotExists(ctx context.Context, bucket, region str
...
+		PublicAccessBlockConfiguration: &s3.PublicAccessBlockConfiguration{
+			BlockPublicAcls:       aws.Bool(true),
+			BlockPublicPolicy:     aws.Bool(true),
+			IgnorePublicAcls:      aws.Bool(true),
+			RestrictPublicBuckets: aws.Bool(true),
+		},
+	}); err != nil {
+		return err
+	}
+
```
abdasgupta, line 379: I think we should add bucket policy here. I received some alert for not having proper bucket policies. Example bucket policy:
```
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Sid": "RestrictToTLSRequestsOnly",
            "Effect": "Deny",
            "Principal": "*",
            "Action": "s3:*",
            "Resource": "arn:aws:s3:::9eeccd1a-d2d1-499b-977c-0d346f9d3b79/*",
            "Condition": {
                "Bool": {
                    "aws:SecureTransport": "false"
                },
                "NumericLessThan": {
                    "s3:TlsVersion": "1.2"
                }
            }
        }
    ]
}
```
More info: https://docs.aws.amazon.com/AmazonS3/latest/userguide/access-policy-language-overview.html

abdasgupta (changes requested):
I think we need to add bucket policy here

shreyas-s-rao:
@abdasgupta thanks, I have addressed your comment now. PTAL

kon-angelo (approved):
/lgtm

PR #766, opened 2023-06-13
Fix S3 bucket policy for different AWS partitions (region groups)

shreyas-s-rao:

/area backup
/kind bug
/platform aws

Adds handling for S3 bucket policy for different [AWS partitions](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference-arns.html) (region groups) such as [China regions](https://aws.amazon.com/blogs/enterprise-strategy/getting-started-with-aws-services-in-aws-china-beijing-region-and-aws-china-ningxia-region/) and [AWS GovCloud (US) regions](https://docs.aws.amazon.com/govcloud-us/latest/UserGuide/using-govcloud-arns.html).

Fixes #

**Release note**:

```bugfix operator
Handle S3 bucket policy IAM ARN for China and GovCloud (US) regions.
```

### DIFF
--- a/pkg/aws/client/client.go
+++ b/pkg/aws/client/client.go
@@ -377,6 +377,14 @@ func (c *Client) CreateBucketIfNotExists(ctx context.Context, bucket, region str
 		return err
 	}
 
+	// Handle bucket policy IAM ARN for different partitions (AWS region groups)
+	arnPartition := "aws"
+	if strings.HasPrefix(region, "cn-") {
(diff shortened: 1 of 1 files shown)


kon-angelo (dismissed):
Maybe a comment with the link like https://docs.aws.amazon.com/IAM/latest/UserGuide/reference-arns.html would be good to know what are the available partitions. Otherwise

/lgtm

PR #771, opened 2023-06-20
Drop support for Shoots with Kubernetes version < 1.22

shafeeqes:

/area open-source
/kind cleanup
/platform aws

Drop support for kubernetes < 1.22 for AWS extension provider

Part of https://github.com/gardener/gardener/issues/7405

We need to vendor gardener@1.74 once it's released
/hold
/squash

**Release note**:

```breaking operator
`provider-aws` no longer supports Shoots or Seeds with Кubernetes version < 1.22.
```


pkg/controller/controlplane/valuesprovider_test.go
```diff
@@ -80,7 +80,7 @@ var _ = Describe("ValuesProvider", func() {
 				Raw: encode(&apisawsv1alpha1.ControlPlaneConfig{
 					CloudControllerManager: &apisawsv1alpha1.CloudControllerManagerConfig{
 						FeatureGates: map[string]bool{
-							"CustomResourceValidation": true,
+							"RotateKubeletServerCertificate": true,
```
kon-angelo, line 83: I am not sure, is this a valid CCM flag ? Because I see this was changed in many places.

pkg/controller/worker/machines.go
```diff
@@ -166,6 +165,7 @@ func (w *workerDelegate) generateMachineConfig() error {
 				},
 				"blockDevices":            blockDevices,
 				"instanceMetadataOptions": instanceMetadataOptions,
+				"srcAndDstChecksEnabled":  false,
```
kon-angelo, line 168: Should we maybe just embed that directly in the machineclass chart instead since it doesn't make sense for it to be configurable anyway ?

kon-angelo (commented):
One thing to note is that all the chart `PDB` changes remove the `policy/v1beta` in favor of `v1` for things that get deployed in the seed. The result is that this version of the extension can no longer be used with seeds `< v1.21` which is kind of a "breaking" change.

pkg/controller/controlplane/valuesprovider_test.go
shafeeqes: I just saw this in `charts/internal/seed-controlplane/charts/cloud-controller-manager/values.yaml` together with  `CustomResourceValidation `, so I just replaced it in examples and tests.

shafeeqes:
> One thing to note is that all the chart `PDB` changes remove the `policy/v1beta` in favor of `v1` for things that get deployed in the seed. The result is that this version of the extension can no longer be used with seeds `< v1.21` which is kind of a "breaking" change.

Yes. I will add a release note.

kon-angelo:
@shafeeqes can you check the `make verify` failing ?

kon-angelo (approved):
/lgtm

PR #778, opened 2023-06-28
Add/dualstack

DockToFuture:
Co-authored-by: axel7born <axel.siebenborn@sap.com>

/area networking
/kind enhancement
/platform aws

This PR adds the `spec.provider.infrastructureConfig.enableDualstack` field which specifies whether dualstack or IPv4 only should be supported by the infrastructure. When dualstack is enabled the VPC gets an /56 IPv6 block from AWS assigned and all subnets get a  /64 IPv6 CIDR block from that /56 CIDR block assigned. A route which routes all IPv6 traffic over the IGW is added to the main route table. 

This field can be set in the shoot spec and has false as default.

```  
spec:
  provider:
    infrastructureConfig:
      enableDualstack: true
```

Fixes #

**Release note**:

```other operator
Infrastructure dualstack support can be enabled via `spec.provider.infrastructureConfig.enableDualstack: true` in the shoot.yaml.
```


timebertt:
@DockToFuture @axel7born @ScheererJ 
Interesting to see that you are also working on the IPv6 topic.

I'm curious to hear what kind of use case you plan on covering with this PR. I expect that it only configures the nodes for dual-stack connectivity, while pods and services stay with IPv4 single-stack connectivity.
I'm wondering whether we should coordinate this work stream with [GEP-21](https://github.com/gardener/gardener/issues/7051), organizationally and technically.
For example, GEP-21 adds the `ipFamilies` field to many relevant gardener resources. I think, aligning you work with this and introducing a new `ipFamilies` field in the `Infrastructure` resource makes more sense than adding new fields in all the `providerConfig` APIs.

I'm happy to have a chat via Slack or in person to align our efforts.

cc @einfachnuralex

ScheererJ:
@timebertt This change is even less invasive than you anticipated. Together with https://github.com/gardener/gardener-extension-provider-aws/pull/717 and a few annotations, it allows us to expose a kubernetes service via a dual-stack loadbalancer without having IPv6 within kubernetes, i.e. users of the loadbalancer can query the domain name for a AAAA record, receive an IPv6 address and connect to the service via IPv6 while inside the cluster everything stays on single-stack IPv4. This allows to easily expose `kube-apiserver` via both IPv4 and IPv6 so that we can continue to work on the shoot side.

pkg/controller/infrastructure/infraflow/reconcile.go
```diff
@@ -522,22 +568,46 @@ func (c *FlowContext) ensureZones(ctx context.Context) error {
...
-				AvailabilityZone: zone.Name,
+				Tags:      tagsWorkers,
+				VpcId:     c.state.Get(IdentifierVPC),
+				CidrBlock: zone.Workers,
+				Ipv6CidrBlocks: func() []string {
+					if subnetCIDR1 != "" {
+						return []string{subnetCIDR1}
+					} else {
+						return nil
+					}
```
ScheererJ, line 579: If there was an array of type `[]*string` we could get rid of this function ugliness...

pkg/controller/infrastructure/infraflow/reconcile.go
```diff
@@ -522,22 +568,46 @@ func (c *FlowContext) ensureZones(ctx context.Context) error {
...
+				}(),
+				AvailabilityZone:            zone.Name,
+				AssignIpv6AddressOnCreation: pointer.Bool(false),
```
ScheererJ, line 582: What does this mean? What would happen if we would set this to `true`? Would that automatically assign a `/64` block to the subnet so that we would not have to do the IP address management?

pkg/controller/infrastructure/infraflow/reconcile.go
```diff
@@ -1382,3 +1452,30 @@ func (c *FlowContext) getZone(item *awsclient.Subnet) *aws.Zone {
...
+
+	baseIP := ipNet.IP
+	maskSize, addrSize := ipNet.Mask.Size()
+
+	if newPrefixLength <= maskSize || newPrefixLength > addrSize {
+		return "", fmt.Errorf("invalid new prefix length")
+	}
+
+	subnetSize := big.NewInt(1)
+	subnetSize = subnetSize.Lsh(subnetSize, uint(addrSize-newPrefixLength))
```
ScheererJ, line 1470: Why not inline this?

```suggestion
	subnetSize := subnetSize.Lsh(big.NewInt(1), uint(addrSize-newPrefixLength))
```

pkg/controller/infrastructure/infraflow/reconcile.go
```diff
@@ -1382,3 +1452,30 @@ func (c *FlowContext) getZone(item *awsclient.Subnet) *aws.Zone {
...
+	offset := big.NewInt(int64(index))
+	offset = offset.Mul(offset, subnetSize)
```
ScheererJ, line 1472: Let's make it concise.

```suggestion
	offset := offset.Mul(big.NewInt(int64(index)), subnetSize)
```

pkg/controller/infrastructure/infraflow/reconcile.go
```diff
@@ -1382,3 +1452,30 @@ func (c *FlowContext) getZone(item *awsclient.Subnet) *aws.Zone {
...
+
+	ipInt := big.NewInt(0)
+	ipInt.SetBytes(baseIP)
+	ipInt = ipInt.Add(ipInt, offset)
```
ScheererJ, line 1476: Why not shorter?

```suggestion
	ipInt := ipInt.Add((&big.Int{}).SetBytes(baseIP), offset)
```

pkg/controller/infrastructure/templates/main.tpl.tf
```diff
@@ -84,6 +87,18 @@ resource "aws_route" "public" {
...
 }
 
+{{ if and .enableDualstack .create.vpc }}
+resource "aws_route" "public-ipv6" {
+  route_table_id         = aws_route_table.routetable_main.id
+  destination_ipv6_cidr_block = "::/0"
+  gateway_id             = {{ .vpc.internetGatewayID }}
+
+  timeouts {
+    create = "5m"
```
ScheererJ, line 97: Does adding a route really take so long?
Why do we not specify a delete timeout?

pkg/controller/infrastructure/templates/main.tpl.tf
```diff
@@ -138,7 +153,10 @@ resource "aws_subnet" "nodes_z{{ $index }}" {
   vpc_id            = {{ $.vpc.id }}
   cidr_block        = "{{ $zone.worker }}"
   availability_zone = "{{ $zone.name }}"
-
+{{- if and $.enableDualstack $.create.vpc}}
+  ipv6_cidr_block = "${cidrsubnet(aws_vpc.vpc.ipv6_cidr_block, 8, (1 + ({{ $index }} * 3)))}"
```
ScheererJ, line 157: Similar to the flow-based approach, we do not use the first block, but start with offset 1. Why is that?

test/integration/infrastructure/infrastructure_test.go
```diff
@@ -225,6 +225,19 @@ var _ = Describe("Infrastructure tests", func() {
 			Expect(err).NotTo(HaveOccurred())
 		})
 
+		It("should successfully create and delete (flow) with dualstack enabled", func() {
+			providerConfig := newProviderConfig(awsv1alpha1.VPC{
+				CIDR:             pointer.String(vpcCIDR),
+				GatewayEndpoints: []string{s3GatewayEndpoint},
```
ScheererJ, line 231: Why is this necessary?

```suggestion
```

test/integration/infrastructure/infrastructure_test.go
```diff
@@ -238,6 +251,19 @@ var _ = Describe("Infrastructure tests", func() {
 			Expect(err).NotTo(HaveOccurred())
 		})
 
+		It("should successfully create and delete (terraformer) with dualstack enabled", func() {
+			providerConfig := newProviderConfig(awsv1alpha1.VPC{
+				CIDR:             pointer.String(vpcCIDR),
+				GatewayEndpoints: []string{s3GatewayEndpoint},
```
ScheererJ, line 257: Same as above.

ScheererJ (commented):
Thanks a lot for the change. I have a few questions/remarks.

pkg/controller/infrastructure/infraflow/reconcile.go
axel7born: No, if we set it to `true`,  ipv6 addresses would automatically be assigned to instances.

pkg/controller/infrastructure/infraflow/reconcile.go
DockToFuture: subnetSize is still unknown at the begining

axel7born: Maybe like this:
```
subnetSize := big.NewInt(1).Lsh(big.NewInt(1), uint(addrSize-newPrefixLength))
```
But I'm not sure if this is really nicer.

pkg/controller/infrastructure/infraflow/reconcile.go
axel7born: The shorter version would look like this
```
ipInt := big.NewInt(0).SetBytes(baseIP).Add(big.NewInt(0).SetBytes(baseIP), offset)
```
I think the long version is easier to read and understand.

pkg/controller/infrastructure/templates/main.tpl.tf
axel7born: `timeouts` is copied from the route above. I wouldn't change it in the PR.

pkg/controller/infrastructure/infraflow/reconcile.go
```diff
@@ -572,8 +556,8 @@ func (c *FlowContext) ensureZones(ctx context.Context) error {
 				VpcId:     c.state.Get(IdentifierVPC),
 				CidrBlock: zone.Workers,
 				Ipv6CidrBlocks: func() []string {
-					if subnetCIDR1 != "" {
-						return []string{subnetCIDR1}
+					if len(subnetCIDRs) == 3 && subnetCIDRs[0] != "" {
+						return []string{subnetCIDRs[0]}
```
axel7born, line 560: I guess, we could also do the append in a loop now?

pkg/controller/infrastructure/infraflow/reconcile.go
ScheererJ: If you want to go for the long version please use more than one variable name. The result may be `ipInt`, but in between it is not.

test/integration/infrastructure/infrastructure_test.go
axel7born: This is needed as in verifyCreation the vpc endpoint is checked https://github.com/DockToFuture/gardener-extension-provider-aws/blob/8c611e76de08b59c9818e7dce46cfd8e7fac349d/test/integration/infrastructure/infrastructure_test.go#L933
It's the same for all tests.

pkg/aws/client/client.go
```diff
@@ -673,15 +673,50 @@ func (c *Client) DeleteVpcDhcpOptions(ctx context.Context, id string) error {
...
+							}
+							vpc.IPv6CidrBlock = ipv6CidrBlock
+							return vpc, nil
+						}
+					}
+				}
+			}
+		}
+		return nil, fmt.Errorf("No IPv6 CIDR Block was assigned to VPC")
+	}
```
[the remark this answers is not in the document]
axel7born: Done.

pkg/aws/client/client.go
```diff
@@ -1347,10 +1492,13 @@ func (c *Client) UpdateSubnetAttributes(ctx context.Context, desired, current *S
 			AssignIpv6AddressOnCreation: toAttributeBooleanValue(desired.AssignIpv6AddressOnCreation),
 			SubnetId:                    aws.String(current.SubnetId),
 		}
-		if _, err := c.EC2.ModifySubnetAttributeWithContext(ctx, input); err != nil {
-			return false, fmt.Errorf("updating AssignIpv6AddressOnCreation failed: %w", err)
+		associated, _ := c.CheckSubnetIPv6Cidr(current.SubnetId)
+		if trueOrFalse(desired.AssignIpv6AddressOnCreation) && current.Ipv6CidrBlocks != nil && associated {
```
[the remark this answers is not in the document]
axel7born: The whole condition didn't make sense anymore. Removed.

ScheererJ (changes requested):
Thanks for adapting the pull request. I have a few more comments.

docs/usage/usage.md
```diff
@@ -185,6 +186,8 @@ It only applies for those worker pools whose `iamInstanceProfile` is not set.
   }
   ```
 </details>
+The `enableDualStack` flag specifies whether dual-stack or IPv4-only should be supported by the infrastructure.
```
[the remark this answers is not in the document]
axel7born: Added the newline.

pkg/apis/aws/v1alpha1/types_infrastructure.go
```diff
@@ -31,6 +31,9 @@ type InfrastructureConfig struct {
 	// +optional
 	EnableECRAccess *bool `json:"enableECRAccess,omitempty"`
 
+	// EnableDualStack specifies whether dual-stack or IPv4 should be supported.
```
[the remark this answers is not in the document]
axel7born: Done.

hack/api-reference/api.md
```diff
@@ -187,6 +187,17 @@ default: true</p>
 </tr>
 <tr>
 <td>
+<code>enableDualStack</code></br>
+<em>
+bool
+</em>
+</td>
+<td>
+<p>EnableDualStack specifies whether dual-stack or IPv4 should be supported.</p>
```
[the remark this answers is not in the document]
axel7born: Generated.

pkg/controller/infrastructure/infraflow/reconcile.go
```diff
@@ -1382,3 +1438,22 @@ func (c *FlowContext) getZone(item *awsclient.Subnet) *aws.Zone {
...
+	}
+
+	baseIP := ipNet.IP
+	maskSize, addrSize := ipNet.Mask.Size()
+
+	if newPrefixLength <= maskSize || newPrefixLength > addrSize {
+		return "", fmt.Errorf("invalid new prefix length")
+	}
+
+	offset := big.NewInt(int64(index)).Mul(big.NewInt(int64(index)), big.NewInt(1).Lsh(big.NewInt(1), uint(addrSize-newPrefixLength)))
```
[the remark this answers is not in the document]
axel7born: I use 0 as value for `big` objects that are directly overwritten.

pkg/controller/infrastructure/infraflow/reconcile.go
```diff
@@ -1382,3 +1438,22 @@ func (c *FlowContext) getZone(item *awsclient.Subnet) *aws.Zone {
...
+	subnetIP := net.IP(big.NewInt(0).SetBytes(baseIP).Add(big.NewInt(0).SetBytes(baseIP), offset).Bytes())
```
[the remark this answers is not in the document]
axel7born: I use `0` as value for `big` objects that are directly overwritten. I hope that make the code less confusing.

pkg/controller/infrastructure/infraflow/reconcile.go
```diff
@@ -375,17 +399,24 @@ outer:
...
 				GatewayId:            c.state.Get(IdentifierInternetGateway),
 			},
 		},
 	}
+	if c.state.Get(IdentifierVpcIPv6CidrBlock) != nil {
+		desired.Routes = append(desired.Routes, &awsclient.Route{
+			DestinationIpv6CidrBlock: pointer.String(allIPv6),
+			GatewayId:                c.state.Get(IdentifierInternetGateway),
+		})
+	}
```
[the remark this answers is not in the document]
DockToFuture: Shall I split it in two tasks?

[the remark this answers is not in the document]
DockToFuture: Or do we want to do it in a separate pr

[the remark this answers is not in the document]
kon-angelo: No, I think for now let's keep like it is

ScheererJ (dismissed):
/lgtm

kon-angelo (changes requested):
@DockToFuture, @axel7born  code-wise the change is okay. I still have some questions from a conceptual point of view that I would like you to help me clarify.

From my understanding the dual stack support is split into 2 parts:
1. Talking ipv6 with the "outside world".
2. ipv4/ipv6 support within the cluster.

The first part is handled by a combination of AWS LB Controller and this PR. I think the most relevant part is the ipv6 route to the IGW. 

The rest of the PR though performs additional changes to the VPC and subnet. The result of these changes combined with your other [PR](https://github.com/gardener/gardener-extension-provider-aws/pull/785/files#diff-ede719e7797a6dd04b91acb4296b78434730159686b71e616b1c989712a7716e) is that the 2 parts I mentioned above are coupled together under the same configuration option and you have some side effects like global VM rollout.

TLDR; my questions are:

1. Is this whole PR the actual minimum needed to support case the ipv6 LB ? Meaning that the VPC and subnet changes are necessary to support the IPv6-enabled LB.
2. See also [my comment on the other pr](https://github.com/gardener/gardener-extension-provider-aws/pull/785#issuecomment-1645295877). If [1] holds, The coupled change as it is now will mean that probably we need to enforce some restriction on the k8s versions that we can reasonable support this, which depend on the CCM version used.
3. If [2] holds then don't we need a validation to prevent clusters from switching this off ? The problem is that via normal means, the reconciliation of the infra takes place before the machines, which means that if you try to switch this off we would try to reconcile away the subnet addresses while we have machines on.

DockToFuture:
> @DockToFuture, @axel7born code-wise the change is okay. I still have some questions from a conceptual point of view that I would like you to help me clarify.
> 
> From my understanding the dual stack support is split into 2 parts:
> 
> 1. Talking ipv6 with the "outside world".
> 2. ipv4/ipv6 support within the cluster.
> 
> The first part is handled by a combination of AWS LB Controller and this PR. I think the most relevant part is the ipv6 route to the IGW.
> 
> The rest of the PR though performs additional changes to the VPC and subnet. The result of these changes combined with your other [PR](https://github.com/gardener/gardener-extension-provider-aws/pull/785/files#diff-ede719e7797a6dd04b91acb4296b78434730159686b71e616b1c989712a7716e) is that the 2 parts I mentioned above are coupled together under the same configuration option and you have some side effects like global VM rollout.
> 
> TLDR; my questions are:
> 
> 1. Is this whole PR the actual minimum needed to support case the ipv6 LB ? Meaning that the VPC and subnet changes are necessary to support the IPv6-enabled LB.
> 2. See also [my comment on the other pr](https://github.com/gardener/gardener-extension-provider-aws/pull/785#issuecomment-1645295877). If [1] holds, The coupled change as it is now will mean that probably we need to enforce some restriction on the k8s versions that we can reasonable support this, which depend on the CCM version used.
> 3. If [2] holds then don't we need a validation to prevent clusters from switching this off ? The problem is that via normal means, the reconciliation of the infra takes place before the machines, which means that if you try to switch this off we would try to reconcile away the subnet addresses while we have machines on.

This PR is required to get dualstack LBs. The other PRs https://github.com/gardener/gardener-extension-provider-aws/pull/785 and https://github.com/gardener/machine-controller-manager-provider-aws/pull/128 are required for enabling IPv6 in the shoot cluster together with a lot more changes which needs to be done.

kon-angelo:
> This PR is required to get dualstack LBs. The other PRs https://github.com/gardener/gardener-extension-provider-aws/pull/785 and https://github.com/gardener/machine-controller-manager-provider-aws/pull/128 are required for enabling IPv6 in the shoot cluster together with a lot more changes which needs to be done.

Since the changes to the VPC and subnet are necessary then this PR is the "MVP" for the dual stack LB. Got it 👍 

On it's own this PR is okay. One thing that may be missing is some validation: We discussed with @axel7born briefly about doing some tests to see if we can disable the dualstack after being enabled and depending on that we may need to add validation to prevent it.

My concern was that in tandem with the other PRs you get some side-effects affecting the Worker resource. In particular, assuming you have a dualstack enabled cluster, then when #785 is deployed it would cause node rollouts.

docs/usage/usage.md
```diff
@@ -186,6 +188,9 @@ It only applies for those worker pools whose `iamInstanceProfile` is not set.
   ```
 </details>
 
+The `dualStack.enabled` flag specifies whether dual-stack or IPv4-only should be supported by the infrastructure.
+When the flag is set to true an amazon provided IPv6 CIDR block will be attached to the VPC, all subnets will receive a `/64` block from it and a route entry is added to the main route table to route all IPv6 traffic over the IGW.
```
MartinWeindel, line 192:
```suggestion
When the flag is set to true an Amazon provided IPv6 CIDR block will be attached to the VPC.
All subnets will receive a `/64` block from it and a route entry is added to the main route table to route all IPv6 traffic over the IGW. 
```

MartinWeindel: minor improvement

MartinWeindel (approved):
/lgtm only nits in docu

axel7born:
I tested switching from `DualStack.Enabled = true` to `DualStack.Enabled = false`. With terraform we get an error in this case, as it can't remove the IPv6 CIDR from the VPC as there are subnets with IPv6 CIDRs assigned. Without terraform it simply doesn't do anything, as we didn't implement that case.
I didn't test with a deployed dualstack LB as I didn't come that far anyway.
I added validation to prevent DualStack from being switched-off.

kon-angelo (approved):
/lgtm

kon-angelo (approved):
/lgtm

PR #670, opened 2022-12-12
Improve logging for csi-driver-node

MartinWeindel:
**How to categorize this PR?**

/area control-plane
/kind enhancement
/platform aws

**What this PR does / why we need it**:
Improve logging for csi-driver-node to log mount operation details.

**Which issue(s) this PR fixes**:
Fixes #418

**Special notes for your reviewer**:

**Release note**:

```other operator
Improve logging for csi-driver-node to provide mount operation details.
```

PR #782, opened 2023-07-12
Increase ebs log level

kon-angelo:

/area control-plane
/kind enhancement
/platform aws

Fixes #

**Release note**:

```other operator
NONE
```

### DIFF
--- a/charts/internal/shoot-system-components/charts/csi-driver-node/templates/daemonset.yaml
+++ b/charts/internal/shoot-system-components/charts/csi-driver-node/templates/daemonset.yaml
@@ -49,7 +49,7 @@ spec:
         - --volume-attach-limit={{ .Values.driver.volumeAttachLimit }}
         {{- end }}
         - --logtostderr
-        - --v=3
+        - --v=4
         - --vmodule=node_*=4,mount*=4,driver=4,controller=4
         env:
         - name: CSI_ENDPOINT


charts/internal/shoot-system-components/charts/csi-driver-node/templates/daemonset.yaml
```diff
@@ -49,7 +49,7 @@ spec:
         - --volume-attach-limit={{ .Values.driver.volumeAttachLimit }}
         {{- end }}
         - --logtostderr
-        - --v=3
+        - --v=4
         - --vmodule=node_*=4,mount*=4,driver=4,controller=4
```
MartinWeindel, line 53: If verbosity should stay at 4, you can delete the `--vmodule` line. It was introduced to increase the verbosity selectively, see https://github.com/gardener/gardener-extension-provider-aws/pull/670

MartinWeindel (approved):

charts/internal/shoot-system-components/charts/csi-driver-node/templates/daemonset.yaml
kon-angelo: On second thought, the `vmodule` is the correct approach. However, I am not sure if the driver is indeed logging everything properly. Particularly, while I was debugging an issue it seems to not log the beginning of staging operations which I would expect to and led me to this PR. 

Let me double-check before proceeding with the PR.

kon-angelo: The problem is that that the `NodeGetCapabilities` which is what causes the logging noise is on the `node.go` file along with the other useful operations. With the logging as it is setup now you can still peak into some internal logs, but you are missing when the operations like staging and publishing are beginning and ending successfully. 

I will close the PR then until I can come up with a better solution to the problem. Thank you @MartinWeindel for catching this

PR #785, opened 2023-07-18
Set ipv6AddressCount and ipv6PrefixCount when dualstack is enabled.

axel7born:

/area networking
/kind enhancement
/platform aws

This PR is needed to add IPv6 support to Gardener by obtaining globally routable IPv6 addresses from AWS. We use Ipv6AddressCount to automatically assign IPv6 addresses from the subnet range. Additionally, we use Ipv6PrefixCount to get delegated prefixes for the pod CIDR

Depends on 
* https://github.com/gardener/machine-controller-manager-provider-aws/pull/128
* https://github.com/gardener/gardener-extension-provider-aws/pull/778 
Fixes #

**Release note**:

```feature operator
Set ipv6AddressCount and ipv6PrefixCount when dualstack is enabled.
```

### DIFF
--- a/pkg/controller/worker/machines.go
+++ b/pkg/controller/worker/machines.go
@@ -100,6 +100,13 @@ func (w *workerDelegate) generateMachineConfig() error {
 			}
 		}
 
+		infrastructureConfig := &awsapi.InfrastructureConfig{}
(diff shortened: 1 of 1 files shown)


kon-angelo:
I think its worth mentioning in the description that an IPv6 IPAM is necessary so we need to wait for AWS CCM versions that support the behavior. Also that restriction means that we can only enable the PR in the "backported" CCM versions that get the IPAM, whomever these are.

axel7born:
Included in #1024

PR #789, opened 2023-08-08
Update region config example

timuthy:

/area scalability
/kind enhancement
/platform aws

This PR updates the region config example.

**Release note**:

```other operator
NONE
```


kon-angelo (approved):
/lgtm

PR #790, opened 2023-08-11
No caching of secrets

MartinWeindel:

/area control-plane
/area cost
/kind enhancement
/platform aws

By default, the controller-runtime caches all objects of kinds which are accessed.
On very large seeds caching secrets can consume GBs of memory.
Therefore it is better to disable caching of secret objects.

Fixes #

I made a small benchmark on an empty seed with 230 secrets (150 big secrets added to make differences more visibile).

Memory usage with caching of secrets:
```bash
$ k -n extension-provider-aws-zj2jw top po
NAME                                               CPU(cores)   MEMORY(bytes)
gardener-extension-provider-aws-6cb999694d-fnfhz   5m           202Mi
gardener-extension-provider-aws-6cb999694d-nk6ln   1m           173Mi
```
During startup, the memory went even shortly up to 500Mi and high CPU usage was also seen.

After disabling caching secrets:
```bash
$ k -n extension-provider-aws-zj2jw top po
NAME                                               CPU(cores)   MEMORY(bytes)
gardener-extension-provider-aws-5bf5ddf647-rdjpr   2m           27Mi
gardener-extension-provider-aws-5bf5ddf647-txkd4   5m           35Mi
```

**Release note**:

```other operator
No caching of secrets
```

### DIFF
--- a/cmd/gardener-extension-provider-aws/app/app.go
+++ b/cmd/gardener-extension-provider-aws/app/app.go
@@ -32,10 +32,12 @@ import (
 	gardenerhealthz "github.com/gardener/gardener/pkg/healthz"
 	machinev1alpha1 "github.com/gardener/machine-controller-manager/pkg/apis/machine/v1alpha1"
 	"github.com/spf13/cobra"
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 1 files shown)


MartinWeindel:
For special case `IgnoreOperationAnnotation=true` we need also https://github.com/gardener/gardener/pull/8348.
But I don't know when this flag is used.

PR #800, opened 2023-09-15
Vendor latest `github.com/gardener/gardener-networking-extension-{calico-cilium}`

acumino:
**How to categorize this PR?**

/area open-source
/kind enhancement
/platform aws

**What this PR does / why we need it**:
This PR vendor `github.com/gardener/gardener-networking-extension-calico` `v1.36.0` and `github.com/gardener/gardener-networking-extension-cilium` `v1.28.0` and drop excludes for client.go.

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:
/invite @ialidzhikov @ScheererJ 

**Release note**:

```other operator
NONE
```

PR #807, opened 2023-09-27
Remove dependency to specific calico and cilium versions.

axel7born:

/area networking
/kind technical-debt
/platform aws

This PR removes the dependency to specific calico and cilium versions.
After each api change of a network extension, the new version needed to be vendored here and a new release has to be made in order to pass admission.
By parsing the network into a general map instead into an object of type `NetworkConfig` we can get rid of this dependency. 

Fixes #

**Release note**:

```noteworthy developer
Remove dependency to specific calico and cilium versions.
```


pkg/admission/mutator/shoot.go
```diff
@@ -89,70 +86,37 @@ func (s *shoot) Mutate(_ context.Context, newObj, oldObj client.Object) error {
 	overlayDisabled := false
 
 	if shoot.Spec.Networking != nil && shoot.Spec.Networking.Type != nil {
```
ScheererJ, line 88: Do we need to check for `shoot.Spec.Networking.Type != nil`? It is not strictly wrong, but probably not required anymore.

pkg/admission/mutator/shoot.go
```diff
@@ -89,70 +86,37 @@ func (s *shoot) Mutate(_ context.Context, newObj, oldObj client.Object) error {
...
-			shoot.Spec.Networking.ProviderConfig = &runtime.RawExtension{
-				Object: networkConfig,
-			}
+		networkConfig, err := s.decodeNetworkConfig(shoot.Spec.Networking.ProviderConfig)
+		if err != nil {
+			return err
+		}
 
-		case cilium.ReleaseName:
-			overlay := &ciliumv1alpha1.Overlay{Enabled: false}
```
ScheererJ, line 126: Do we want to use constants for `overlay` and `enabled` to prevent typos? These strings are used multiple times here.

ScheererJ (changes requested):
Thanks for the nice improvement. It will certainly make the delivery process more flexible.

ScheererJ (approved):
/lgtm

DockToFuture (approved):
/lgtm

shafeeqes:
You could also remove the `replace k8s.io/client-go => k8s.io/client-go v0.26.4` redirective, then we don't need https://github.com/gardener/gardener-extension-provider-aws/pull/800

pkg/admission/mutator/shoot.go
```diff
@@ -201,22 +171,15 @@ func (s *shoot) decodeControlplaneConfig(controlPlaneConfig *runtime.RawExtensio
...
 
-func (s *shoot) decodeCalicoNetworkConfig(network *runtime.RawExtension) (*calicov1alpha1.NetworkConfig, error) {
-	networkConfig := &calicov1alpha1.NetworkConfig{}
-	if network != nil && network.Raw != nil {
-		if _, _, err := s.decoder.Decode(network.Raw, nil, networkConfig); err != nil {
-			return nil, err
+func (s *shoot) decodeNetworkConfig(network *runtime.RawExtension) (map[string]interface{}, error) {
+	var networkConfig map[string]interface{}
+	if network == nil || network.Raw == nil {
+		network = &runtime.RawExtension{
```
kon-angelo, line 177: maybe you can return here directly - skipping unmarshal ?

kon-angelo (approved):
/lgtm

PR #820, opened 2023-11-02
Switch image registry for aws-ebs-csi-driver to support IPv6

DockToFuture:

/area networking
/kind enhancement
/platform aws

Switch image registry for `aws-ebs-csi-driver` to `registry.k8s.io` to support IPv6.

Fixes #

**Release note**:

```other operator
The registry of the aws-ebs-csi-driver image is switched from ECR (`public.ecr.aws`) to `registry.k8s.io` because the ECR does not support image pulls over IPv6.
```


ialidzhikov (approved):
/lgtm

PR #847, opened 2023-12-29
[dnsrecords] create AAAA alias targets for load balancers with IPv6 addresses

MartinWeindel:

/area control-plane
/kind enhancement
/platform aws

On creating a `DNSRecord` for a AWS load balancer, the requested `CNAME`  record was always mapped to an `A` alias target recordset only. 
With this PR, creation of  `A` and/or `AAAA` recordsets can be controlled with the annotation `dns.gardener.cloud/ip-stack`. Possible values for the annotation are `ipv4`, `ipv6`, and `dual-stack`. In this way, e.g. dual-stack load balancers for the apiserver can be supported.

Additionally, the list of canonical hosted zone IDs was updated with some newer AWS zones.

Fixes #

**Release note**:

```other operator
[dnsrecords] AAAA alias targets for load balancers with IPv6 addresses with annotation `dns.gardener.cloud/ip-stack=dual-stack`
```

### DIFF
--- a/pkg/aws/client/client_route53.go
+++ b/pkg/aws/client/client_route53.go
@@ -25,6 +25,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/service/route53"
+	"github.com/gardener/external-dns-management/pkg/controller/provider/aws/data"
 )
 
 // GetDNSHostedZones returns a map of all DNS hosted zone names mapped to their IDs.
@@ -91,8 +92,9 @@ func normalizeZoneId(zoneId string) string {
 
 // CreateOrUpdateDNSRecordSet creates or updates the DNS recordset in the DNS hosted zone with the given zone ID,
 // with the given name, type, values, and TTL.
-func (c *Client) CreateOrUpdateDNSRecordSet(ctx context.Context, zoneId, name, recordType string, values []string, ttl int64) error {
-	rrs := newResourceRecordSet(name, recordType, newResourceRecords(recordType, values), ttl)
+// A CNAME record for AWS load balancers in a known zone may be mapped to A and/or AAAA recordsets with alias target.
+func (c *Client) CreateOrUpdateDNSRecordSet(ctx context.Context, zoneId, name, recordType string, values []string, ttl int64, stack IPStack) error {
+	rrs := newResourceRecordSets(name, recordType, newResourceRecords(recordType, values), ttl, stack)
 	if err := c.waitForRoute53RateLimiter(ctx); err != nil {
 		return err
 	}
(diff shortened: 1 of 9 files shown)


pkg/aws/client/client_route53.go
```diff
@@ -216,40 +224,80 @@ func newResourceRecords(recordType string, values []string) []*route53.ResourceR
...
 		}
 	}
-	return recordType
+	return false
 }
 
 var (
 	// original code: https://github.com/kubernetes-sigs/external-dns/blob/master/provider/aws/aws.go
 	// see: https://docs.aws.amazon.com/general/latest/gr/elb.html
+	// Please keep in sync with list in https://github.com/gardener/external-dns-management/blob/master/pkg/controller/provider/aws/aliastarget.go
```
kon-angelo, line 300: Is it worth it to "package" this and consume it, compared to maintaining the copy? The package can be in any of provider-aws/external-dns-management.

MartinWeindel: We would need to refactor external-dns-management first to avoid introducing ghost dependencies.
The list will not change so often and it is not critical to have the latest one.
I will give it a try.

MartinWeindel: After refactoring external-dns-management (https://github.com/gardener/external-dns-management/pull/348), the table is now reused from from there.
@kon-angelo Thanks for the hint!

pkg/aws/client/client_route53.go
```diff
@@ -91,83 +93,88 @@ func normalizeZoneId(zoneId string) string {
...
+	out, err := c.Route53.ListResourceRecordSetsWithContext(ctx, input)
 	if ignoreResourceRecordSetNotFound(err) != nil {
 		return nil, err
 	}
 	if out == nil || len(out.ResourceRecordSets) == 0 { // no records in zone
 		return nil, nil
 	}
-	if rrs := out.ResourceRecordSets[0]; normalizeName(aws.StringValue(rrs.Name)) == name && aws.StringValue(rrs.Type) == recordType {
-		return out.ResourceRecordSets[0], nil
+	var toDelete []*route53.ResourceRecordSet
```
ScheererJ, line 164: Why is this variable called `toDelete` when it has nothing to do with deletion in this getter function? The function result may be used for deletion, but here I find the name rather confusing.

ScheererJ (commented):
Thanks a lot for this nice improvement.

I just have a minor question.

pkg/aws/client/client_route53.go
MartinWeindel: Good point. Renamed it to `recordSets` to avoid imposing purpose of method call.

ScheererJ (dismissed):
/lgtm

MartinWeindel:
/hold
`AAAA` records are created if a DNS lookup of the target returns IPv6 addresses or if the DNS lookup query fails. Although the additional `AAAA` record make no harm in the later case, it is a wasted record and seems to happen much more frequently than expected.
Needs to be discussed how to trigger it, e.g. by some annotation on the DNSRecord

kon-angelo (approved):
/lgtm

PR #853, opened 2024-01-09
Inserts `Architecture`  from worker to machineclass

sssash18:

/area control-plane
/kind enhancement
/platform aws

This PR  takes the `architecture` field from worker and then inserts it in the machine class
Fixes Part of  #https://github.com/gardener/autoscaler/issues/122

**Release note**:

```other operator
Inserts `architecture` from worker to the machine class
```

### DIFF
--- a/charts/internal/machineclass/templates/machineclass.yaml
+++ b/charts/internal/machineclass/templates/machineclass.yaml
@@ -27,6 +27,7 @@ metadata:
     {{- end }}
 {{- if $machineClass.nodeTemplate }}
 nodeTemplate:
+  architecture: {{$machineClass.nodeTemplate.architecture}}
   capacity:
 {{ toYaml $machineClass.nodeTemplate.capacity | indent 4 }}
   instanceType: {{ $machineClass.nodeTemplate.instanceType }}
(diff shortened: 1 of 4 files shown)


charts/internal/machineclass/templates/machineclass.yaml
```diff
@@ -24,6 +24,7 @@ metadata:
 {{- end }}
 {{- if $machineClass.nodeTemplate }}
 nodeTemplate:
+  architecture: {{$machineClass.nodeTemplate.architecture}}
```
kon-angelo, line 27:
```suggestion
  architecture: {{ $machineClass.nodeTemplate.architecture }}
```

kon-angelo: Nit: make the change use the same formatting as the rest of the file

kon-angelo (changes requested):

charts/internal/machineclass/templates/machineclass.yaml
sssash18: done

rishabh-11 (commented):
Thanks for the changes. @kon-angelo gardener now uses `0.53.0` version of MCM, and it already has the required field.

/lgtm

kon-angelo (approved):
/lgtm

PR #854, opened 2024-01-10
ECR credential provider 

kon-angelo:

/area control-plane
/kind enhancement
/platform aws

After kubernetes v1.27 the kubelet will no longer try to authenticate with ECR for private registries. Instead the extensible kubelet mechanism for image credential providers should be used. This PR configures the kubelet and the gardener-node-agent to pull the ecr-credentials-binary from an image. 

Unfortunately the ecr-credential-provider is not published by https://github.com/kubernetes/cloud-provider-aws in any image, hence this PR introduces an additional image to download the binary and build an image so that it can be used by the gardener-node-agent. 

Fixes #

In case https://github.com/kubernetes/cloud-provider-aws/pull/825 gets merged, then we will replace our custom dockerfile with the upstream CCM image.

**Release note**:

```feature user
Adapt Kubelet to use ecr-credential-provider on k8s >=v1.27
```


pkg/webhook/controlplane/ensurer.go
```diff
@@ -449,7 +529,39 @@ done
...
+	if err != nil {
+		return err
+	}
+
+	// return early
+	if !k8sGreaterEqual127 {
+		return nil
+	}
+
+	binConfig, err := e.ensureCredentialHelperBin()
```
AndreasBurger, line 554: Also consider changing the function name as nothing is ensured by them. Basically they just generate the file's representation the ensurer can work with, so maybe just `getCredentialHelperBinConfig` and `getCredentialHelperBinConfigFileConfig()`. Though admittedly the latter isn't all that great either.

Though one could argue that it is ensured that an image with the `ECRCredentialProviderImageName` is part of the BoM at that point.

AndreasBurger (approved):
lgtm, some very minor naming concerns

kon-angelo:
/hold

we need to solve the issue about how to publish the ecr image.

MartinWeindel (commented):
In the source code, I only have two nits.

pkg/webhook/controlplane/ensurer.go
```diff
@@ -418,8 +438,68 @@ ExecStart=/opt/bin/mtu-customizer.sh
...
+		Permissions: &permissions,
+		Content: extensionsv1alpha1.FileContent{
+			Inline: &extensionsv1alpha1.FileContentInline{
+				Data: string(configJson),
+			},
+		},
+	}, nil
+}
+
+func (e *ensurer) ensureMTUFiles() *extensionsv1alpha1.File {
```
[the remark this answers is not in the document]
kon-angelo: I only do it because I would have to return an actual object in my return statements (e.g. in each error checking statement) and I think it looks more simplified if I instead returned nil -> making the return object a pointer.

MartinWeindel (approved):
/lgtm

imagevector/images.yaml
```diff
@@ -272,3 +272,16 @@ images:
       confidentiality_requirement: 'low'
       integrity_requirement: 'high'
       availability_requirement: 'low'
+- name: ecr-credential-provider
+  sourceRepository: github.com/gardener/cloud-provider-aws
```
AndreasBurger, line 276: I'd have expected this to be `github.com/gardener/ecr-credential-provider` or `github.com/kubernetes/cloud-provider-aws`. Is there a reason to refer to this repository here, as it isn't even indirectly involved with the image?

AndreasBurger (approved):
lgtmm

AndreasBurger (approved):
lgtm

PR #857, opened 2024-01-17
Fix hook-me script

kon-angelo:

/area dev-productivity
/kind regression
/platform aws

Adapt scripts after last g/g vendoring

Fixes #

**Release note**:

```other operator
NONE
```


timuthy (approved):
Thank you 🙂 
/lgtm

Makefile
```diff
@@ -68,18 +68,28 @@ start:
...
 		-ldflags $(LD_FLAGS) \
 		./cmd/$(EXTENSION_PREFIX)-$(ADMISSION_NAME) \
 		--webhook-config-server-host=0.0.0.0 \
 		--webhook-config-server-port=$(WEBHOOK_CONFIG_PORT) \
 		--webhook-config-mode=$(WEBHOOK_CONFIG_MODE) \
         $(WEBHOOK_PARAM)
 
+.PHONY: hook-me
+hook-me:
+	bash $(GARDENER_HACK_DIR)/hook-me.sh $(EXTENSION_NAMESPACE) $$(kubectl get namespace | grep $(NAME) | head -n1| cut -d" " -f1) $(WEBHOOK_CONFIG_PORT)
```
timuthy, line 91: No it's inconsistent to https://github.com/gardener/gardener-extension-provider-openstack/blob/4531784c98d2e6339bc94f296f56484bb83bfa37/Makefile#L90. Why do we do it differently here?

kon-angelo: This has a bit more smart to find the proper namespace of the extension in a seed which usually has a suffix. It is not particularly smart but it does a better job at targeting the correct ns.

timuthy, line 91:
```suggestion
hook-me: $(KUBECTL)
	bash $(GARDENER_HACK_DIR)/hook-me.sh $(EXTENSION_NAMESPACE) $$(kubectl get namespace -o custom-columns=NAME:.metadata.name | grep $(NAME) | head -n1) $(WEBHOOK_CONFIG_PORT)
```

timuthy (changes requested):

PR #855, opened 2024-01-16
update ebs driver and configuration

kon-angelo:
**How to categorize this PR?**

/area control-plane
/kind enhancement
/platform aws

**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Release note**:

```dependency user
Updated csi-attacher configuration. The timeout for the attach/detach operations has been increased to 1 minute (from the default of 15 seconds) and the maximum wait period between retries has been reduced to 3 minutes (from the default of 5).
```
```dependency user
Updated aws-ebs-csi-driver -> v1.26.1
```
```dependency user
Updated volume-modifier-for-k8s -> v0.1.3
```

PR #870, opened 2024-01-26
Fix csi-volume-modifier deployment after update to v0.1.3

vpnachev:

/area control-plane
/kind bug
/platform aws

Adapt `aws-csi-volume-modifier` container in the `csi-driver-controller` deployment after the update to v0.1.3 (ref: https://github.com/gardener/gardener-extension-provider-aws/pull/855) as it is described in https://github.com/awslabs/volume-modifier-for-k8s/pull/14

Fixes #

No release notes as the update to v0.1.3 has not been released.

**Release note**:

```other operator
NONE
```

### DIFF
--- a/charts/internal/shoot-system-components/charts/csi-driver-node/templates/clusterrole-csi-volume-modifier.yaml
+++ b/charts/internal/shoot-system-components/charts/csi-driver-node/templates/clusterrole-csi-volume-modifier.yaml
@@ -13,6 +13,3 @@ rules:
   - apiGroups: [""]
     resources: ["events"]
     verbs: ["create", "patch", "update"]
-  - apiGroups: [ "coordination.k8s.io" ]
-    resources: [ "leases" ]
-    verbs: [ "get", "list", "watch", "update", "patch" ]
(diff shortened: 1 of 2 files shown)


charts/internal/shoot-system-components/charts/csi-driver-node/templates/clusterrole-csi-volume-modifier.yaml
```diff
@@ -13,3 +13,6 @@ rules:
   - apiGroups: [""]
     resources: ["events"]
     verbs: ["create", "patch", "update"]
+  - apiGroups: [ "coordination.k8s.io" ]
+    resources: [ "leases" ]
+    verbs: [ "get", "list", "watch", "update", "patch" ]
```
kon-angelo, line 18: This shouldn't be needed because of https://github.com/gardener/gardener-extension-provider-aws/blob/master/charts/internal/shoot-system-components/charts/csi-driver-node/templates/role-csi-volume-modifier.yaml

kon-angelo (changes requested):

kon-angelo (approved):
/lgtm

PR #876, opened 2024-02-02
Use `extensionswebhook.EnsureFileWithPath` instead of custom `appendUniqueFile` func

ialidzhikov:

/kind cleanup
/platform aws

When I was working on the registry-cache extension, I noticed that in various components we create `appendUniqueFile` func to append a file to the OSC files. Then, I noticed that there is the `extensionswebhook.EnsureFileWithPath` func that does the same thing.

N/A

N/A

**Release note**:

```other operator
NONE
```

### DIFF
--- a/pkg/webhook/controlplane/ensurer.go
+++ b/pkg/webhook/controlplane/ensurer.go
@@ -440,7 +440,7 @@ done
 `
 	)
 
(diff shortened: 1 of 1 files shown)


AndreasBurger (approved):

kon-angelo (approved):
/lgtm

PR #887, opened 2024-03-04
Remove obsolete test for load balancer which is known to be IPv4 only

MartinWeindel:

/area testing
/kind cleanup
/platform aws

The integration test case using a load balancer known to be IPv4 only is obsolete, as action is now relying on an annotation and not on the IP addresses of the load balancer itself.
There are separate test cases for all annotation values. This test, which was optional anyway, can be removed.
Followup of #847 

**Release note**:

```other operator
NONE
```


kon-angelo (approved):
/lgtm

PR #894, opened 2024-03-08
add operatingsystem label

tedteng:

/area ops-productivity
/kind enhancement
/platform aws

https://github.com/gardener/machine-controller-manager/issues/880 

add os information as labels in machine class objects
image

**Release note**:

```other operator
add os information as labels in machine class objects.
```

### DIFF
--- a/charts/internal/machineclass/templates/machineclass.yaml
+++ b/charts/internal/machineclass/templates/machineclass.yaml
@@ -18,9 +18,14 @@ kind: MachineClass
 metadata:
   name: {{ $machineClass.name }}
   namespace: {{ $.Release.Namespace }}
-{{- if $machineClass.labels }}
+{{- if or $machineClass.operatingSystem $machineClass.labels }}
   labels:
+    {{- if $machineClass.operatingSystem }}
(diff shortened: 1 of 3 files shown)


tedteng:
Testing result
shoot status
image
image

`machineclass` with labels
image

charts/internal/machineclass/templates/machineclass.yaml
```diff
@@ -18,9 +18,14 @@ kind: MachineClass
 metadata:
   name: {{ $machineClass.name }}
   namespace: {{ $.Release.Namespace }}
-{{- if $machineClass.labels }}
+{{- if or $machineClass.operatingSystem $machineClass.labels }}
```
kon-angelo, line 21: I actually think this condition is not necessary (and same for the {{end}} at l. 29. Can you give that a shot ?

tedteng: yes, it also works. new commit 0f346ec0 @kon-angelo

AndreasBurger (approved):

PR #933, opened 2024-04-24
Change the default capacity of volumes for `Etcd` from `80Gi` to `25Gi`.

### LINKED ISSUE #932: Change the default volume size for `Etcd`s from 80Gi to 25Gi (some smaller value without causing performance degradation)

/area cost
/area storage
/kind technical-debt
/platform aws

New volumes created on AWS must be around `25Gi` (or a similar, smaller size which can be obtained after doing more analysis), instead of over provisioning by creating volumes with `80Gi`.

**Why is this needed**:

New volumes which are provisioned for `Etcd`s on AWS currently are created with size `80Gi`. This size is a historical artifact of [gp2](https://docs.aws.amazon.com/ebs/latest/userguide/general-purpose.html#gp2-performance)'s performance characteristics. Larger volumes were provisioned so that a desirable amount of IOPS were available for the `Etcd`.

The type of volumes that are created now are not `gp2` anymore, but `gp3`; and this issue with performance is not seen in `gp3`. However, `80Gi` is still the size of the disks being created.

This can be rectified by creating smaller volumes, and thereby save on costs (especially true for HA control planes).

renormalize:

/area cost
/area storage
/kind technical-debt
/platform aws

This PR changes the capacity for volumes for `Etcd`s from 80Gi to 25Gi.

* Change the default capacity from `80Gi` to `25Gi` in `charts/gardener-extension-provider-aws/values.yaml`.

* Change the default capacity from `80Gi` to `25Gi` in `example/00-componentconfig.yaml`.

* Change tests to check if `*etcd.Spec.StorageCapacity` equals `25Gi` instead of `80Gi`.

* Run `make generate` to generate the new chart in `example/controller-registration.yaml`.

Volumes of size 80Gi are unnecessary since moving from `gp2` to `gp3`. See #932 for more.

Fixes #932

/invite @shreyas-s-rao 

**Release note**:

```feature user
New `Etcd` `gp3` volumes are now created with `25Gi` capacity instead of `80Gi` to save on storage costs.
```


shreyas-s-rao (approved):
/lgtm

shreyas-s-rao:
/hold
We will need to test how druid would react to a change in `Etcd` `Spec.StorageCapacity`, since currently it does not automatically change the PVC template for the etcd statefulset (forbidden change).

renormalize:
@shreyas-s-rao
As discussed, I've added checks to see if the older `Etcd` has an `80Gi` volume. If it does, resize to the new default `25Gi` is not performed. For any other value, regular reconciliation occurs. Please let me know if you have any suggestions.

shreyas-s-rao:
@renormalize thanks for making the change. I would suggest that you simply check whether the `oldEtcd` is nil (ie, it's a new etcd), or if the `oldEtcd.Spec.StorageCapacity` is not set, then set it to 25Gi. Or else, don't mutate it. In other words, we want to set 25Gi only for new etcds, since right now druid does not allow changing `storageCapacity` for existing etcds yet.

pkg/webhook/controlplaneexposure/ensurer_test.go
```diff
@@ -56,7 +56,7 @@ var _ = Describe("Ensurer", func() {
 			checkETCDMain(etcd)
 		})
 
-		It("should modify existing elements of etcd-main statefulset", func() {
+		It("should modify existing elements of etcd-main statefulset which are not 80Gi", func() {
```
shreyas-s-rao, line 59:
```suggestion
		It("should modify etcd storage config when etcd is new and specifies a storage config", func() {
```

The code and tests here have been copied over from g/g 4 years ago, when etcd was being as statefulset directly by g/g code (druid did not exist back then). So the test specs also talk about statefulsets rather than `Etcd` resources.

Can you please make corresponding changes to all the test spec names in this file?
- `should modify etcd-main storage config when etcd-main is new and does not specify a storage config`
- `should modify etcd-main storage config when etcd-main is new and specifies a storage config`
- `should not modify etcd-main storage config when etcd-main is old`
- `should always modify etcd-events storage config`

pkg/webhook/controlplaneexposure/ensurer_test.go
```diff
@@ -76,6 +76,34 @@ var _ = Describe("Ensurer", func() {
...
+			)
+
+			// Create ensurer
+			ensurer := NewEnsurer(etcdStorage, logger)
+
+			// Call EnsureETCDStatefulSet method and check the result
+			err := ensurer.EnsureETCD(ctx, dummyContext, newEtcd, oldEtcd)
+			Expect(err).To(Not(HaveOccurred()))
+			Expect(*newEtcd.Spec.StorageClass).To(Equal("gardener.cloud-fast"))
+			Expect(*newEtcd.Spec.StorageCapacity).To(Equal(resource.MustParse("80Gi")))
```
shreyas-s-rao, line 104: I would recommend renaming function `checkETCDMain()` to `checkNewETCDMain()` and adding another function `checkOldETCDMain()` which checks for storage capacity to be `80Gi` instead, and use `checkOldETCDMain()` here.

Also, I notice that `checkETCDEvents()` only checks for storage class and not storage capacity. Can you please add a check for storage capacity to be equal to `10Gi` for that?

shreyas-s-rao (changes requested):
@renormalize thanks for making the changes. Overall looks good. Just a few changes required in the tests. PTAL.

pkg/webhook/controlplaneexposure/ensurer_test.go
renormalize: Addressed in [9eab017](https://github.com/gardener/gardener-extension-provider-aws/pull/933/commits/9eab0176574b78dac948e23b12e4c50148b3669c).

pkg/webhook/controlplaneexposure/ensurer_test.go
renormalize: Addressed in [9eab017](https://github.com/gardener/gardener-extension-provider-aws/pull/933/commits/9eab0176574b78dac948e23b12e4c50148b3669c).

shreyas-s-rao (approved):
@renormalize thanks for making the requested changes.

/lgtm

renormalize:
I've tested the changes on an extension-provider setup through @shreyas-s-rao's guidance.

I've done the following to test the changes:

- Spawn the setup with gardener/gardener on master, with the latest release of gardener-extension-provider-aws.
- Create 2 shoots, and hibernate one of them. `Etcd` volumes created with size `80Gi`.
- Change the chart and image of the `controllerdeployment` of `provider-aws` to chart generated in this PR, and the image that is built from this PR.
- Check the volume sizes of the `Etcd`s for the two spawned shoots. They did not change.
- Create a new shoot. `Etcd` volume is created with size `25Gi`.
- Wake the hibernated shoot. Volume size did not change.
- Reconcile all three shoots. Volume size did not change.
- Change the capacity in the `Etcd` CR, to some value other than the defined value.
  - Old `Etcd`s remain at `80Gi` when changed to some other value.
  - New `Etcd`s remain at `25Gi` when changed to some other value.

Thus:

- [x] We've validated that older `Etcd`s of running shoots do not get resized due to the changes, size remains at `80Gi`.
- [x] We've validated that hibernated shoots which are woken up after the changes from this PR are deployed behave as expected, just like shoots which are already running with `80Gi`.
- [x] New shoots that are created get deployed with volumes of size `25Gi`.

/ping @gardener/gardener-extension-provider-aws-maintainers

renormalize:
/ping @gardener/gardener-extension-provider-aws-maintainers Could you take a look, since this PR directly affects storage costs?

kon-angelo (approved):
/lgtm

PR #979, opened 2024-06-11
Add validation of image availability in CloudProfile

### LINKED ISSUE #972: Validate `Shoot` against available AMI configuration in `CloudProfile`

/area robustness
/area usability
/area ops-productivity
/kind enhancement
/platform aws

The AWS admission webhook should deny shoot creations if the shoot to be created has an invalid image configuration, i.e. at least one worker pool uses an Image that is not available in that region.

**Example**:

_Shoot_
```yaml
...
spec:
  region: region: us-east-1
  workers:
  - name: worker-1
     image:
       name: gardenlinux
       version: 1443.5.0
...
```

_CloudProfile_
```yaml
...
spec:
  machineImages:
  - name: gardenlinux
     version: 1443.5.0
...
  providerConfig:
    kind: CloudProfileConfig
    machineImages:
   - name: gardenlinux
      versions:
      - regions:
        - ami: ami-000111222333
           name: eu-central-1
```

The shoot above is created but doesn't succeed during reconciliation due to the missing image configuration. Furthermore, shoot **deletion will be blocked** for the same reason (`MachineClasses`es) are re-deployed during deletion.

AndreasBurger:

/area robustness
/kind enhancement
/platform aws

With this PR, the validator validates that the configured images of workers are known in the cloud profile and denies admisssion if they are not.

Fixes #972

**Release note**:

```other operator
Validation of shoots now takes the CloudProfile into account to make sure that the configured images are defined.
```

### DIFF
--- a/pkg/apis/aws/validation/infrastructure.go
+++ b/pkg/apis/aws/validation/infrastructure.go
@@ -16,6 +16,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 
+	"github.com/gardener/gardener-extension-provider-aws/pkg/apis/aws"
 	apisaws "github.com/gardener/gardener-extension-provider-aws/pkg/apis/aws"
 )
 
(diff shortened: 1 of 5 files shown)


hebelsan:
we should check if this validation is also needed in other providers.

pkg/apis/aws/validation/infrastructure.go
```diff
@@ -65,6 +66,66 @@ func validateInfrastructureConfigZones(oldInfra, infra *apisaws.InfrastructureCo
 	return allErrs
 }
 
+// ValidateShootConfigAgainstCloudProfile validates the shoot with respect to the given (AWS) CloudProfile
+func ValidateShootConfigAgainstCloudProfile(shoot *core.Shoot, awsCloudProfile *aws.CloudProfileConfig, fldPath *field.Path) field.ErrorList {
```
kon-angelo, line 70: Anything related to "workers" better belong in the `apis/validation/shoot.go` - somewhere around https://github.com/gardener/gardener-extension-provider-aws/blob/806141bda213457d04cb950cbd1b968df18dd421/pkg/apis/aws/validation/shoot.go#L31

I would classify the image validation as part of the "worker" resource validation.

kon-angelo (changes requested):
2 more comments from my side:
a) some tests would be nice 😉 
b)The code for the validation has a very tricky edge case that needs to be handled properly. Basically, you can only validate the case of an non-existant AMI in the creation of the shoot only. This is better reflected in another piece of code particularly in https://github.com/gardener/gardener-extension-provider-aws/blob/806141bda213457d04cb950cbd1b968df18dd421/pkg/controller/worker/machine_images.go#L40

You see it could be the case that a machineImage is removed from a cloudprofile (due to expiration), but the worker should still be able to function properly. That happens because we save the currently running machineimages in the worker status and eventually merge the 2 sources. Now the issue here, is that at shoot validation you don't have access to that worker resource hence why this validation only makes sense in the "strict" creation case.

pkg/apis/aws/validation/worker.go
```diff
@@ -86,6 +86,44 @@ func ValidateWorkerConfig(workerConfig *apisaws.WorkerConfig, volume *core.Volum
...
+
+	return allErrs
+}
+
+func ValidateWorkersAgainstCloudProfileOnUpdate(_, newWorkers []core.Worker, region string, awsCloudProfile *apisaws.CloudProfileConfig, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	// Validate the existence of the images the new/updated workers are to use. Validating the images used by old workers is not possible at this point, as they might
+	// have been removed from the CloudProfile already.
+	allErrs = append(allErrs, validateWorkersConfigAgainstCloudProfile(newWorkers, region, awsCloudProfile, fldPath)...)
```
kon-angelo, line 102: I am not sure this is okay. Indeed the machine images may be gone from a cloudprofile, but we do not want to prevent other updates to the shoot spec because of that. 

Hence the set of workers that can be validated with this image checking is only the subnet of workers whose image we tried to update (or maybe even a general update to them) or new workers.

So to recap, the newWorkers here should contain the subnet of workers that:
a) are new (`not existing in oldWorkers`)
b) images changes (`newWorkers exists in oldWorkers` && `newWorker.Image != oldWorker.Image`)

WDYT?

kon-angelo (changes requested):

pkg/apis/aws/validation/worker.go
```diff
@@ -89,37 +89,47 @@ func ValidateWorkerConfig(workerConfig *apisaws.WorkerConfig, volume *core.Volum
...
-	allErrs = append(allErrs, validateWorkersConfigAgainstCloudProfile(newWorkers, region, awsCloudProfile, fldPath)...)
+	for i, newWorker := range newWorkers {
+		var w core.Worker
+		for _, oldWorker := range oldWorkers {
+			if newWorker.Name == oldWorker.Name {
+				w = oldWorker
+			}
+		}
+		// Validate only new Workers or those whose image has changed.
+		if w.Name == "" || newWorker.Machine.Image != w.Machine.Image {
```
kon-angelo, line 111: You can add a `break` after `w = oldWorker`. 
Potentially you can have a bit more elegant solution by constructing a map[string]core.Worker from the oldWorkers and doing the checks from that:
```
if v, ok := myMap[workerName]; !ok || v.Machine.Image != ...
```

It's an lgtm and the logic seems correct but I had to think couple extra seconds for the `w.Name == ""` in retrospect hence the above proposals.

kon-angelo (commented):
Overall lgtm. Maybe a test case for this convoluted update case ?

kon-angelo (approved):
/lgtm

PR #983, opened 2024-06-19
Replace occurrences of `+` in operating system labels for MachineClasses with `_`

### LINKED ISSUE #982: Labels created from machine images for MachineClasses might contain invalid characters and fail to deploy

/area quality
/kind bug
/platform aws

**What happened**:

CloudProfiles might contain machine images whose version number is something like `1.2.3-prerelease+build456` which is valid semver. However, once the version number contains the `+build456` build identifier, the provider extension fails to deploy the MachineClass. 

[Here](https://github.com/gardener/gardener-extension-provider-aws/blob/df388be6af471ce4897fdefc4c98a6eba8947e8e/charts/internal/machineclass/templates/machineclass.yaml#L9-L11), the MachineClass labels are taken straight from the [OS image](https://github.com/gardener/gardener-extension-provider-aws/blob/df388be6af471ce4897fdefc4c98a6eba8947e8e/pkg/controller/worker/machines.go#L205-L208) and thus, one label in this example would be `operatingSystemVersion: 1.2.3-prerelease+build456` - however, a `+` is not a permissible character to appear in a label.

MachineClasses can be deployed successfully even for OS images that contain `+build` build identifiers in their version number.

Insert a machine image version that has a `+build` build identifier in is version number into a CloudProfile and try creating a Shoot with this machine image. Deploying the MachineClass will ultimately fail with

```
Invalid value: "1.2.3+build456": a valid label must be an empty string or consist of alphanumeric characters, ''-'', ''_'' or ''.'', and must start and end with an alphanumeric character (e.g. ''MyValue'',  or ''my_value'',  or ''12345'', regex used for validation is ''(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?''), could not apply object of kind "MachineClass" "shoot--myproject--mycluster/shoot--myproject--mycluster-myworkerpool-f3216": MachineClass.machine.sapcloud.io "shoot--myproject--mycluster/shoot--myproject--mycluster-myworkerpool-f3216"
```

This issue seems to apply to the other provider extensions as well.

**Environment**:

- Gardener version (if relevant): v1.97.0
- Extension version: v1.55.1
- Kubernetes version (use `kubectl version`): -
- Cloud provider or hardware configuration: AWS

MrBatschner:

/area quality
/kind bug
/platform aws

Replaces occurrences of the `+` character in values to the MachineClass label `operatingSystemVersion` with `_` as `+` is not a permitted character in Kubernetes labels.

Fixes #982 

```other operator
A problem with deploying MachineClasses that reference an operating system image whose version contains a `+` character was fixed. 
```

### DIFF
--- a/pkg/controller/worker/machines.go
+++ b/pkg/controller/worker/machines.go
@@ -10,6 +10,7 @@ import (
 	"path/filepath"
 	"sort"
 	"strconv"
+	"strings"
 
 	"github.com/Masterminds/semver/v3"
 	"github.com/gardener/gardener/extensions/pkg/controller"
@@ -204,7 +205,7 @@ func (w *workerDelegate) generateMachineConfig(ctx context.Context) error {
 			if pool.MachineImage.Name != "" && pool.MachineImage.Version != "" {
 				machineClassSpec["operatingSystem"] = map[string]interface{}{
 					"operatingSystemName":    pool.MachineImage.Name,
-					"operatingSystemVersion": pool.MachineImage.Version,
+					"operatingSystemVersion": strings.Replace(pool.MachineImage.Version, "+", "_", -1),
 				}
 			}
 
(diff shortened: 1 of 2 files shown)


AndreasBurger (approved):
lgtm

PR #984, opened 2024-06-21
Add option to deploy aws ipam controller

DockToFuture:

/area networking
/kind enhancement
/platform aws

The aws-ipam-controller is required to support prefix delegation for IPv6 and dual-stack shoot clusters. With this pull request the aws-ipam-controller is deployed per default into the controlplane of the shoot cluster. It runs for IPv6 and dual stack shoot clusters and is scaled down to zero for IPv4 shoot clusters. Once active it disables the `allocate-node-cidrs` field in the kube-controller-manager. For cloud-controller-manager the `allocate-node-cidrs` field is already disabled.

Fixes #

**Release note**:

```other operator
The aws-ipam-controller is enabled for IPv6 and dual stack shoot clusters. It replaces the IPAM of the cloud-controller-manager and kube-controller-manager and supports additionally prefix delegation in case of IPv6 or dual stack. 
```


kon-angelo:
@DockToFuture do you think that this should be exposed on a shoot level ? Would it make sense as a seed featuregate instead ? I don't see value in the user manually fiddling with the IPAM - rather it is something that we know from the IP family support and we can configure it ourselves.

pkg/controller/controlplane/valuesprovider.go
```diff
@@ -587,6 +625,74 @@ func getCRCChartValues(
...
+			nodeCidrMaskSizeIPv4 = *cluster.Shoot.Spec.Kubernetes.KubeControllerManager.NodeCIDRMaskSize
+		}
+		if cluster.Shoot.Spec.Networking != nil &&
+			len(cluster.Shoot.Spec.Networking.IPFamilies) == 1 &&
+			cluster.Shoot.Spec.Networking.IPFamilies[0] == v1beta1.IPFamilyIPv6 {
+			nodeCidrMaskSizeIPv6 = *cluster.Shoot.Spec.Kubernetes.KubeControllerManager.NodeCIDRMaskSize
+		}
+	}
+
+	podNetwork := "192.168.0.0/16"
```
kon-angelo, line 662: Can you maybe explain where this value is coming from and why it is hardcoded here ? Shouldn't we return an error if it's missing ?

pkg/controller/controlplane/valuesprovider.go
```diff
@@ -587,6 +625,74 @@ func getCRCChartValues(
...
+		},
+		"podLabels": map[string]interface{}{
+			v1beta1constants.LabelPodMaintenanceRestart: "true",
+		},
+		"region":               cp.Spec.Region,
+		"mode":                 mode,
+		"nodeCIDRMaskSizeIPv4": nodeCidrMaskSizeIPv4,
+		"nodeCIDRMaskSizeIPv6": nodeCidrMaskSizeIPv6,
+	}
+	enabled := (cluster.Shoot.Spec.Networking != nil &&
```
kon-angelo, line 683: 1. I feel that this condition is used in many places already, hence it would be nice if we had an utility function for it already.
2. Maybe a clarification for the enablement and the FeatureGate. My understanding is the following: we need the IPAM controller no matter what for dual stack and ipv6 and hiding behind the feature gate is not changing much. I though the FG was there to enable ipv4 using the new controller for all shoots (including the ipv4 only shoots)

pkg/controller/controlplane/valuesprovider.go
DockToFuture: The ipam controller requires a value for the podNetwork. In case of IPv6 you do not want to specify it and take a default value. I took the same value as specified in the ccm for the podNetwork, see: https://github.com/gardener/gardener-extension-provider-aws/blob/180b867ff68470bae677d1d4c2fe097c6949e043/charts/internal/seed-controlplane/charts/cloud-controller-manager/values.yaml#L4

pkg/controller/controlplane/valuesprovider.go
DockToFuture: Done and added additional test case to test all feature gate scenarios.

pkg/controller/controlplane/valuesprovider_test.go
```diff
@@ -102,6 +104,7 @@ var _ = Describe("ValuesProvider", func() {
 	)
 
 	BeforeEach(func() {
+		format.MaxLength = 0
```
AndreasBurger, line 107: 👍

AndreasBurger (approved):
lgtm

pkg/controller/controlplane/valuesprovider.go
```diff
@@ -587,6 +623,64 @@ func getCRCChartValues(
...
+		}
+		if isIPv6(cluster) {
+			nodeCidrMaskSizeIPv6 = *cluster.Shoot.Spec.Kubernetes.KubeControllerManager.NodeCIDRMaskSize
+		}
+	}
+
+	// podNetwork is default IPv4 range for aws-ipam-controller in case IPv6 is used.
+	podNetwork := "192.168.0.0/16"
+	if slices.Contains(cluster.Shoot.Spec.Networking.IPFamilies, v1beta1.IPFamilyIPv4) {
+		podNetwork = extensionscontroller.GetPodNetwork(cluster)
```
kon-angelo, line 656: @DockToFuture you should rebase because with the latest vendored g/g, I believe that this method returns a slice.

kon-angelo (dismissed):

AndreasBurger (approved):

hebelsan (commented):
/lgtm

DockToFuture:
Can we get this change in?

charts/internal/seed-controlplane/charts/cloud-controller-manager/templates/deployment.yaml
```diff
@@ -39,7 +39,7 @@ spec:
         imagePullPolicy: IfNotPresent
         command:
         - /bin/aws-cloud-controller-manager
-        - --allocate-node-cidrs=true
+        - --allocate-node-cidrs=false
```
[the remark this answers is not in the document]
kon-angelo: So long story short, the IPAM in Gardener is usually done by the KCM. The flag `allocate-node-cidrs` really exists, but I have not been able to make the CCM handle IPAM regardless of the value of this flag. I am not sure if I am missing other options, but AFAIK, the IPAM was never done by the CCM. Hence, I just set it to its "real" value.

The other change touching the webhook/ensurer.go is the real IPAM change.

hebelsan (dismissed):

pkg/webhook/controlplane/ensurer.go
```diff
@@ -234,6 +241,11 @@ func ensureKubeControllerManagerCommandLineArgs(c *corev1.Container, k8sVersion
 
 	c.Command = extensionswebhook.EnsureNoStringWithPrefix(c.Command, "--cloud-config=")
 	c.Command = extensionswebhook.EnsureNoStringWithPrefix(c.Command, "--external-cloud-volume-plugin=")
+
+	// provider-aws is already deploying CCM with --allocate-node-cidrs. We should safely disable IPAM management from KCM.
```
DockToFuture, line 245: provider-aws is already deploying CCM with --allocate-node-cidrs=false.

kon-angelo: The comment was from another solution that did not work so well. Thank you for noticing

AndreasBurger (approved):

hebelsan (approved):

PR #1011, opened 2024-07-18
Extend/infrastructure config

DockToFuture:

/area networking
/kind enhancement
/platform aws

This PR continues the work from https://github.com/gardener/gardener/pull/9998 to allow propagating network ranges from provider extension to Shoot status. It adds fields for the node, pod and service CIDRs to the `InfrastructureConfig`. When defined they are taken into account as well as the fields defined in the networking section of the shoot.yaml file. The PR is a prerequisite to later add the delegated prefixes provided by the AWS infrastructure provider. 

Fixes #

**Release note**:

```other operator
Propagate network ranges from provider extension to Shoot status.
```


pkg/apis/aws/v1alpha1/types_infrastructure.go
```diff
@@ -35,6 +35,16 @@ type InfrastructureConfig struct {
...
+
+	// Pods are the CIDRs of the pod network.
+	// +optional
+	Pods []string `json:"pods,omitempty"`
+	// Nodes are the CIDRs of the node network.
+	// +optional
+	Nodes []string `json:"nodes,omitempty"`
+	// Services are the CIDRs of the service network.
+	// +optional
+	Services []string `json:"services,omitempty"`
```
kon-angelo, line 47: I am sorry but I don't see the connection with the new types introduced here. The PR in g/g adds new infrastructure status that you can populate already with whatever is in the `shoot.spec.networking`.

After the IPv6 PR you just need to propagate ranges that you read/created from the infra reconciliation step and that doesn't require anything from the infrastructureConfig right ?

kon-angelo (changes requested):

DockToFuture:
Not needed any more

PR #1024, opened 2024-08-05
Add ipv6 support.

axel7born:

/area networking
/kind enhancement
/platform aws

The PR adds required infrastructure changes to deploy IPv6 shoots.

Fixes #

**Release note**:

```other operator
Add ipv6 support to deploy IPv6 shoots.
```


pkg/controller/controlplane/valuesprovider.go
```diff
@@ -464,12 +469,18 @@ func getConfigChartValues(
...
 		"vpcID":       infraStatus.VPC.ID,
 		"subnetID":    subnet.ID,
 		"clusterName": cp.Namespace,
 		"zone":        subnet.Zone,
-	}, nil
+	}
+
+	if ipFamilies != nil && sets.New[v1beta1.IPFamily](ipFamilies...).Has(v1beta1.IPFamilyIPv6) {
+		config["nodeIPFamilies"] = "ipv6"
+	}
```
kon-angelo, line 481: Does that mean that we cannot have "dual stack" or maybe better that the ipv6 is a superset of the ipv4 ? (given that here we only care about ipv6 and disregard the potential ipv4 family)

pkg/controller/infrastructure/infraflow/reconcile.go
```diff
@@ -967,22 +1064,67 @@ func (c *FlowContext) deleteNATGateway(zoneName string) flow.TaskFn {
...
+		} else {
+			log.Info("creating...")
+			created, err := c.client.CreateEgressOnlyInternetGateway(ctx, desired)
+			if err != nil {
+				return err
+			}
+			c.state.Set(IdentifierEgressOnlyInternetGateway, created.EgressOnlyInternetGatewayId)
+		}
+	}
+	return nil
```
kon-angelo, line 1095: - reverse the conditional for an early return
- can the ip families change in a running shoot ? because if yes, then maybe a cleanup is also needed here

pkg/controller/infrastructure/infraflow/reconcile.go
```diff
@@ -158,13 +163,20 @@ func (c *FlowContext) ensureManagedVpc(ctx context.Context) error {
...
 	}
-	if c.config.Networks.VPC.CIDR == nil {
+
+	if isIPv4(c.ipFamilies) && c.config.Networks.VPC.CIDR == nil {
 		return fmt.Errorf("missing VPC CIDR")
 	}
-	desired.CidrBlock = *c.config.Networks.VPC.CIDR
+
+	if !isIPv4(c.ipFamilies) && c.config.Networks.VPC.CIDR == nil {
+		desired.CidrBlock = "10.0.0.0/16"
```
kon-angelo, line 175: I think I forgot the outcome of that discussion around this hardcoded IP and it would be nice if here or better in a comment write something about it.

pkg/controller/infrastructure/infraflow/reconcile.go
```diff
@@ -581,35 +666,45 @@ func (c *FlowContext) ensureZones(ctx context.Context) error {
...
 			&awsclient.Subnet{
 				Tags:                        tagsPublic,
 				VpcId:                       c.state.Get(IdentifierVPC),
-				CidrBlock:                   zone.Public,
 				AvailabilityZone:            zone.Name,
-				AssignIpv6AddressOnCreation: ptr.To(false),
+				AssignIpv6AddressOnCreation: ptr.To(isIPv6(c.ipFamilies)),
+				CidrBlock: func(cidr string) string {
+					if cidr == "" {
+						return "10.0.32.0/20"
```
kon-angelo, line 683: also here regarding the "magic IP"

pkg/controller/infrastructure/actuator_reconcile.go
```diff
@@ -395,6 +407,18 @@ func generateTerraformInfraConfig(ctx context.Context, infrastructure *extension
 		dhcpDomainName = fmt.Sprintf("%s.compute.internal", infrastructure.Spec.Region)
 	}
 
+	isIPv4 := true
+	isIPv6 := false
+	if sets.New[v1beta1.IPFamily](ipFamilies...).Has(v1beta1.IPFamilyIPv6) {
```
hebelsan, line 412: Is there a special reason why we use sets, couldn't we just use `slices.Contains()` here?

pkg/controller/controlplane/valuesprovider.go
axel7born: I changed it, also in [cloud-provider-config.yaml](https://github.com/gardener/gardener-extension-provider-aws/blob/6817edffe6109c78f3b3a4ba194d988040f35498/charts/internal/cloud-provider-config/templates/cloud-provider-config.yaml#L15-L21), so that it should also work in the dual-stack case.

pkg/controller/infrastructure/infraflow/reconcile.go
axel7born: I changed the condition here. 
For now ipfamilies are immutable. However, later we will want to migrate ipv4 shoots to dualstack and also back. However, I would add the clean-up code later, when we need it and are able to test it.

pkg/controller/infrastructure/infraflow/reconcile.go
axel7born: Probably, it is better not to automatically assign an IPv4 address here and leave that to the user?

pkg/controller/infrastructure/infraflow/reconcile.go
axel7born: Same as above, it is probably better to let the user configure the CIDR for the public range.

pkg/aws/client/client.go
```diff
@@ -1202,6 +1206,79 @@ func (c *Client) DeleteInternetGateway(ctx context.Context, id string) error {
 	return ignoreNotFound(err)
 }
 
+func (c *Client) CreateEgressOnlyInternetGateway(ctx context.Context, gateway *EgressOnlyInternetGateway) (*EgressOnlyInternetGateway, error) {
```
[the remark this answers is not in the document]
axel7born: I added a comment.

pkg/controller/infrastructure/actuator_reconcile.go
axel7born: No special reason, I changed it to slices.Contains().

pkg/controller/infrastructure/infraflow/reconcile.go
axel7born: Though, I need to adapt the validation.

pkg/apis/aws/validation/infrastructure_test.go
```diff
@@ -165,7 +165,7 @@ var _ = Describe("InfrastructureConfig validation", func() {
 			It("should forbid empty zones", func() {
 				infrastructureConfig.Networks.Zones = nil
 
-				errorList := ValidateInfrastructureConfig(infrastructureConfig, &nodes, &pods, &services)
+				errorList := ValidateInfrastructureConfig(infrastructureConfig, []core.IPFamily{core.IPFamilyIPv4}, &nodes, &pods, &services)
```
hebelsan, line 168: Just a small nit:
Since this array is used so many times in this file, I think it would be nice to create a var like:
`var familyIPv4 = []core.IPFamily{core.IPFamilyIPv4}`

hebelsan (commented):
Just a very small comment, other than that lgtm.

pkg/aws/client/client.go
```diff
@@ -1558,8 +1643,10 @@ func (c *Client) UpdateSubnetAttributes(ctx context.Context, desired, current *S
 		modified = true
 	}
 	privateDnsHostnameTypeOnLaunch := desired.PrivateDnsHostnameTypeOnLaunch
-	if privateDnsHostnameTypeOnLaunch == nil {
+	if privateDnsHostnameTypeOnLaunch == nil && desired.CidrBlock != "" && !(desired.Ipv6Native != nil && *desired.Ipv6Native) {
 		privateDnsHostnameTypeOnLaunch = aws.String(ec2.HostnameTypeIpName)
+	} else {
+		privateDnsHostnameTypeOnLaunch = aws.String(ec2.HostnameTypeResourceName)
```
AndreasBurger, line 1649: This should be identical to
```golang
	privateDnsHostnameTypeOnLaunch := aws.String(ec2.HostnameTypeResourceName)
	if desired.PrivateDnsHostnameTypeOnLaunch == nil && desired.CidrBlock != "" && !(desired.Ipv6Native != nil && *desired.Ipv6Native) {
		privateDnsHostnameTypeOnLaunch = aws.String(ec2.HostnameTypeIpName)
	}
```
which reads strange to me. If the desired DnsHostnameType is unset, we set it to IpName if IPv4 is enabled, I think. All other cases will just lead to using the resource name type.
Are there no good reasons to honour a desired value of `ec2.HostnameTypeIpName`, because that will not be possible with the change?

pkg/controller/infrastructure/infraflow/reconcile.go
```diff
@@ -1420,17 +1553,17 @@ func (c *FlowContext) getSubnetKey(item *awsclient.Subnet) (zoneName, subnetKey
 			for _, key := range []string{IdentifierZoneSubnetWorkers, IdentifierZoneSubnetPublic, IdentifierZoneSubnetPrivate} {
 				switch key {
 				case IdentifierZoneSubnetWorkers:
-					if value == helper.GetSuffixSubnetWorkers() {
+					if value == fmt.Sprintf("%s-%s", c.namespace, helper.GetSuffixSubnetWorkers()) {
```
kon-angelo, line 1556: can this value be changed freely ? What happens to existing clusters that have already a state and the key will change ?

kon-angelo (changes requested):

pkg/controller/infrastructure/templates/main.tpl.tf
```diff
@@ -144,19 +163,33 @@ resource "aws_security_group_rule" "nodes_egress_all" {
...
+
+  {{ if $.isIPv6 }}
+  ipv6_native = true
+  assign_ipv6_address_on_creation = true
+  ipv6_cidr_block = "${cidrsubnet({{ $.vpc.ipv6CidrBlock }}, 8, (({{ $index }} * 3)))}"
+  enable_resource_name_dns_aaaa_record_on_launch = true
+  {{ end }}
+  {{- if $.dualStack.enabled }}
   ipv6_cidr_block = "${cidrsubnet({{ $.vpc.ipv6CidrBlock }}, 8, (({{ $index }} * 3)))}"
   assign_ipv6_address_on_creation = false
```
kon-angelo, line 191: don't this settings collide with the IPv6 settings ?

kon-angelo (changes requested):

pkg/apis/aws/validation/infrastructure.go
```diff
@@ -109,18 +110,20 @@ func ValidateInfrastructureConfig(infra *apisaws.InfrastructureConfig, nodesCIDR
 	for i, zone := range infra.Networks.Zones {
 		zonePath := networksPath.Child("zones").Index(i)
 
-		internalPath := zonePath.Child("internal")
-		cidrs = append(cidrs, cidrvalidation.NewCIDR(zone.Internal, internalPath))
-		allErrs = append(allErrs, cidrvalidation.ValidateCIDRIsCanonical(internalPath, zone.Internal)...)
-
 		publicPath := zonePath.Child("public")
 		cidrs = append(cidrs, cidrvalidation.NewCIDR(zone.Public, publicPath))
 		allErrs = append(allErrs, cidrvalidation.ValidateCIDRIsCanonical(publicPath, zone.Public)...)
```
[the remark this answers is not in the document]
axel7born: The public subnet needs an IPv4 range for load balancers. This range needs to be configured by the user. Using a magic constant could be problematic, especially for existing VPCs. This IPv4 range should be validated.

pkg/controller/infrastructure/infraflow/reconcile.go
axel7born: I'm wondering how this should have worked before. The tag is set [here](https://github.com/gardener/gardener-extension-provider-aws/blob/ef563c1cd252449234644121e9c301d4c60712a1/pkg/controller/infrastructure/infraflow/context.go#L315) and could have never matched before. Or am I missing something.

pkg/controller/infrastructure/infraflow/context.go
```diff
@@ -155,6 +160,7 @@ func NewFlowContext(opts Opts) (*FlowContext, error) {
 		infra:         opts.Infrastructure,
 		client:        opts.AwsClient,
 		runtimeClient: opts.RuntimeClient,
+		ipFamilies:    opts.IPFamilies,
```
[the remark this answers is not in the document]
axel7born: I get it from `c.Shoot.Spec.Networking.IPFamilies`, which I don't have here. Or am I missing something?

pkg/apis/aws/validation/infrastructure_test.go
axel7born: done

pkg/aws/client/client.go
axel7born: Could catch! Fixed.

pkg/apis/aws/validation/infrastructure.go
```diff
@@ -109,18 +110,20 @@ func ValidateInfrastructureConfig(infra *apisaws.InfrastructureConfig, nodesCIDR
...
 
-		workerPath := zonePath.Child("workers")
-		cidrs = append(cidrs, cidrvalidation.NewCIDR(zone.Workers, workerPath))
-		allErrs = append(allErrs, cidrvalidation.ValidateCIDRIsCanonical(workerPath, zone.Workers)...)
-		workerCIDRs = append(workerCIDRs, cidrvalidation.NewCIDR(zone.Workers, workerPath))
+		if ipFamilies != nil && slices.Contains(ipFamilies, core.IPFamilyIPv4) {
```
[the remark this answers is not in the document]
axel7born: done

pkg/apis/aws/validation/shoot.go
```diff
@@ -20,7 +21,7 @@ import (
 func ValidateNetworking(networking *core.Networking, fldPath *field.Path) field.ErrorList {
 	allErrs := field.ErrorList{}
 
-	if networking.Nodes == nil {
+	if networking.Nodes == nil && slices.Contains(networking.IPFamilies, core.IPFamilyIPv4) {
```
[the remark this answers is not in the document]
axel7born: done

pkg/controller/infrastructure/infraflow/reconcile.go
```diff
@@ -975,22 +1062,68 @@ func (c *FlowContext) deleteNATGateway(zoneName string) flow.TaskFn {
...
+
 func (c *FlowContext) ensurePrivateRoutingTable(zoneName string) flow.TaskFn {
 	return func(ctx context.Context) error {
 		log := LogFromContext(ctx)
 		child := c.getSubnetZoneChild(zoneName)
 		id := child.Get(IdentifierZoneRouteTable)
-		cidrBlock := "0.0.0.0/0"
+
+		allIPv4 := "0.0.0.0/0"
+		allIPv6 := "::/0"
```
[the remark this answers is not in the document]
axel7born: done

pkg/controller/infrastructure/infraflow/reconcile.go
```diff
@@ -975,22 +1062,68 @@ func (c *FlowContext) deleteNATGateway(zoneName string) flow.TaskFn {
...
+
+		routes := []*awsclient.Route{}
+
+		routes = append(routes, &awsclient.Route{
+			DestinationCidrBlock: ptr.To(allIPv4),
+			NatGatewayId:         child.Get(IdentifierZoneNATGateway),
+		})
+
+		if c.ipFamilies == nil || slices.Contains(c.ipFamilies, v1beta1.IPFamilyIPv6) {
```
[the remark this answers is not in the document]
axel7born: done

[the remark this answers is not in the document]
axel7born: done

pkg/controller/infrastructure/templates/main.tpl.tf
axel7born: Added the missing `else`.

axel7born:
I think I addressed all comments. Could you take a look again, please?

pkg/controller/infrastructure/infraflow/reconcile.go
```diff
@@ -236,6 +247,14 @@ func (c *FlowContext) ensureExistingVpc(ctx context.Context) error {
 		return fmt.Errorf("Internet Gateway not found for VPC %s", vpcID)
 	}
 	c.state.Set(IdentifierInternetGateway, gw.InternetGatewayId)
+
+	if slices.Contains(c.ipFamilies, v1beta1.IPFamilyIPv6) {
+		eogw, err := c.client.FindEgressOnlyInternetGatewayByVPC(ctx, vpcID)
+		if err != nil || eogw == nil {
+			return fmt.Errorf("Egress-Only Internet Gateway not found for VPC %s", vpcID)
+		}
+		c.state.Set(IdentifierEgressOnlyInternetGateway, eogw.EgressOnlyInternetGatewayId)
```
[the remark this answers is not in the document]
kon-angelo: You can fetch it with:
```
	cluster, err := extensionscontroller.GetCluster(ctx, r.client, infrastructure.Namespace)
	if err != nil {
		return reconcile.Result{}, err
	}

```

PR #1068, opened 2024-09-24
Introduce `InfrastructureConfigFromCluster` and use it instead of API call

Kostov6:

/area quality
/kind bug
/platform aws

This PR introduces `InfrastructureConfigFromCluster` that extracts the infrastructureConfig from a cluster resource and uses it in `pkg/webhook/controlplane/ensurer.go` instead of an API call.

This fixes a issue where cluster force deletion may fail

**Release note**:

```bugfix operator
An issue causing Shoot force deletion to fail because the control plane webhook failing to get the Infrastructure object from the Seed cluster is now fixed. The control plane webhook now reads the infrastructureConfig from the Shoot spec in the Cluster resource. Previously, it was fetching the Infrastructure object from the Seed cluster and was reading the infrastructureConfig from there.
```

### DIFF
--- a/pkg/apis/aws/helper/scheme.go
+++ b/pkg/apis/aws/helper/scheme.go
@@ -28,6 +28,7 @@ var (
 func init() {
 	Scheme = runtime.NewScheme()
 	utilruntime.Must(install.AddToScheme(Scheme))
+	utilruntime.Must(extensionsv1alpha1.AddToScheme(Scheme))
 
 	decoder = serializer.NewCodecFactory(Scheme, serializer.EnableStrict).UniversalDecoder()
 }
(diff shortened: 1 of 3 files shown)


pkg/apis/aws/helper/scheme.go
```diff
@@ -44,6 +45,18 @@ func CloudProfileConfigFromCluster(cluster *controller.Cluster) (*api.CloudProfi
 	return cloudProfileConfig, nil
 }
 
+// InfrastructureFromCluster decodes the infrastructure for a shoot cluster
+func InfrastructureFromCluster(cluster *controller.Cluster) (*extensionsv1alpha1.Infrastructure, error) {
+	var infra *extensionsv1alpha1.Infrastructure
+	if cluster != nil && cluster.Shoot != nil && cluster.Shoot.Spec.Provider.InfrastructureConfig != nil && cluster.Shoot.Spec.Provider.InfrastructureConfig.Raw != nil {
+		infra = &extensionsv1alpha1.Infrastructure{}
+		if _, _, err := decoder.Decode(cluster.Shoot.Spec.Provider.InfrastructureConfig.Raw, nil, infra); err != nil {
```
ialidzhikov, line 53: How does this work and did you test this at all? It should not be possible to decode `shoot.spec.provider.infrastructureConfig` to `*extensionsv1alpha1.Infrastructure`. `shoot.spec.provider.infrastructureConfig` should be decoded to `*api.InfrastructureConfig` instead.

ialidzhikov (changes requested):

ialidzhikov (dismissed):
/lgtm

kon-angelo (approved):
/lgtm

ialidzhikov:
@Kostov6 when you rebased the PR, you reverted the fixes for https://github.com/gardener/gardener-extension-provider-aws/pull/1068#discussion_r1776785054 that I pushed 🙈 . See https://github.com/gardener/gardener-extension-provider-aws/pull/1086

PR #1070, opened 2024-09-27
Remove min/maxAllowed

hebelsan:

/area control-plane
/kind enhancement
/platform aws

Remove VPA (min|max)Allowed settings.

minAllowed may cause wasted resources while maxAllowed may cause under-reservation and hence noisy neighbor issues.

Fixes #

**Release note**:

```other operator
The provider-aws extension no longer configures min/maxAllowed in any managed VPA resource.
```


charts/gardener-extension-admission-aws/charts/runtime/templates/vpa.yaml
```diff
@@ -9,8 +9,7 @@ spec:
   resourcePolicy:
     containerPolicies:
     - containerName: '*'
-      minAllowed:
-        memory: {{ required ".Values.global.vpa.resourcePolicy.minAllowed.memory is required" .Values.global.vpa.resourcePolicy.minAllowed.memory }}
+      controlledValues: RequestsOnly
```
AndreasBurger, line 12: Empty sets are falsy. I am not sure what the actual desired outcome is (@kon-angelo) but with this we get the blanket policy when we configure _anything_ in `Values.global.vpa.resourcePolicy` and nothing otherwise.
Did not catch that on the original PR :/.

hebelsan: Agree, we probably want to remove all vpa configs from values.yaml and also remove the check `{{- if .Values.global.vpa.resourcePolicy }}` in vpa.yaml.

kon-angelo: I agree we need to remove the conditional

charts/gardener-extension-admission-aws/values.yaml
```diff
@@ -14,7 +14,7 @@ global:
   healthPort: 8081
   vpa:
     enabled: true
-    resourcePolicy:
+#    resourcePolicy:
 #      minAllowed:
 #        memory: 64Mi
```
kon-angelo, line 19: Either remove the block or let it all be a chart value

charts/gardener-extension-provider-aws/values.yaml
```diff
@@ -14,7 +14,7 @@ maxSurge: 50%
 resources: {}
 vpa:
   enabled: true
-  resourcePolicy:
+#  resourcePolicy:
 #    minAllowed:
 #      memory: 64Mi
```
kon-angelo, line 19: same.

kon-angelo (commented):
/lgtm

kon-angelo (approved):
/lgtm

PR #1072, opened 2024-09-30
Update cloud-provider images

hebelsan:

/area control-plane
/kind enhancement
/platform aws

Update cloud provider images (including CSI driver images)

Fixes #

**Release note**:

```other operator
Update cloud-provider images including CSI driver images
```


pkg/controller/worker/machines.go
```diff
@@ -234,7 +238,7 @@ func (w *workerDelegate) generateMachineConfig(ctx context.Context) error {
 				MaxUnavailable: worker.DistributePositiveIntOrPercent(zoneIdx, pool.MaxUnavailable, zoneLen, pool.Minimum),
 				// TODO: remove the csi topology label when AWS CSI driver stops using the aws csi topology key - https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/899
 				// add aws csi driver topology label if it's not specified
-				Labels:                       utils.MergeStringMaps(pool.Labels, map[string]string{awsCSIDriverTopologyKey: zone}),
+				Labels:                       utils.MergeStringMaps(pool.Labels, map[string]string{AwsCSIDriverTopologyKey: zone}),
```
kon-angelo, line 241: Can you remove the comment ? I don't think this is applicable, although we have to use a topology label ourselves still.

pkg/controller/worker/machines.go
```diff
@@ -32,6 +32,11 @@ import (
 	awsapihelper "github.com/gardener/gardener-extension-provider-aws/pkg/apis/aws/helper"
 )
 
+const (
+	// AwsCSIDriverTopologyKey describes the topology key used by the AWS CSI driver
+	AwsCSIDriverTopologyKey = "topology.kubernetes.io/zone"
```
kon-angelo, line 37: Can you maybe add a comment with the issue/PR for the upstream change ? I think for the near future, there may be questions about this.

Also I would remove `Aws` prefix

imagevector/images.yaml
```diff
@@ -223,7 +223,7 @@ images:
 - name: csi-snapshot-validation-webhook
   sourceRepository: github.com/kubernetes-csi/external-snapshotter
   repository: registry.k8s.io/sig-storage/snapshot-validation-webhook
-  tag: "v7.0.1"
+  tag: "v8.1.0"
```
kon-angelo, line 226: Did you test the snapshot validation with these changes ? I believe that there are some RBAC changes that we did in the other upgrade PRs (e.g. https://github.com/gardener/gardener-extension-provider-azure/pull/974)

kon-angelo (changes requested):
Did you try to make an "upgrade" test to check if the new EBS CSI version created any compatibility issue ? From the upgrade notes that shouldn't be the case, but better be safe

imagevector/images.yaml
hebelsan: Done

hebelsan:
Locally tested PV upgrade and snapshot creation.

charts/internal/shoot-system-components/charts/csi-driver-node/templates/clusterrole-csi-provisioner.yaml
```diff
@@ -6,10 +6,10 @@ metadata:
 rules:
 - apiGroups: [""]
   resources: ["persistentvolumes"]
-  verbs: ["get", "list", "watch", "create", "delete"]
+  verbs: ["get", "list", "watch", "create", "patch", "delete"]
```
kon-angelo, line 9: Can you check the other repos too for this change ?

kon-angelo (approved):
/lgtm

kon-angelo (approved):

PR #1080, opened 2024-10-07
SC Gardener aws-ebs-csi-driver image update

scs-gardener-team-gitlab:

/area storage
/area control-plane
/area delivery
/kind bug
/platform aws

SC Gardener landscapes require the aws csi-driver to run in legacyXFS mode. To do this the parameter needs to be configurable inside of the values.yaml file as well as added to the daemonset if aforementioned parameter is set to `True`.

Fixes #

Test manually. Paused the corresponding managed resource object, edited the csi-driver pod and its image for one of our cluster to use the `1.35.0` version. Then, edited the pod spec to include the extra legacyXFS patameter. The pod is running fine with no issues in legacyXFS mode.

**Release note**:

```other operator
The following images are updated:
registry.k8s.io/provider-aws/aws-ebs-csi-driver: v1.29.0 -> v1.35.0
```

### DIFF
--- a/charts/internal/shoot-system-components/charts/csi-driver-node/templates/daemonset.yaml
+++ b/charts/internal/shoot-system-components/charts/csi-driver-node/templates/daemonset.yaml
@@ -49,7 +49,7 @@ spec:
         - --volume-attach-limit={{ .Values.driver.volumeAttachLimit }}
         {{- end }}
         {{- if .Values.driver.legacyXFS }}
-        - --legacyXFS={{ .Values.driver.legacyXFS }}
+        - --legacy-xfs={{ .Values.driver.legacyXFS }}
         {{- end }}
         - --logtostderr
         - --v=3
--- a/charts/internal/shoot-system-components/charts/csi-driver-node/templates/daemonset.yaml
+++ b/charts/internal/shoot-system-components/charts/csi-driver-node/templates/daemonset.yaml
@@ -48,6 +48,9 @@ spec:
         {{- if .Values.driver.volumeAttachLimit }}
         - --volume-attach-limit={{ .Values.driver.volumeAttachLimit }}
         {{- end }}
+        {{- if .Values.driver.legacyXFS }}
+        - --legacyXFS=true
+        {{- end }}
         - --logtostderr
         - --v=3
         - --vmodule=node_*=4,mount*=4,driver=4,controller=4
--- a/charts/internal/shoot-system-components/charts/csi-driver-node/templates/daemonset.yaml
+++ b/charts/internal/shoot-system-components/charts/csi-driver-node/templates/daemonset.yaml
@@ -49,7 +49,7 @@ spec:
         - --volume-attach-limit={{ .Values.driver.volumeAttachLimit }}
         {{- end }}
         {{- if .Values.driver.legacyXFS }}
-        - --legacyXFS={{ .Values.driver.volumeAttachLimit }}
+        - --legacyXFS={{ .Values.driver.legacyXFS }}
         {{- end }}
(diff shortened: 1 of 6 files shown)


scs-gardener-team-gitlab:
Additional Links:
- PR on the AWS EBS CSI driver repo: https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/2121
- Github issue that brought it to AWS's attention: https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/2110

kon-angelo:
The PR does not really enable the XFS option though does it ? From a user's perspective, how can he enable the new Helm chart option if it's not exposed via the shoot spec (e.g. controlplaneConfig)

kon-angelo (changes requested):
See comment above

hebelsan:
lgtm. Let's merge this [PR](https://github.com/gardener/gardener-extension-provider-aws/pull/1072/files) first though.

aabond:
Hi @kon-angelo and @hebelsan ,

Are we ready to proceed with merging this PR?

pkg/controller/controlplane/valuesprovider.go
```diff
@@ -720,6 +720,12 @@ func getControlPlaneShootChartValues(
 		}
 	}
 
+	if value, ok := cluster.Shoot.Annotations[aws.legacyXFS]; ok {
+		csiDriverNodeValues["driver"] = map[string]interface{}{
+			"legacyXFS": value,
+		}
+	}
+
```
kon-angelo, line 728: With the code as it is, you would overwrite the value of `VolumeAttachLimit` in the `driver` key of the map. You should only create the map if it's not there.

kon-angelo (changes requested):

charts/internal/shoot-system-components/charts/csi-driver-node/templates/daemonset.yaml
```diff
@@ -48,6 +48,9 @@ spec:
         {{- if .Values.driver.volumeAttachLimit }}
         - --volume-attach-limit={{ .Values.driver.volumeAttachLimit }}
         {{- end }}
+        {{- if .Values.driver.legacyXFS }}
+        - --legacyXFS={{ .Values.driver.legacyXFS }}
```
kon-angelo, line 52: Also this flag is not even correct. https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/7fd319a43ee6fe339517051226c506304d3f5f94/pkg/driver/options.go#L120 is the proper format.

kon-angelo (changes requested):

AndreasBurger (approved):

docs/usage/usage.md
```diff
@@ -603,6 +603,10 @@ The Kubernetes scheduler allows configurable limit for the number of volumes tha
...
 
+### Other CSI options
+The newer versions of EBS CSI driver are not readily compatible with the use of XFS volumes on nodes using a kernel version <= 5.4.
+A workaround was added that enables the use of a "legacy XFS" mode that introduces a backwards compatible volume formating for the older kernels. You can enable this option for your shoot by using by annotating it with `aws.provider.extensions.gardener.cloud/legacy-xfs=true`.
```
hebelsan, line 608:
```suggestion
A workaround was added that enables the use of a "legacy XFS" mode that introduces a backwards compatible volume formating for the older kernels. You can enable this option for your shoot by annotating it with `aws.provider.extensions.gardener.cloud/legacy-xfs=true`.
```

AndreasBurger (approved):

PR #1083, opened 2024-10-11
Fix routing table association deletion

hebelsan:

/area quality
/kind bug
/platform aws

Fix cleanup of routing table associations.
If no routing table association is found, we should also check in the routing table retrieved by the client.

Fixes #

**Release note**:

```bugfix user
Fix cleanup of routing table associations
```


kon-angelo (commented):
/lgtm

kon-angelo (approved):

PR #1095, opened 2024-10-23
Enforce restriction with regards to IPv6 support

ScheererJ:

/area networking
/area robustness
/kind enhancement
/platform aws

Enforce restriction with regards to IPv6 support.

Fixes #

This check replace https://github.com/gardener/gardener/pull/10716.

**Release note**:

```other operator
Added validation to allow only IPv6-only shoot clusters, but not dual-stack as it is not supported, yet.
```

### DIFF
--- a/pkg/admission/validator/shoot_test.go
+++ b/pkg/admission/validator/shoot_test.go
@@ -312,6 +312,15 @@ var _ = Describe("Shoot validator", func() {
 				}))))
 			})
 
+			It("should allow with IPv6-only networking", func() {
+				c.EXPECT().Get(ctx, cloudProfileKey, &gardencorev1beta1.CloudProfile{}).SetArg(2, *cloudProfile)
+
+				shoot.Spec.Networking.IPFamilies = []core.IPFamily{core.IPFamilyIPv6}
+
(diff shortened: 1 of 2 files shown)


pkg/admission/validator/shoot_test.go
```diff
@@ -300,11 +300,15 @@ var _ = Describe("Shoot validator", func() {
 				c.EXPECT().Get(ctx, cloudProfileKey, &gardencorev1beta1.CloudProfile{}).SetArg(2, *cloudProfile)
 
 				shoot.Spec.Networking.Nodes = nil
+				shoot.Spec.Networking.IPFamilies = []core.IPFamily{core.IPFamilyIPv4, core.IPFamilyIPv6}
```
timebertt, line 303: Can you add a positive test case for allowing single-stack configurations?

ScheererJ: Added single-stack IPv6-only test in https://github.com/gardener/gardener-extension-provider-aws/pull/1095/commits/8894dd503369c57173b9112575fef25b6e092268.

PTAL.

timebertt (approved):

PR #1099, opened 2024-10-24
Always enable AWS load balancer controller for clusters with IPv6

ScheererJ:

/area networking
/kind enhancement
/platform aws

Always enable AWS load balancer controller for clusters with IPv6.

AWS cloud controller manager only supports purely IPv4 load balancers. It will not work in IPv6-only environments and does not offer the full functionality in dual-stack environments. Therefore, this change always enabled the fully IPv6-capable AWS load balancer controller in case a shoot cluster with IPv6 is created.

Fixes #

**Release note**:

```other operator
AWS load balancers controller is always enabled for IPv6-only and dual-stack shoot clusters.
```

### DIFF
--- a/pkg/admission/mutator/shoot.go
+++ b/pkg/admission/mutator/shoot.go
@@ -139,6 +139,14 @@ func (s *shoot) Mutate(_ context.Context, newObj, oldObj client.Object) error {
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


DockToFuture (approved):
/lgtm

kon-angelo (approved):
/lgtm

PR #1106, opened 2024-10-29
Fix nil check after findRouteTable

hebelsan:

/area robustness
/kind bug
/platform aws

We should check if routeTable is nil before accessing it's associations.

Fixes #

**Release note**:

```bugfix operator
Fix missing nil check after getting route table
```


kon-angelo (approved):
/lgtm

AndreasBurger (approved):

PR #1108, opened 2024-10-29
Add IPv4 address to internal subnet for IPv6 only shoots.

axel7born:
The purpose of the internal subnet is to deploy internal load balancers. Since a load balancer can only be deployed to a subnet that has an IPv4 range, we need to assign an IPv4 range to the internal subnet.

/area networking
/kind bug
/platform aws

Fixes #

**Release note**:

```other operator
Fixed an issue preventing the deployment of internal load balancers in IPv6-only shoots.
```


kon-angelo (dismissed):
/lgtm

Nit: I would add a comment about why the internal subnet should always be at the very least dual-stack even on ipv6 only case.

Nit2: Maybe rephrase the title/release note to reflect first nit.

kon-angelo (approved):
/lgtm

PR #1111, opened 2024-10-30
Route table duplicate "0.0.0.0/0" rule creation

kon-angelo:

/area control-plane
/kind bug
/platform aws

Fixes #

**Release note**:

```other operator
Fix an issue where the "0.0.0.0/0" route creation would fail if the nat-gateway was previously deleted.
```

### DIFF
--- a/pkg/aws/client/client.go
+++ b/pkg/aws/client/client.go
@@ -2134,6 +2134,7 @@ func IsNotFoundError(err error) bool {
 // IsAlreadyAssociatedError returns true if the given error is a awserr.Error indicating that an AWS resource was already associated.
 func IsAlreadyAssociatedError(err error) bool {
 	if aerr, ok := err.(awserr.Error); ok && aerr.Code() == "Resource.AlreadyAssociated" {
+		aerr.Code()
 		return true
 	}
 	return false
--- a/pkg/aws/client/client.go
+++ b/pkg/aws/client/client.go
@@ -2134,7 +2134,6 @@ func IsNotFoundError(err error) bool {
 // IsAlreadyAssociatedError returns true if the given error is a awserr.Error indicating that an AWS resource was already associated.
 func IsAlreadyAssociatedError(err error) bool {
 	if aerr, ok := err.(awserr.Error); ok && aerr.Code() == "Resource.AlreadyAssociated" {
-		aerr.Code()
 		return true
 	}
 	return false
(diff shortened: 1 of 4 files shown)


pkg/aws/client/updater.go
```diff
@@ -102,54 +103,73 @@ func (u *updater) UpdateSecurityGroup(ctx context.Context, desired, current *Sec
...
-		for _, cr := range current.Routes {
-			if reflect.DeepEqual(cr, dr) {
-				continue outerCreate
+
+		// The current route table may contain routes not created by the infrastructure controller e.g. the aws-custom-route-controller.
+		// These extra routes will be included in the routesToDelete, but they need to be skipped.
+		if !slices.ContainsFunc(desired.Routes, func(r *Route) bool {
+			desiredRouteDestination, err := r.DestinationId()
+			if err != nil {
+				log.Error(err, "failed to find suitable destination for desired route", "route")
```
AndreasBurger, line 148: This will result in a line like `msg="failed to find suitable destination for desired route" error="no route destination found"`, which is a little bit redundant for my taste.
Maybe something more like
```golang
log.Error(err, "error deleting route", "route")
```
Which is not really that much better but does not contain the same information twice.
Also, I think there is an argument missing, don't you mean:
```golang
log.Error(err, "failed to find suitable destination for desired route", "route", route)
```

pkg/aws/client/client.go
```diff
@@ -2134,6 +2134,7 @@ func IsNotFoundError(err error) bool {
 // IsAlreadyAssociatedError returns true if the given error is a awserr.Error indicating that an AWS resource was already associated.
 func IsAlreadyAssociatedError(err error) bool {
 	if aerr, ok := err.(awserr.Error); ok && aerr.Code() == "Resource.AlreadyAssociated" {
+		aerr.Code()
```
hebelsan, line 2137: Is this a debug line?

kon-angelo: nice catch

kon-angelo:
@AndreasBurger @hebelsan PTAL

hebelsan (commented):
lgtm, however we never make use of the bool return value `modified` of the function `UpdateRouteTable`.
Should we therefore drop it?

kon-angelo:
> lgtm, however we never make use of the bool return value `modified` of the function `UpdateRouteTable`. Should we therefore drop it?

The API for the `updater` is locally consistent, so I don't agree necessarily agree. We can consider it in a different change

hebelsan (dismissed):

AndreasBurger (approved):

PR #1119, opened 2024-11-05
fix test cardinality

kon-angelo:

/area control-plane
/kind test
/platform aws

- Make VPC deletion of the test more robust
- Deletion of infra is needed to count a test as passed
- Deletion of infra occurs after each test (and not at the end of the suite)

Fixes #

**Release note**:

```other operator
NONE
```


hebelsan (approved):
/lgtm

PR #1161, opened 2024-12-05
fix: use UDP for upstream DNS by default

domdom82:

/area networking
/kind enhancement
/platform aws

In the past there have been issues with using TCP for upstream DNS requests on AWS. The AWS infrastructure does not handle lots of TCP connections for DNS queries well and the cluster may run into rate throttling leading to "connection timeout" issues during DNS resolution.

To avoid users having to manually set the `spec.SystemComponents.NodeLocalDNS.ForceTCPToUpstreamDNS` property for AWS, this PR sets a default value (false) if the property is not set on the shoot. This will make the shoot use UDP for upstream DNS by default and should prevent the issue from the start.

**Release note**:

```other user
Shoots with NodeLocalDNS enabled will use UDP instead of TCP for upstream DNS queries by default to avoid performance issues on AWS.
```

### DIFF
--- a/pkg/admission/mutator/shoot.go
+++ b/pkg/admission/mutator/shoot.go
@@ -151,6 +151,16 @@ func (s *shoot) Mutate(_ context.Context, newObj, oldObj client.Object) error {
 		Object: controlPlaneConfig,
 	}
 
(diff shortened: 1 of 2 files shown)


AndreasBurger (approved):
lgtm

PR #1167, opened 2024-12-10
set annotations for dual stack lb based on IPFamily of kube-dns service

DockToFuture:

/area networking
/kind bug
/platform aws

Set annotations for aws dual stack loadbalancer based on IPFamily of kube-dns service.

Fixes #

**Release note**:

```other operator
none
```


pkg/webhook/shoot/loadbalancer_service.go
```diff
@@ -8,11 +8,20 @@ import (
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-func (m *mutator) mutateService(_ context.Context, service *corev1.Service) error {
+func (m *mutator) mutateService(ctx context.Context, service *corev1.Service, shootClient client.Client) error {
 	if service.Spec.Type == corev1.ServiceTypeLoadBalancer {
-		for _, v := range service.Spec.IPFamilies {
+		if service.Annotations == nil {
+			service.Annotations = make(map[string]string, 1)
+		}
```
ScheererJ, line 19: Why not use `metav1.SetMetaDataAnnotation(...)` instead of handling the special case here?

pkg/webhook/shoot/loadbalancer_service.go
```diff
@@ -8,11 +8,20 @@ import (
...
+		kubeDNSService := &corev1.Service{}
+		if err := shootClient.Get(ctx, types.NamespacedName{Name: "kube-dns", Namespace: "kube-system"}, kubeDNSService); err != nil {
+			return err
+		}
+		for _, v := range kubeDNSService.Spec.IPFamilies {
```
ScheererJ, line 24: Should we also add a check for internal load balancers? With the current approach we make all load balancers in a dual stack cluster automatically externally visible.

ScheererJ (approved):
/lgtm

hebelsan (approved):
/lgtm

PR #1175, opened 2024-12-17
Set IPv4 CIDR to empty string for IPv6 only shoots.

axel7born:

/area networking
/kind bug
/platform aws

With  https://github.com/gardener/gardener-extension-provider-aws/pull/1158 we use the IPv4 range for the subnet ID if present. 
However, the worker subnet is created as IPv6 only subnet without IPv4 range. During reconciliation, the worker subnet is considered as missing though a subnet with the required IPv6 range exists. 

With this change, we use an empty string as IPv4 CIDRBlock, even if the range is configured in the spec.

Fixes #

**Release note**:

```bug user
Fixes an issue, where reconciliation of IPv6 shoots fails, when an IPv4 range for the worker subnet is configured.
```


kon-angelo (approved):
/lgtm

PR #1205, opened 2025-01-27
[ci:component:github.com/gardener/ecr-credential-provider:v1.31.0->v1.32.0]

gardener-robot-ci-2:


LucaBernstein:
/close
Will be added with k8s v1.32 support.

PR #1239, opened 2025-03-03
[release-v1.60] Fix `StorageClass` deployment

timuthy:

/area delivery
/kind bug
/platform aws

The `StorageClass` must not be considered if deployed by the `gardener-operator` to the runtime cluster. Otherwise different extensions fight against creating the same resource in the cluster.

Fixes #

**Release note**:

```bugfix operator
Fixed an issue that caused deployment issues with the `gardener.cloud-fast` storage class when the extension was deployed by `gardener-operator` in the garden runtime cluster. The deployment of this `StorageClass` object is now only done for AWS seeds.
```


kon-angelo (approved):
/lgtm

PR #1261, opened 2025-03-25
Add AWS S3 bucket level immutability feature.

ishan16696:

/area backup
/kind enhancement
/platform aws

This PR add support for AWS S3 bucket level immutability feature which can be configured in `.spec.providerConfig` of `BackupBucket` resource.

Part of https://github.com/gardener/gardener/issues/10866

For AWS S3: Object versioning is a pre-requisite for enabling object lock on bucket.

**Release note**:

```noteworthy operator
Added support for immutable backup buckets in the AWS provider extension. Operators can configure immutability settings. Please refer to this doc: https://github.com/gardener/gardener-extension-provider-aws/blob/master/docs/usage/usage.md#BackupBucket
```


ishan16696:
**Tests I have performed:**



Test | Before | Tests description | After | Test Outcome
-- | -- | -- | -- | --
Update | Unlocked backup-bucket | Updating the existing backup-bucket with immutability features with mode: `governance` (by updating seed `.spec.backup`) | Object Locked enabled correctly |   Passed ✅ 
Update | Unlocked backup-bucket | Updating the existing backup-bucket with immutability features with mode: `compliance` (by updating seed `.spec.backup`) | Object Locked enabled correctly |   Passed ✅ 
Create | NA | creation of backup-bucket with object lock enabled settings with mode: `compliance/governance` | Object Locked enabled bucket created |  Passed ✅ 
Delete | Object Lock enabled backup bucket present with backup-entry | On shoot deletion, bucket entry should be deleted | backup bucket entry got deleted provided no immutable snapshots should be present in bucket else it will keep retrying |   Passed ✅ 
Delete | Object Lock enabled backup bucket present with no backup-entry | On seed deletion, backup-bucket should be deleted | backup bucket got deleted provided |   Passed ✅ 
Delete | No bucket was present | On seed deletion, deletion should be succeed | as no backup-bucket present, no action is taken |   Passed ✅ 
Delete | backup-bucket was present with orphan backup-entry | On seed deletion, backup-bucket should be deleted provided all object present are mutable | errored |   Failed 🔴 Issue is reported here: https://github.com/gardener/gardener-extension-provider-aws/issues/1288





**Seed admission validation tests:**


Bucket state before test | Test description | Bucket state after test | Outcome | Expectation
-- | -- | -- | -- | --
Object lock enabled bucket with mode: `governance` | try to increase retentionPeriod | retentionPeriod of bucket will increased | accepted |  Passed ✅ 
Object lock enabled bucket with mode: `governance` | try to set retentionPeriod to `0` | retentionPeriod will remain the same | rejected |  Passed ✅ 
Object lock enabled bucket with mode: `governance` |  try to set retentionPeriod to `32h` (not multiple of 24h) | retentionPeriod will remain the same | rejected |  Passed ✅ 
Object lock enabled bucket with mode: `governance` |  try to decrease the retentionPeriod | retentionPeriod will get decrease | accepted | Passed ✅ 
Object lock enabled bucket with mode: `compliance` | try to increase retentionPeriod | retentionPeriod of bucket will increase | accepted |  Passed ✅ 
Object lock enabled bucket with mode: `compliance` | try to set retentionPeriod to `0` | retentionPeriod will remain the same | rejected |  Passed ✅ 
Object lock enabled bucket with mode `compliance` |  try to set retentionPeriod to `32h` (not multiple of 24h) | retentionPeriod will remain the same | rejected |  Passed ✅ 
Object lock enabled bucket with mode `compliance` | try to decrease the retentionPeriod | etentionPeriod will remain the same | rejected |  Passed ✅ 
Object lock enabled bucket with mode `governance` | try to change the mode to `compliance` | mode will get change to `compliance` | accepted |  Passed ✅ 
Object lock enabled bucket with mode `compliance` | try to change the mode to `governance` | mode will remain the same | rejected |  Passed ✅ 
Object lock enabled bucket with retentionType `bucket` | try to change the retentionType to anything | retentionType will remain the same | rejected |  Passed ✅

pkg/aws/client/client.go
```diff
@@ -304,44 +306,29 @@ func (c *Client) GetDHCPOptions(ctx context.Context, vpcID string) (map[string]s
...
-					return nil
-				}
-				return err
+	})
+	if err != nil {
+		var apiErr smithy.APIError
+		if errors.As(err, &apiErr) {
+			if apiErr.ErrorCode() == "NoSuchBucket" {
+				// bucket doesn't exist, no action required
+				return nil
```
seshachalam-yv, line 317: I’m seeing the same NoSuchBucket check in multiple places. Could we pull this into a small helper function

pkg/aws/client/client.go
```diff
@@ -458,20 +451,71 @@ func (c *Client) CreateBucketIfNotExists(ctx context.Context, bucket, region str
...
 
-	_, err = c.S3.PutBucketLifecycleConfiguration(ctx, putBucketLifecycleConfigurationInput)
+	if _, err = c.S3.PutBucketLifecycleConfiguration(ctx, putBucketLifecycleConfigurationInput); err != nil {
+		return err
+	}
+
+	if backupbucketConfig != nil && backupbucketConfig.Immutability != nil {
+		// update the bucket with object lock settings.
+		// note: object versioning is automatically get enabled if bucket is created with object lock.
+		return c.UpdateBucket(ctx, bucket, backupbucketConfig, true)
```
seshachalam-yv, line 461: Right now the `createBucket` not only creates the bucket but also immediately calls UpdateBucket. That mixes two responsibilities in one function. Could we split those into separate steps

ishan16696: > That mixes two responsibilities in one function

nopes, `createBucket` with object lock enabled is actually itself a 2 step process:
you can check this : https://github.com/gardener/etcd-backup-restore/blob/master/docs/usage/enabling_immutable_snapshots.md#create-s3-buckets-with-object-lock-enabled

> Could we split those into separate steps

for reconciliation logic it should be single step process.

seshachalam-yv: I understand that enabling versioning and object lock involves multiple API calls. However, the name `updateBucket` might be too generic and could benefit from a more descriptive alternative that clearly reflects the specific configurations being applied. For instance, names like `configureBucketVersioningAndLock` or `enableBucketVersioningAndObjectLock` would more accurately convey the function's purpose.

seshachalam-yv (changes requested):

pkg/admission/validator/seed.go
```diff
@@ -0,0 +1,162 @@
...
+	// TODO: @ishan16696 to remove these conditions "newConfig == nil || newConfig.Immutability == nil" to not allow disablement of immutability settings.
+	if oldConfig == nil || oldConfig.Immutability == nil || newConfig == nil || newConfig.Immutability == nil {
+		return allErrs
+	}
+
+	// TODO: @ishan16696 uncomment this piece of code, so once disablement of the immutability settings on bucket is not allowed.
+	/*
+		if newConfig == nil || newConfig.Immutability == nil || *newConfig.Immutability == (apisaws.ImmutableConfig{}) {
+			allErrs = append(allErrs, field.Invalid(immutabilityPath, newConfig, "immutability cannot be disabled"))
+			return allErrs
```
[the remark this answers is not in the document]
ishan16696: for now, we are allowing the disabling of default immutability settings but we will not always allow it once this feature become mature ... right ?
We will eventually put this validation to stop the disabling of default immutability

pkg/aws/client/client.go
seshachalam-yv: 👍

pkg/aws/client/client.go
ishan16696: addressed you review comments.

pkg/admission/validator/seed.go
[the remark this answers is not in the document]
ishan16696: addressed you review comments. PTAL

ishan16696:
I have rebased the PR on latest master.

seshachalam-yv (approved):
/lgtm

pkg/apis/aws/types_backupbucket.go
```diff
@@ -0,0 +1,48 @@
...
+// RetentionType defines the level at which immutability properties are applied on objects.
+type RetentionType string
+
+const (
+	// BucketLevelImmutability sets the immutability feature on the bucket level.
+	BucketLevelImmutability RetentionType = "bucket"
+	// ComplianceMode is for "compliance" mode immutability.
+	ComplianceMode string = "compliance"
+	// GovernanceMode mode is of "governance" mode immutability.
+	GovernanceMode string = "governance"
```
kon-angelo, line 20: Maybe here it would fit if these were a separate type

pkg/apis/aws/v1alpha1/types_backupbucket.go
```diff
@@ -0,0 +1,49 @@
...
+// ImmutableConfig represents the immutability configuration for a backup bucket.
+type ImmutableConfig struct {
+	// RetentionType specifies the type of retention for the backup bucket.
+	// Currently allowed value is:
+	// - "bucket": retention policy applies on the entire bucket.
+	RetentionType RetentionType `json:"retentionType"`
+
+	// RetentionPeriod specifies the immutability retention period for the backup bucket.
+	// S3 only supports immutability durations in days or years, therefore this field must be set as multiple of 24h.
+	RetentionPeriod metav1.Duration `json:"retentionPeriod"`
```
kon-angelo, line 44: If it's only days then we could go for an integer approach to make our lives easier, but it would diverge from the other implementations

pkg/controller/backupbucket/actuator.go
```diff
@@ -6,43 +6,162 @@ package backupbucket
...
+	bucketVersioningStatus, err := awsClient.GetBucketVersioningStatus(ctx, bb.Name)
+	if err != nil {
+		var apiErr smithy.APIError
+		if errors.As(err, &apiErr) {
+			switch apiErr.ErrorCode() {
+			case "NoSuchBucket":
+				// bucket doesn't exist, create the bucket with buckupbucket config (if provided)
+				return util.DetermineError(awsClient.CreateBucket(ctx, bb.Name, bb.Spec.Region, backupbucketConfig), helper.KnownCodes)
+			case "PermanentRedirect":
+				return util.DetermineError(fmt.Errorf("bucket exists in different region"), helper.KnownCodes)
```
kon-angelo, line 86: Maybe you can return exactly the error that you want here (if you expect to return a particular error code) instead of going sideways by producing a string that may result in the error you want.

pkg/controller/backupbucket/actuator.go
```diff
@@ -6,43 +6,162 @@ package backupbucket
...
+	}
+
+	bucketVersioningStatus, err := awsClient.GetBucketVersioningStatus(ctx, bb.Name)
+	if err != nil {
+		var apiErr smithy.APIError
+		if errors.As(err, &apiErr) {
+			switch apiErr.ErrorCode() {
+			case "NoSuchBucket":
+				// bucket doesn't exist, create the bucket with buckupbucket config (if provided)
+				return util.DetermineError(awsClient.CreateBucket(ctx, bb.Name, bb.Spec.Region, backupbucketConfig), helper.KnownCodes)
```
kon-angelo, line 84: You could wrap the reconcile function itself with one of these `util.DetermineError` instead. Might make it easier for you.

pkg/aws/client/client.go
```diff
@@ -458,19 +448,66 @@ func (c *Client) CreateBucketIfNotExists(ctx context.Context, bucket, region str
...
 
-	_, err = c.S3.PutBucketLifecycleConfiguration(ctx, putBucketLifecycleConfigurationInput)
+	if _, err = c.S3.PutBucketLifecycleConfiguration(ctx, putBucketLifecycleConfigurationInput); err != nil {
+		return err
+	}
+
+	if backupbucketConfig != nil && backupbucketConfig.Immutability != nil {
+		// update the bucket with object lock settings.
+		// note: object versioning is automatically get enabled if bucket is created with object lock.
+		return c.UpdateBucketConfig(ctx, bucket, backupbucketConfig, true)
```
kon-angelo, line 458: Isn't it better if we don't really hide that call inside the Create function (single responsibility principle). Let's try to make the create function just ensure that there is a bucket and that's it.

If this secondary call fails, you would still need to call a proper update from the actuator anyway, so let's just have a single code path for that update call. WDYT ?

ishan16696: > Isn't it better if we don't really hide that call inside the Create function (single responsibility principle)

`createBucket` with object lock enabled is actually itself a 2 step process:
you can check this : https://github.com/gardener/etcd-backup-restore/blob/master/docs/usage/enabling_immutable_snapshots.md#create-s3-buckets-with-object-lock-enabled

for reconciliation logic isn't it should be single step process as `CreateBucket` is an interface and it shouldn't reveal the implementation details, it should just create the bucket with object lock enabled (if config provided)

pkg/apis/aws/v1alpha1/types_backupbucket.go
ishan16696: yes, on this I don't have any strong opinion

pkg/aws/client/client.go
```diff
@@ -350,6 +334,12 @@ func (c *Client) CreateBucketIfNotExists(ctx context.Context, bucket, region str
 		},
 	}
 
+	// If immutability settings are provided then create bucket with object lock enabled.
+	// Note: while creating a bucket with object lock enabled, object versioning will automatically gets enabled.
+	if backupbucketConfig != nil && backupbucketConfig.Immutability != nil {
```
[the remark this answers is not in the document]
ishan16696: I move this check to reconcile logic from `CreateBucket` but this will make reconciliation logic little complicated, moreover, this check will still be required in `UpdateBucketConfig`.

pkg/aws/client/client.go
ishan16696: > If this secondary call fails, you would still need to call a proper update from the actuator anyway, so let's just have a single code path for that update call. 

we can also do that in 2 step in actuator itself but that will make code will look like this:
```go
	// bucket doesn't exist, create the bucket with buckupbucket config (if provided)
	if err := util.DetermineError(awsClient.CreateBucket(ctx, bb.Name, bb.Spec.Region, backupbucketConfig), helper.KnownCodes); err != nil {
		return err
	} else {
		return util.DetermineError(awsClient.UpdateBucketConfig(ctx, bb.Name, backupbucketConfig, true), helper.KnownCodes)
	}
```
little complicated and make actuator logic hard to maintain ..... wdyt ?

ishan16696: and I include this https://github.com/gardener/gardener-extension-provider-aws/pull/1261#discussion_r2066662774 check as well in reconciliation logic then it become more hard to read and maintain, IMO.

pkg/apis/aws/types_backupbucket.go
ishan16696: addressed. PTAL

pkg/aws/client/client.go
ishan16696: I have made the changes, PTAL

pkg/aws/client/client.go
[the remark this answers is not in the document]
ishan16696: I have made the changes, PTAL

pkg/apis/aws/v1alpha1/types_backupbucket.go
hebelsan: +1 in favour of the integer approach

pkg/aws/client/client.go
hebelsan: For consistency, we should update the `UpdateBucketConfig()` parameters to accept only `s3.PutBucketVersioningInput` and `s3.PutObjectLockConfigurationInput`, rather than the higher-level `backupbucketConfig`, to keep the client as free of business logic as possible.

ishan16696: I have made the required changes. PTAL

kon-angelo (approved):
/lgtm

PR #1198, opened 2025-01-21
Update worker doc

AndreasBurger:
**How to categorize this PR?**

/area documentation
/kind enhancement
/platform aws

**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Release note**:

```other operator

```

PR #1276, opened 2025-04-11
[GEP-31] Adapt extension to support in-place updates

shafeeqes:

/area control-plane delivery open-source scalability
/kind enhancement
/platform aws

This PR adapts the extension to support in-place updates.
- The worker controller needs to set the appropriate update strategy in the machine deployment
- The changes to worker providerConfig or relevant fields(which affect the worker pool hash calculation) should be denied.
- The providerConfig is skipped from worker pool hash calculation.

> [!NOTE]  
> Changing these fields will cause a change in the calculation of the WorkerPool hash. Currently, machine-controller-manager does not provide an UpdateMachine call to update the data volumes or provider config in-place. `gardener-node-agent` cannot update the provider config in-place either. So we disallow changing these fields if the update strategy is in-place.

Part of https://github.com/gardener/gardener/issues/10219

~/hold until gardener/gardener is released with required changes.
and `machine-controller-manager-provider-aws` image is available.~

**Release note**:

```feature user
This extension now supports in-place node updates. Read more about it [here](https://github.com/gardener/gardener/blob/master/docs/proposals/31-inplace-node-update.md).
```


hebelsan (approved):
/lgtm

hebelsan:
@shafeeqes, have you tested that changes of the providerConfig like volume IOPS do happen in place with `strategy: AutoInPlaceUpdate`?

shafeeqes:
Today, only the data strings added in `computeAdditionalHashDataV2` cause a rolling update, and everything else is "in-place". So if it works today, it should work with the new strategy as well. I have not tested it separately.

shafeeqes:
/unhold
Added https://github.com/gardener/gardener-extension-provider-aws/pull/1276/commits/c47398218d067d9b8a55265b05b735fe9a041e6d, see https://github.com/gardener/gardener/blob/fb3f728ba0be9560d64f02e7688222f3a4f51ab2/extensions/pkg/controller/worker/machines.go#L133-L136

kon-angelo (approved):
/lgtm

shafeeqes:
Integration tests in general seems to be broken, see https://github.com/gardener/gardener-extension-provider-aws/pull/1198#issuecomment-2893020584 and not related to this PR.

kon-angelo (approved):
/lgtm

hebelsan (approved):

PR #1313, opened 2025-05-12
Improve Documentation and Admission Checks for IPv6/Dual-Stack and Pod Overlay Network.

ScheererJ:

/area networking
/kind enhancement
/platform aws

Improve Documentation and Admission Checks for IPv6/Dual-Stack and Pod Overlay Network.

Previously, it was possible to directly try to migrate an IPv4-only cluster with pod overlay network. This results in a non-working cluster with a broken pod network. Therefore, we now prevent the dual-stack migration unless the cluster is already working in native routing mode, i.e. unless the overlay network was disabled.

In addition to that, the IPv6 documentation is updated to reflect the fact that dual-stack migration is now possible.

Fixes #

**Release note**:

```other operator
Enabling IPv6 without disabling overlay network is no longer possible.
```

### DIFF
--- a/docs/usage/ipv6.md
+++ b/docs/usage/ipv6.md
@@ -149,13 +149,28 @@ An egress-only internet gateway is required for outbound internet traffic (IPv6)
 
 ### Migration of IPv4-only Shoot Clusters to Dual-Stack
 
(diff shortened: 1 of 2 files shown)


DockToFuture (approved):
/lgtm

kon-angelo (approved):
/lgtm

PR #1315, opened 2025-05-13
Purge unused Machine Controller Manager RBAC resources

georgibaltiev:

/area security
/kind cleanup
/platform aws

After the Machine Controller Manager was moved from the provider extension codebases to the core, the charts that define the component's resources have been deleted, as well as most resources that have been deployed to that point - [PR reference](https://github.com/gardener/gardener-extension-provider-aws/pull/872/files).
However, some ClusterRoles and ClusterRoleBindings that were used for the component still remain on certain seeds.
This PR adds a function to purge those, since they are not in use anymore.

The ClusterRoles|ClusterRoleBindings in question are named in the following way:
```
extensions.gardener.cloud:provider-aws:<shoot-name>:machine-controller-manager
```

N/A

N/A

**Release note**:

```other operator
Obsolete ClusterRoles and ClusterRoleBindings that were leftovers from the machine-controller-manager component are now cleaned up.
```

### DIFF
--- a/cmd/gardener-extension-provider-aws/app/app.go
+++ b/cmd/gardener-extension-provider-aws/app/app.go
@@ -8,6 +8,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"slices"
 	"time"
 
 	druidcorev1alpha1 "github.com/gardener/etcd-druid/api/core/v1alpha1"
@@ -16,6 +17,7 @@ import (
 	"github.com/gardener/gardener/extensions/pkg/controller/controlplane/genericactuator"
 	"github.com/gardener/gardener/extensions/pkg/controller/heartbeat"
 	heartbeatcmd "github.com/gardener/gardener/extensions/pkg/controller/heartbeat/cmd"
+	extensionsworkercontroller "github.com/gardener/gardener/extensions/pkg/controller/worker"
 	"github.com/gardener/gardener/extensions/pkg/util"
 	webhookcmd "github.com/gardener/gardener/extensions/pkg/webhook/cmd"
 	"github.com/gardener/gardener/pkg/client/kubernetes"
@@ -270,11 +272,13 @@ func NewControllerManagerCommand(ctx context.Context) *cobra.Command {
 			}
 
 			// TODO (georgibaltiev): Remove after the release of version 1.64.0
-			log.Info("Adding migration runnables")
-			if err := mgr.Add(manager.RunnableFunc(func(ctx context.Context) error {
-				return purgeMachineControllerManagerRBACResources(ctx, mgr.GetClient(), log)
-			})); err != nil {
-				return fmt.Errorf("error adding migrations: %w", err)
+			if !slices.Contains(controllerSwitches.Disabled, extensionsworkercontroller.ControllerName) {
+				log.Info("Adding migration runnables")
+				if err := mgr.Add(manager.RunnableFunc(func(ctx context.Context) error {
+					return purgeMachineControllerManagerRBACResources(ctx, mgr.GetClient(), log)
+				})); err != nil {
+					return fmt.Errorf("error adding migrations: %w", err)
+				}
 			}
 
 			if err := mgr.Start(ctx); err != nil {
--- a/cmd/gardener-extension-provider-aws/app/app.go
+++ b/cmd/gardener-extension-provider-aws/app/app.go
@@ -269,7 +269,7 @@ func NewControllerManagerCommand(ctx context.Context) *cobra.Command {
 				return fmt.Errorf("could not add ready check for webhook server to manager: %w", err)
 			}
 
-			// TODO (georgibaltiev): Remove after the release of version 1.62.0
+			// TODO (georgibaltiev): Remove after the release of version 1.64.0
 			log.Info("Adding migration runnables")
 			if err := mgr.Add(manager.RunnableFunc(func(ctx context.Context) error {
 				return purgeMachineControllerManagerRBACResources(ctx, mgr.GetClient(), log)
--- a/cmd/gardener-extension-provider-aws/app/app.go
+++ b/cmd/gardener-extension-provider-aws/app/app.go
@@ -269,6 +269,14 @@ func NewControllerManagerCommand(ctx context.Context) *cobra.Command {
 				return fmt.Errorf("could not add ready check for webhook server to manager: %w", err)
 			}
 
+			// TODO (georgibaltiev): remove after the extension's next release
+			log.Info("Adding migration runnables")
+			if err := mgr.Add(manager.RunnableFunc(func(ctx context.Context) error {
+				return purgeMachineControllerManagerRBACResources(ctx, mgr.GetClient())
+			})); err != nil {
+				return fmt.Errorf("error adding migrations: %w", err)
+			}
+
 			if err := mgr.Start(ctx); err != nil {
 				return fmt.Errorf("error running manager: %w", err)
 			}
--- a/cmd/gardener-extension-provider-aws/app/app.go
+++ b/cmd/gardener-extension-provider-aws/app/app.go
@@ -269,7 +269,7 @@ func NewControllerManagerCommand(ctx context.Context) *cobra.Command {
 				return fmt.Errorf("could not add ready check for webhook server to manager: %w", err)
 			}
 
-			// TODO (georgibaltiev): remove after the extension's next release
+			// TODO (georgibaltiev): Remove after the release of version 1.62.0
 			log.Info("Adding migration runnables")
 			if err := mgr.Add(manager.RunnableFunc(func(ctx context.Context) error {
 				return purgeMachineControllerManagerRBACResources(ctx, mgr.GetClient())
--- a/cmd/gardener-extension-provider-aws/app/app.go
+++ b/cmd/gardener-extension-provider-aws/app/app.go
@@ -272,7 +272,7 @@ func NewControllerManagerCommand(ctx context.Context) *cobra.Command {
 			// TODO (georgibaltiev): Remove after the release of version 1.62.0
 			log.Info("Adding migration runnables")
 			if err := mgr.Add(manager.RunnableFunc(func(ctx context.Context) error {
-				return purgeMachineControllerManagerRBACResources(ctx, mgr.GetClient())
+				return purgeMachineControllerManagerRBACResources(ctx, mgr.GetClient(), log)
 			})); err != nil {
 				return fmt.Errorf("error adding migrations: %w", err)
 			}
--- a/cmd/gardener-extension-provider-aws/app/app.go
+++ b/cmd/gardener-extension-provider-aws/app/app.go
@@ -8,7 +8,6 @@ import (
 	"context"
 	"fmt"
 	"os"
-	"slices"
 	"time"
 
 	druidcorev1alpha1 "github.com/gardener/etcd-druid/api/core/v1alpha1"
@@ -17,7 +16,6 @@ import (
 	"github.com/gardener/gardener/extensions/pkg/controller/controlplane/genericactuator"
 	"github.com/gardener/gardener/extensions/pkg/controller/heartbeat"
 	heartbeatcmd "github.com/gardener/gardener/extensions/pkg/controller/heartbeat/cmd"
-	extensionsworkercontroller "github.com/gardener/gardener/extensions/pkg/controller/worker"
 	"github.com/gardener/gardener/extensions/pkg/util"
 	webhookcmd "github.com/gardener/gardener/extensions/pkg/webhook/cmd"
 	"github.com/gardener/gardener/pkg/client/kubernetes"
@@ -272,7 +270,7 @@ func NewControllerManagerCommand(ctx context.Context) *cobra.Command {
 			}
 
 			// TODO (georgibaltiev): Remove after the release of version 1.64.0
-			if !slices.Contains(controllerSwitches.Disabled, extensionsworkercontroller.ControllerName) {
+			if reconcileOpts.ExtensionClass != "garden" {
 				log.Info("Adding migration runnables")
 				if err := mgr.Add(manager.RunnableFunc(func(ctx context.Context) error {
 					return purgeMachineControllerManagerRBACResources(ctx, mgr.GetClient(), log)
--- a/cmd/gardener-extension-provider-aws/app/migrations.go
+++ b/cmd/gardener-extension-provider-aws/app/migrations.go
@@ -1,4 +1,4 @@
-// SPDX-FileCopyrightText: 2024 SAP SE or an SAP affiliate company and Gardener contributors
+// SPDX-FileCopyrightText: 2025 SAP SE or an SAP affiliate company and Gardener contributors
 //
 // SPDX-License-Identifier: Apache-2.0
 
--- a/cmd/gardener-extension-provider-aws/app/migrations.go
+++ b/cmd/gardener-extension-provider-aws/app/migrations.go
@@ -33,7 +33,7 @@ func purgeMachineControllerManagerRBACResources(ctx context.Context, c client.Cl
 
 	for _, clusterRoleBinding := range clusterRoleBindingList.Items {
 		if nameRegex.Match([]byte(clusterRoleBinding.Name)) {
-			log.Info("Removing ClusterRoleBinding %s", client.ObjectKeyFromObject(&clusterRoleBinding))
+			log.Info("Removing ClusterRoleBinding", "ClusterRoleBinding", client.ObjectKeyFromObject(&clusterRoleBinding))
 			if err := kutil.DeleteObject(
 				ctx,
 				c,
@@ -50,13 +50,13 @@ func purgeMachineControllerManagerRBACResources(ctx context.Context, c client.Cl
 
 	for _, clusterRole := range clusterRoleList.Items {
 		if nameRegex.Match([]byte(clusterRole.Name)) {
-			log.Info("Removing ClusterRole %s", client.ObjectKeyFromObject(&clusterRole))
+			log.Info("Removing ClusterRole", "ClusterRole", client.ObjectKeyFromObject(&clusterRole))
 			if err := kutil.DeleteObject(
 				ctx,
 				c,
 				&rbacv1.ClusterRole{ObjectMeta: metav1.ObjectMeta{Name: clusterRole.Name}},
 			); err != nil {
-				return fmt.Errorf("failed to delete clusterRole: %w", err)
+				return fmt.Errorf("failed to delete ClusterRole: %w", err)
 			}
 		}
 	}
--- a/cmd/gardener-extension-provider-aws/app/migrations.go
+++ b/cmd/gardener-extension-provider-aws/app/migrations.go
@@ -13,18 +13,14 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+var nameRegex = regexp.MustCompile("extensions.gardener.cloud:provider-aws:shoot--.*:machine-controller-manager")
+
 // TODO (georgibaltiev): Remove after the release of version 1.62.0
 func purgeMachineControllerManagerRBACResources(ctx context.Context, client client.Client) error {
 	var (
 		clusterRoleBindingList = &rbacv1.ClusterRoleBindingList{}
 		clusterRoleList        = &rbacv1.ClusterRoleList{}
-		nameRegex              *regexp.Regexp
 	)
-
-	nameRegex, err := regexp.Compile("extensions.gardener.cloud:provider-aws:shoot--.*:machine-controller-manager")
-	if err != nil {
-		return fmt.Errorf("failed to compile regex: %w", err)
-	}
 
 	if err := client.List(ctx, clusterRoleBindingList); err != nil {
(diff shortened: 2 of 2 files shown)


plkokanov:
I posted some comments that are also relevant to this PR in https://github.com/gardener/gardener-extension-provider-gcp/pull/1064, as there was already a discussion started there.

cmd/gardener-extension-provider-aws/app/migrations.go
```diff
@@ -0,0 +1,53 @@
...
+		return fmt.Errorf("failed to compile regex: %w", err)
+	}
+
+	if err := client.List(ctx, clusterRoleBindingList); err != nil {
+		return fmt.Errorf("failed to list clusterRoleBindings: %w", err)
+	}
+
+	for _, clusterRoleBinding := range clusterRoleBindingList.Items {
+		if nameRegex.Match([]byte(clusterRoleBinding.Name)) {
+			if err := client.Delete(ctx, clusterRoleBinding.DeepCopy()); err != nil {
```
ialidzhikov, line 35: Use the `kutil.DeleteObject` helper func as it already ignores not found errors.
Not very likely to have a not found error in this case but as a safeguard we usually ignore not found errors on deletion.

For exampe, if we have a bug that the runnable is executed by both of the replicas (and not only by the leader one) then both of the replicas will race for the deletion and a delete request in such case could fail with not found error.

cmd/gardener-extension-provider-aws/app/migrations.go
```diff
@@ -0,0 +1,53 @@
...
+		}
+	}
+
+	if err := client.List(ctx, clusterRoleList); err != nil {
+		return fmt.Errorf("failed to list clusterRoles: %w", err)
+	}
+
+	for _, clusterRole := range clusterRoleList.Items {
+		if nameRegex.Match([]byte(clusterRole.Name)) {
+			if err := client.Delete(ctx, clusterRole.DeepCopy()); err != nil {
```
ialidzhikov, line 47: Use the `kutil.DeleteObject` helper func as it already ignores not found errors.

cmd/gardener-extension-provider-aws/app/migrations.go
```diff
@@ -0,0 +1,53 @@
...
+				return fmt.Errorf("failed to delete clusterRole: %w", err)
+			}
+		}
+	}
+	return nil
```
ialidzhikov, line 52:
```suggestion
	}

	return nil
```

cmd/gardener-extension-provider-aws/app/migrations.go
```diff
@@ -0,0 +1,53 @@
...
+		nameRegex              *regexp.Regexp
+	)
+
+	nameRegex, err := regexp.Compile("extensions.gardener.cloud:provider-aws:shoot--.*:machine-controller-manager")
+	if err != nil {
+		return fmt.Errorf("failed to compile regex: %w", err)
+	}
+
+	if err := client.List(ctx, clusterRoleBindingList); err != nil {
+		return fmt.Errorf("failed to list clusterRoleBindings: %w", err)
```
ialidzhikov, line 30:
```suggestion
		return fmt.Errorf("failed to list ClusterRoleBindings: %w", err)
```

cmd/gardener-extension-provider-aws/app/migrations.go
```diff
@@ -0,0 +1,53 @@
...
+	}
+
+	for _, clusterRoleBinding := range clusterRoleBindingList.Items {
+		if nameRegex.Match([]byte(clusterRoleBinding.Name)) {
+			if err := client.Delete(ctx, clusterRoleBinding.DeepCopy()); err != nil {
+				return fmt.Errorf("failed to delete clusterRoleBinding: %w", err)
```
ialidzhikov, line 36:
```suggestion
				return fmt.Errorf("failed to delete ClusterRoleBinding %s: %w", client.ObjectKeyFromObject(&clusterRoleBinding), err)
```

Full diff:
```diff
diff --git a/cmd/gardener-extension-provider-aws/app/migrations.go b/cmd/gardener-extension-provider-aws/app/migrations.go
index 360f90bc..4cfc2276 100644
--- a/cmd/gardener-extension-provider-aws/app/migrations.go
+++ b/cmd/gardener-extension-provider-aws/app/migrations.go
@@ -14,7 +14,7 @@ import (
 )
 
 // TODO (georgibaltiev): Remove after the release of version 1.62.0
-func purgeMachineControllerManagerRBACResources(ctx context.Context, client client.Client) error {
+func purgeMachineControllerManagerRBACResources(ctx context.Context, c client.Client) error {
 	var (
 		clusterRoleBindingList = &rbacv1.ClusterRoleBindingList{}
 		clusterRoleList        = &rbacv1.ClusterRoleList{}
@@ -26,25 +26,25 @@ func purgeMachineControllerManagerRBACResources(ctx context.Context, client clie
 		return fmt.Errorf("failed to compile regex: %w", err)
 	}
 
-	if err := client.List(ctx, clusterRoleBindingList); err != nil {
+	if err := c.List(ctx, clusterRoleBindingList); err != nil {
 		return fmt.Errorf("failed to list clusterRoleBindings: %w", err)
 	}
 
 	for _, clusterRoleBinding := range clusterRoleBindingList.Items {
 		if nameRegex.Match([]byte(clusterRoleBinding.Name)) {
-			if err := client.Delete(ctx, clusterRoleBinding.DeepCopy()); err != nil {
-				return fmt.Errorf("failed to delete clusterRoleBinding: %w", err)
+			if err := c.Delete(ctx, clusterRoleBinding.DeepCopy()); err != nil {
+				return fmt.Errorf("failed to delete ClusterRoleBinding %s: %w", client.ObjectKeyFromObject(&clusterRoleBinding), err)
 			}
 		}
 	}
 
-	if err := client.List(ctx, clusterRoleList); err != nil {
+	if err := c.List(ctx, clusterRoleList); err != nil {
 		return fmt.Errorf("failed to list clusterRoles: %w", err)
 	}
 
 	for _, clusterRole := range clusterRoleList.Items {
 		if nameRegex.Match([]byte(clusterRole.Name)) {
-			if err := client.Delete(ctx, clusterRole.DeepCopy()); err != nil {
+			if err := c.Delete(ctx, clusterRole.DeepCopy()); err != nil {
 				return fmt.Errorf("failed to delete clusterRole: %w", err)
 			}
 		}

```

cmd/gardener-extension-provider-aws/app/migrations.go
```diff
@@ -0,0 +1,53 @@
...
+	}
+
+	if err := client.List(ctx, clusterRoleList); err != nil {
+		return fmt.Errorf("failed to list clusterRoles: %w", err)
+	}
+
+	for _, clusterRole := range clusterRoleList.Items {
+		if nameRegex.Match([]byte(clusterRole.Name)) {
+			if err := client.Delete(ctx, clusterRole.DeepCopy()); err != nil {
+				return fmt.Errorf("failed to delete clusterRole: %w", err)
```
ialidzhikov, line 48:
```suggestion
				return fmt.Errorf("failed to delete ClusterRole %s: %w", client.ObjectKeyFromObject(&clusterRole), err)
```

cmd/gardener-extension-provider-aws/app/migrations.go
ialidzhikov, line 47: Do we need `DeepCopy` here?

ialidzhikov (commented):
Thanks for the PR! Suggestions - inline.

Can you also confirm that the only leaked resources from https://github.com/gardener/gardener-extension-provider-aws/pull/872 are these ClusterRoles and ClusterRoleBindings?

cmd/gardener-extension-provider-aws/app/migrations.go
```diff
@@ -0,0 +1,53 @@
...
+	"context"
+	"fmt"
+	"regexp"
+
+	rbacv1 "k8s.io/api/rbac/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// TODO (georgibaltiev): Remove after the release of version 1.62.0
+func purgeMachineControllerManagerRBACResources(ctx context.Context, client client.Client) error {
```
ialidzhikov, line 17: Could you add log entries when the runnable starts (at the place where we invoke this func), when we are about to delete a ClusterRole/ClusterRoleBinding and when the runnable (this func) ends?

cmd/gardener-extension-provider-aws/app/app.go
```diff
@@ -269,6 +269,14 @@ func NewControllerManagerCommand(ctx context.Context) *cobra.Command {
 				return fmt.Errorf("could not add ready check for webhook server to manager: %w", err)
 			}
 
+			// TODO (georgibaltiev): Remove after the release of version 1.62.0
+			log.Info("Adding migration runnables")
+			if err := mgr.Add(manager.RunnableFunc(func(ctx context.Context) error {
+				return purgeMachineControllerManagerRBACResources(ctx, mgr.GetClient())
+			})); err != nil {
+				return fmt.Errorf("error adding migrations: %w", err)
+			}
```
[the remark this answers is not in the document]
plkokanov: These runnables are only executed for the leader replicas. This is the part of the code that adds runnables to each group:

https://github.com/kubernetes-sigs/controller-runtime/blob/85ee7a9efa4b4cec8e8ef473e3ed1d7adf8dc132/pkg/manager/runnable_group.go#L55-L76

It seems like they decided that if the runnable does not implement the `LeaderElectionRunnable` interface it is added by default to the leader election runnables.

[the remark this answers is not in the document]
ialidzhikov: Thanks for explaining! Then, let's keep it as it is. 
I was not sure if a runnable added with the `RunnableFunc` will be executed for the leader replica only or not.

cmd/gardener-extension-provider-aws/app/app.go
```diff
@@ -269,6 +269,14 @@ func NewControllerManagerCommand(ctx context.Context) *cobra.Command {
 				return fmt.Errorf("could not add ready check for webhook server to manager: %w", err)
 			}
 
+			// TODO (georgibaltiev): Remove after the release of version 1.62.0
```
ialidzhikov, line 272: Nit: In Gardener, we usually keep such migration codes for at least 3 releases.

From https://github.com/gardener/gardener/blob/v1.118.0/docs/development/process.md#deprecations-and-backwards-compatibility:
> In case you have to run migration code (_which is mostly internal_), please add a [TODO statement](#todo-statements) to remove it only after **3 releases**.

```suggestion
			// TODO (georgibaltiev): Remove after the release of version 1.64.0
```

cmd/gardener-extension-provider-aws/app/migrations.go
```diff
@@ -0,0 +1,66 @@
...
+	kutil "github.com/gardener/gardener/pkg/utils/kubernetes"
+	"github.com/go-logr/logr"
+	rbacv1 "k8s.io/api/rbac/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var nameRegex = regexp.MustCompile("extensions.gardener.cloud:provider-aws:shoot--.*:machine-controller-manager")
+
+// TODO (georgibaltiev): Remove after the release of version 1.62.0
```
ialidzhikov, line 21:
```suggestion
// TODO (georgibaltiev): Remove after the release of version 1.64.0
```

cmd/gardener-extension-provider-aws/app/migrations.go
```diff
@@ -0,0 +1,66 @@
...
+func purgeMachineControllerManagerRBACResources(ctx context.Context, c client.Client, log logr.Logger) error {
+	log.Info("Starting the removal of obsolete ClusterRoles and ClusterRoleBindings")
```
ialidzhikov, line 23: Nit: For consistency reasons between logs (atm use "remove") and error messages (atm use "delete").

```suggestion
	log.Info("Starting the deletion of obsolete ClusterRoles and ClusterRoleBindings")
```

cmd/gardener-extension-provider-aws/app/migrations.go
```diff
@@ -0,0 +1,66 @@
...
+		clusterRoleList        = &rbacv1.ClusterRoleList{}
+	)
+
+	if err := c.List(ctx, clusterRoleBindingList); err != nil {
+		return fmt.Errorf("failed to list ClusterRoleBindings: %w", err)
+	}
+
+	for _, clusterRoleBinding := range clusterRoleBindingList.Items {
+		if nameRegex.Match([]byte(clusterRoleBinding.Name)) {
+			log.Info("Removing ClusterRoleBinding", "ClusterRoleBinding", client.ObjectKeyFromObject(&clusterRoleBinding))
```
ialidzhikov, line 36: From https://github.com/gardener/gardener/blob/v1.118.0/docs/development/logging.md#keys-and-values:

> Use [lowerCamelCase](https://en.wiktionary.org/wiki/lowerCamelCase) for keys. Don't put spaces in keys, as it will make log processing with simple tools like jq harder.

```suggestion
			log.Info("Deleting clusterRoleBinding", "ClusterRoleBinding", client.ObjectKeyFromObject(&clusterRoleBinding))
```

cmd/gardener-extension-provider-aws/app/migrations.go
```diff
@@ -0,0 +1,66 @@
...
+		}
+	}
+
+	if err := c.List(ctx, clusterRoleList); err != nil {
+		return fmt.Errorf("failed to list ClusterRoles: %w", err)
+	}
+
+	for _, clusterRole := range clusterRoleList.Items {
+		if nameRegex.Match([]byte(clusterRole.Name)) {
+			log.Info("Removing ClusterRole", "ClusterRole", client.ObjectKeyFromObject(&clusterRole))
```
ialidzhikov, line 53:
```suggestion
			log.Info("Deleting ClusterRole", "clusterRole", client.ObjectKeyFromObject(&clusterRole))
```

cmd/gardener-extension-provider-aws/app/migrations.go
```diff
@@ -0,0 +1,66 @@
...
+				ctx,
+				c,
+				&rbacv1.ClusterRole{ObjectMeta: metav1.ObjectMeta{Name: clusterRole.Name}},
+			); err != nil {
+				return fmt.Errorf("failed to delete ClusterRole: %w", err)
+			}
+		}
+	}
+
+	log.Info("Succesfully removed the obsolete ClusterRoles and ClusterRoleBindings")
```
ialidzhikov, line 64:
```suggestion
	log.Info("Successfully deleted the obsolete ClusterRoles and ClusterRoleBindings")
```

ialidzhikov (commented):
Final round of comments, otherwise lgtm

cmd/gardener-extension-provider-aws/app/migrations.go
ialidzhikov: This suggestion is not applied. Could you check again?

cmd/gardener-extension-provider-aws/app/app.go
```diff
@@ -269,6 +271,16 @@ func NewControllerManagerCommand(ctx context.Context) *cobra.Command {
 				return fmt.Errorf("could not add ready check for webhook server to manager: %w", err)
 			}
 
+			// TODO (georgibaltiev): Remove after the release of version 1.64.0
+			if !slices.Contains(controllerSwitches.Disabled, extensionsworkercontroller.ControllerName) {
```
ialidzhikov, line 275: I would rather suggest introducing env var whether the extension runs in runtime cluster or not. Otherwise, we are relying on implicit contract.

Suggestion:
```diff
diff --git a/charts/gardener-extension-provider-aws/templates/deployment.yaml b/charts/gardener-extension-provider-aws/templates/deployment.yaml
index 61d0d7a3..2c03dd60 100644
--- a/charts/gardener-extension-provider-aws/templates/deployment.yaml
+++ b/charts/gardener-extension-provider-aws/templates/deployment.yaml
@@ -94,6 +94,10 @@ spec:
         - name: IMAGEVECTOR_OVERWRITE
           value: /charts_overwrite/images_overwrite.yaml
         {{- end }}
+        {{- if (include "runtimeCluster.enabled" .) }}
+        - name: RUNTIME_CLUSTER
+          value: enabled
+        {{- end }}
         securityContext:
           allowPrivilegeEscalation: false
         livenessProbe:
diff --git a/cmd/gardener-extension-provider-aws/app/app.go b/cmd/gardener-extension-provider-aws/app/app.go
index 93604cbe..9498a829 100644
--- a/cmd/gardener-extension-provider-aws/app/app.go
+++ b/cmd/gardener-extension-provider-aws/app/app.go
@@ -8,7 +8,6 @@ import (
 	"context"
 	"fmt"
 	"os"
-	"slices"
 	"time"
 
 	druidcorev1alpha1 "github.com/gardener/etcd-druid/api/core/v1alpha1"
@@ -17,7 +16,6 @@ import (
 	"github.com/gardener/gardener/extensions/pkg/controller/controlplane/genericactuator"
 	"github.com/gardener/gardener/extensions/pkg/controller/heartbeat"
 	heartbeatcmd "github.com/gardener/gardener/extensions/pkg/controller/heartbeat/cmd"
-	extensionsworkercontroller "github.com/gardener/gardener/extensions/pkg/controller/worker"
 	"github.com/gardener/gardener/extensions/pkg/util"
 	webhookcmd "github.com/gardener/gardener/extensions/pkg/webhook/cmd"
 	"github.com/gardener/gardener/pkg/client/kubernetes"
@@ -272,7 +270,7 @@ func NewControllerManagerCommand(ctx context.Context) *cobra.Command {
 			}
 
 			// TODO (georgibaltiev): Remove after the release of version 1.64.0
-			if !slices.Contains(controllerSwitches.Disabled, extensionsworkercontroller.ControllerName) {
+			if runtimeCluster := os.Getenv("RUNTIME_CLUSTER"); runtimeCluster != "enabled" {
 				log.Info("Adding migration runnables")
 				if err := mgr.Add(manager.RunnableFunc(func(ctx context.Context) error {
 					return purgeMachineControllerManagerRBACResources(ctx, mgr.GetClient(), log)

```

@plkokanov @kon-angelo WDYT?

cmd/gardener-extension-provider-aws/app/migrations.go
ialidzhikov: ping

cmd/gardener-extension-provider-aws/app/migrations.go
```diff
@@ -0,0 +1,66 @@
...
+				ctx,
+				c,
+				&rbacv1.ClusterRole{ObjectMeta: metav1.ObjectMeta{Name: clusterRole.Name}},
+			); err != nil {
+				return fmt.Errorf("failed to delete ClusterRole: %w", err)
+			}
+		}
+	}
+
+	log.Info("Succesfully deleted the obsolete ClusterRoles and ClusterRoleBindings")
```
ialidzhikov, line 64: It was part of my previous comment but not yet addressed.

```suggestion
	log.Info("Successfully deleted the obsolete ClusterRoles and ClusterRoleBindings")
```

cmd/gardener-extension-provider-aws/app/app.go
kon-angelo: Or a simple boolean would work that we can parse, but otherwise a fine suggestion.

plkokanov: We discussed this on slack with @georgibaltiev on Friday and saw that the way he has made the check whether the worker is disabled would not work anyway. For the runtime cluster we just set which controller is enabled and don't explicitly disable controllers: https://github.com/gardener/gardener-extension-provider-aws/blob/36063c06aa0a86e490a29c8e077d9b60f96aa98c/charts/gardener-extension-provider-aws/templates/deployment.yaml#L74-L77

Apart from the env variable we also talked about reusing the option that specifies the class of the extensions that this controller manages: https://github.com/gardener/gardener-extension-provider-aws/blob/36063c06aa0a86e490a29c8e077d9b60f96aa98c/charts/gardener-extension-provider-aws/templates/deployment.yaml#L77
So that a check like this could be used:
```
if reconcileOpts.ExtensionClass != "garden" {
...
```

I think I'm more inclined towards reusing the extension class because:
1. This code is only temporary
2. The extension class is anyway how we separate which controller is responsible for which extensions depending on the cluster type.
3. There isn't any other place where we check directly whether the controller runs "for" the `runtime` cluster or not.
4. Additionally to 2, we might have 2 controllers that run in the runtime cluster (although for one of them `runtime.enabled=false`) when the runtime cluster is both a garden runtime and seed cluster, so `runtime=enabled` for one controller, and no such environment for the other controller might be a bit confusing considering there's already the flag.

georgibaltiev: If I understand correctly, the question here is about specificness - the ExtensionClass variable is designed with the idea for us check what type of cluster we are working on, while the RUNTIME_CLUSTER environment variable is not so set in stone and could be evaluated to false, even if we are in a runtime environment, correct?

kon-angelo: I am also on board with the ExtensionClass. Thank you @plkokanov @georgibaltiev

kon-angelo (approved):
/lgtm

PR #1363, opened 2025-06-12
Set AWS_REGION for cloud controller manager

dimityrmirchev:

/area quality
/kind bug
/platform aws

When using workload identity I observe the following error in cloud controller manager.

```
W0612 06:40:35.700604       1 instances_v2.go:90] Failed to get node topology. Moving on without setting labels: "operation error EC2: DescribeInstanceTopology, get identity: get credentials: failed to refresh cached credentials, failed to retrieve credentials, operation error STS: AssumeRoleWithWebIdentity, failed to resolve service endpoint, endpoint rule error, Invalid Configuration: Missing Region"
E0612 06:40:35.948326       1 topology.go:95] Error describing instance topology: "operation error EC2: DescribeInstanceTopology, get identity: get credentials: failed to refresh cached credentials, failed to retrieve credentials, operation error STS: AssumeRoleWithWebIdentity, failed to resolve service endpoint, endpoint rule error, Invalid Configuration: Missing Region"
```

This error is fixed when `AWS_REGION` env variable is set for the CCM deployment.

Fixes #

**Release note**:

```bugfix operator
A bug causing `cloud-controller-manager` to fail to get node topology because of missing region configuration was fixed.
```

### DIFF
--- a/charts/internal/seed-controlplane/charts/cloud-controller-manager/templates/deployment.yaml
+++ b/charts/internal/seed-controlplane/charts/cloud-controller-manager/templates/deployment.yaml
@@ -78,6 +78,8 @@ spec:
         - name: AWS_SHARED_CREDENTIALS_FILE
           value: /srv/cloudprovider/credentialsFile
         {{- end }}
+        - name: AWS_REGION
+          value: {{ .Values.region }}
         securityContext:
           allowPrivilegeEscalation: false
         livenessProbe:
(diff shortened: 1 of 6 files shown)


kon-angelo:
It can be tackled with this PR at least for the upstream controllers. If current provider-aws can deal with the credentials as is, then the rest of our components should be safe(r)

dimityrmirchev:
As far as I can see we either set `--region` flag or `AWS_REGION` for the upstream components. I added the `region` value to the shared credentials file as suggested with https://github.com/gardener/gardener-extension-provider-aws/pull/1363/commits/b17574e153ac2572a819f3a8c8d1668cd0570cab

kon-angelo (approved):
/lgtm

PR #1383, opened 2025-07-04
Support for configuring dedicated tenancy for VPC

### LINKED ISSUE #1377: As a user, I want to be able to enable dedicated tenancy for my gardener-managed VPC, such that all my instances get dedicated tenancy automatically

/area control-plane
/kind enhancement
/platform aws

AWS allows for a VPC to have "dedicated tenancy" enabled: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/change-tenancy-vpc.html
With this setting, all instances created in this VPC will automatically get dedicated tenancy enabled.

This is a setting that cannot be enabled for already existing VPCs, but AWS allows to disable it. Not sure if we want to allow modifications of this property, given this is something that can only be changed in one direction.

Additional context:
* aws-sdk-v2 has this as a property on `CreateVpcInput` already: https://github.com/aws/aws-sdk-go-v2/blob/32873b942d0ba5f63d200a39321173123075aad4/service/ec2/api_op_CreateVpc.go#L67-L77
* We most likely just need to expose it in the spec and pass it in here: https://github.com/gardener/gardener-extension-provider-aws/blob/5043f4e86ed3bbd3e4545c147eb8a041aaf23d8b/pkg/aws/client/client.go#L764

A similar request already exists on a worker group level: https://github.com/gardener/gardener-extension-provider-aws/issues/1162

voelzmo:

/area control-plane
/kind enhancement
/platform aws

Add the option to configure [dedicated tenancy](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/change-tenancy-vpc.html) for a gardener-managed VPC using a new property `enableDedicatedTenancyForVPC`.

Fixes #1377 

* This PR introduces the property as immutable, you can set this only on VPC creation. Technically, AWS allows for _disabling_ this on an existing VPC, but there is no way back: enabling on an existing VPC is forbidden. But this seems dangerous, as it allows you to remove an important compliance configuration when there is no way back.
* While the property on AWS level is a simple `instanceTenancy` allowing for values `default` and `dedicated` (and for instances, there is an additional value `host`), this PR introduces a boolean config option `enableDedicatedTenancyForVPC`
* There was already [an existing property `awsclient.VPC.InstanceTenancy` of type `*string`](https://github.com/gardener/gardener-extension-provider-aws/blob/229638b4655819bd0483dadf33412404bb7f2830/pkg/aws/client/types.go#L219) while `instanceTenancy` in the aws-sdk is [a string-alias enum of type `Tenancy`](https://github.com/aws/aws-sdk-go-v2/blob/32873b942d0ba5f63d200a39321173123075aad4/service/ec2/api_op_CreateVpc.go#L67-L77). Not sure if this could/should be changed on our end such that we don't have the conversions from enum to string and back? This seems to have come in with the tf-to-flow migration, so not sure if this is coincidence or intentional.

**Release note**:

```other operator
Support for configuring dedicated tenancy for VPC with a new property `enableDedicatedTenancyForVPC`
```

### DIFF
--- a/pkg/apis/aws/validation/infrastructure_test.go
+++ b/pkg/apis/aws/validation/infrastructure_test.go
@@ -18,6 +18,8 @@ import (
 	. "github.com/gardener/gardener-extension-provider-aws/pkg/apis/aws/validation"
 )
 
+var dedicatedTenancyEnabled *bool
+
 var _ = Describe("InfrastructureConfig validation", func() {
 	var (
 		infrastructureConfig *apisaws.InfrastructureConfig
@@ -40,6 +42,10 @@ var _ = Describe("InfrastructureConfig validation", func() {
 	)
 
 	BeforeEach(func() {
+		dedicatedTenancyEnabled = nil
+	})
+
+	JustBeforeEach(func() {
 		infrastructureConfig = &apisaws.InfrastructureConfig{
 			Networks: apisaws.Networks{
 				VPC: apisaws.VPC{
@@ -54,7 +60,7 @@ var _ = Describe("InfrastructureConfig validation", func() {
 					},
 				},
 			},
-			EnableDedicatedTenancyForVPC: ptr.To(false),
+			EnableDedicatedTenancyForVPC: dedicatedTenancyEnabled,
 		}
 	})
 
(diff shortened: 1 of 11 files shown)


pkg/aws/client/client.go
```diff
@@ -765,6 +765,7 @@ func (c *Client) CreateVpc(ctx context.Context, desired *VPC) (*VPC, error) {
 		CidrBlock:                   aws.String(desired.CidrBlock),
 		AmazonProvidedIpv6CidrBlock: aws.Bool(desired.AssignGeneratedIPv6CidrBlock),
 		TagSpecifications:           desired.ToTagSpecifications(ec2types.ResourceTypeVpc),
+		InstanceTenancy:             ec2types.Tenancy(*desired.InstanceTenancy),
```
wpross, line 768: As you already mentioned, the `InstanceTenancy *string` in the `VPC` struct in `pkg/aws/client/types.go` already existed and using it requires this conversion. I would change it to `InstanceTenancy *types.Tenancy` (from https://github.com/aws/aws-sdk-go-v2/tree/main/service/ec2/types). This ensures type safety and is easier to handle/read.

pkg/controller/infrastructure/infraflow/reconcile.go
```diff
@@ -186,12 +186,17 @@ func (c *FlowContext) getIpFamilies() []v1beta1.IPFamily {
 func (c *FlowContext) ensureManagedVpc(ctx context.Context) error {
 	log := LogFromContext(ctx)
 	log.Info("using managed VPC")
+	instanceTenancy := ptr.To(string(ec2types.TenancyDefault))
```
wpross, line 189: maybe it would be good to add a comment like
```
// Default to shared tenancy unless dedicated tenancy is explicitly enabled.
```

wpross (changes requested):

pkg/aws/client/client.go
voelzmo: Done. I wasn't sure about what happens when migrating existing terraform state to flow and if there was a reason why this property was introduced as ` *string`. Also not entirely sure how to test if this is an issue?

voelzmo:
rebased on current master and brought back the import of `ec2types` which got removed in https://github.com/gardener/gardener-extension-provider-aws/pull/1397

pkg/apis/aws/validation/infrastructure_test.go
```diff
@@ -584,6 +585,18 @@ var _ = Describe("InfrastructureConfig validation", func() {
 				})),
 			))
 		})
+
+		It("should forbid changing dedicated tenancy configuration", func() {
+			newInfrastructureConfig := infrastructureConfig.DeepCopy()
+			newInfrastructureConfig.EnableDedicatedTenancyForVPC = ptr.To(true)
+
+			errorList := ValidateInfrastructureConfigUpdate(infrastructureConfig, newInfrastructureConfig)
+
```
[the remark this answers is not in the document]
wpross: You're right. Unfortunately, it's not consistent for all tests in the filed, but that shouldn't be your problem as your tests are consistent :)

wpross (approved):
/lgtm

Issue #1288, opened 2025-04-21
Backup-bucket deletion will fail if there are orphan backupbucket entry(s) present in bucket.


ishan16696:
/area backup
/kind bug
/platform aws

**What happened**:
While testing this PR: https://github.com/gardener/gardener-extension-provider-aws/pull/1261, I have observed that the backup-bucket controller fails to delete a backup bucket if there are orphaned backup entries present.
It happened due to this `func DeleteBucketIfExists()` failed to calls `func DeleteObjectsWithPrefix()` as error code returned wasn't `BucketAlreadyExists`.
https://github.com/gardener/gardener-extension-provider-aws/blob/5f900092991436b748ac8fc805adb3440d93ac54/pkg/aws/client/client.go#L467-L485

**What you expected to happen**:
On seed deletion, `backup-bucket controller` should be able to delete the backup-bucket even if some orphan backup-entries present in the bucket.

**How to reproduce it (as minimally and precisely as possible)**:
1. Set the local seed with backup enabled.
2. Create some backup-entry, put some object in that backup-entry.
3. Trigger the seed deletion.

Logs of `gardener-extension-provider-aws`:
```
{"level":"info","ts":"2025-04-20T18:29:31.872Z","msg":"Starting the deletion of BackupBucket","controller":"backupbucket","namespace":"","name":"0adbe95d-4e8a-4015-a773-94349706a1ea","reconcileID":"9f95c479-d794-4fc5-b23f-e3aa7601f8dc"}
{"level":"info","ts":"2025-04-20T18:29:31.872Z","msg":"Starting deletion of BackupBucket...","controller":"backupbucket","namespace":"","name":"0adbe95d-4e8a-4015-a773-94349706a1ea","reconcileID":"9f95c479-d794-4fc5-b23f-e3aa7601f8dc"}
[...]

PR #1384, opened 2025-07-07
Fix a bug in deletion of backup-bucket if backup-bucket didn't even get created.

ishan16696:
/area backup
/kind bug
/platform aws

While working on this bug https://github.com/gardener/gardener-extension-provider-aws/issues/1288, I discovered that S3 does not allow the creation of a bucket if a bucket with the same name already exists in a different region. (It returns error code as `PermanentRedirect`). I addressed this error during the backup-bucket creation process in the reconciliation https://github.com/gardener/gardener-extension-provider-aws/blob/229638b4655819bd0483dadf33412404bb7f2830/pkg/controller/backupbucket/actuator_reconcile.go#L89-L90, but I didn't handle it during the backup-bucket deletion process. This PR resolves the issue by managing the deletion of the backup bucket as backup-bucket was never created.

Fixes #

**Release note**:

```bugfix operator
Fix a bug in the deletion process of a backup bucket when the bucket was somehow never got created.
```


hebelsan:
Can we maybe use consts instead of strings for these errors, especially since we use them in multiple places?
Potentially place them here: https://github.com/gardener/gardener-extension-provider-aws/blob/13c196240de035e4c0e72227e989fb47826bc88c/pkg/aws/types.go

ishan16696:
> Can we maybe use consts instead of strings for these errors, especially since we use them in multiple places?

done, I added them here: https://github.com/gardener/gardener-extension-provider-aws/blob/13c196240de035e4c0e72227e989fb47826bc88c/pkg/aws/client/types.go#L27

hebelsan (approved):
/lgtm

PR #1397, opened 2025-07-15
Fix failed NATGateway reconciliation

hebelsan:

/area robustness
/kind bug
/platform aws

Multiple issues were identified:
- we never wait for the subnet to be available which can lead to race conditions and a NAT in failed state
- we filter for the failed state in createNat which is why we created an additional NATGateway
- we never filter and delete the failed nat gateway.
- we list all 4 NATGWs and don't filter for the failed ones for the EgressCIDRs

Fixes #

**Release note**:

```bugfix operator
Fix failed NATGateway reconciliation
```

### DIFF
--- a/pkg/controller/infrastructure/infraflow/reconcile.go
+++ b/pkg/controller/infrastructure/infraflow/reconcile.go
@@ -18,7 +18,6 @@ import (
 	"text/template"
 	"time"
 
-	ec2types "github.com/aws/aws-sdk-go-v2/service/ec2/types"
 	"github.com/gardener/gardener/pkg/apis/core/v1beta1"
 	"github.com/gardener/gardener/pkg/utils/flow"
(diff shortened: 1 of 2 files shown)


kon-angelo (commented):
Just 2 cents, but since we wait for the subnet to become available, I would suggest we do the same for the NATGW.

hebelsan:
For NatGateways there's no such build in function to wait for it to become available like there is for subnets.
We do already have our own implementation though:
https://github.com/gardener/gardener-extension-provider-aws/blob/b3cb10d7ab129a9f592eac9e5caae8ff26c1efe8/pkg/controller/infrastructure/infraflow/reconcile.go#L1178

hebelsan:
I reverted the manual deletion of a NatGateway if it is in a failed state since this is done automatically like you mentioned in your comment.

kon-angelo (approved):
/lgtm

PR #1403, opened 2025-07-18
Allow creation or more that one IPv6 shoot in one VPC

axel7born:

/area networking
/kind bug
/platform aws

Due to the static IPv6 CIDR calculation subnets of multiple shoots get the same CIDR and the creation fails for all but one shoot. This PR retries the subnet creation with the next /64 CIDR block in case of an `InvalidSubnet.Conflict` error.

Fixes #

**Release note**:

```bug user
Now it is possible to create more than shoot with IPv6 networking in the same VPC.

```

### DIFF
--- a/pkg/controller/infrastructure/infraflow/reconcile.go
+++ b/pkg/controller/infrastructure/infraflow/reconcile.go
@@ -842,7 +842,7 @@ func (c *FlowContext) addSubnetReconcileTasks(g *flow.Graph, desired, current *a
 	}
 	suffix := fmt.Sprintf("%s-%s", zoneName, subnetKey)
 	if ptr.Deref(desired.AssignIpv6AddressOnCreation, true) {
-		return c.AddTask(g, "ensure subnet "+suffix,
+		return c.AddTask(g, "ensure IPv6 subnet "+suffix,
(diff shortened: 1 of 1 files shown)


pkg/controller/infrastructure/infraflow/reconcile.go
```diff
@@ -929,12 +929,33 @@ func (c *FlowContext) ensureSubnet(subnetKey string, desired, current *awsclient
...
+			var lastErr error
+			for attempts := 0; attempts < 256; attempts++ {
+				created, err := c.client.CreateSubnet(ctx, desired, defaultTimeout)
+				if err == nil {
+					zoneChild.Set(subnetKey, created.SubnetId)
+					return nil
+				}
+				// Check for InvalidSubnet.Conflict error
+				apiErrCode := awsclient.GetAWSAPIErrorCode(err)
+				if apiErrCode == "InvalidSubnet.Conflict" {
```
hebelsan, line 941: We assume this error only occurs for IPv6 subnets, but I'm in favour of having two different ensurerSubnets tasks one for IPv4 and the other for IPv6.

hebelsan (changes requested):

kon-angelo (approved):
/lgtm

PR #1409, opened 2025-07-22
Fix setting NAT ID only after available

hebelsan:

/area quality
/kind enhancement
/platform aws

We should only persist state and set Nat ID after it's available.

Fixes #

**Release note**:

```bugfix operator
Fix persisting NAT ID before NAT is available
```


kon-angelo (approved):
/lgtm

PR #1443, opened 2025-08-14
fix gatewayEndpoints naming validation

### LINKED ISSUE #1442: gatewayEndpointPattern not validated properly
/area control-plane
/kind bug
/platform aws

**What happened**:
creating vpc gatewayEndpoints in shoot is failing when endpoint contains "-"
this happened to me on creation of "guardduty-data" endpoint 
checking the validation code also shows that "_" would be allowed - but thats not a valid character within domain names

gatewayEndoints names should sucessfully validate against regex [a-zA-Z0-9\-] instead of \w

danpaul81:
/area control-plane
/kind bug
/platform aws

gatewayEndpoints naming follows domain naming. current code allowed for A-Za-z0-9\_ (which is wrong and misses \- char)

Fixes #1442

### DIFF
--- a/pkg/apis/aws/validation/infrastructure.go
+++ b/pkg/apis/aws/validation/infrastructure.go
@@ -21,7 +21,7 @@ import (
 )
 
 // valid values for networks.vpc.gatewayEndpoints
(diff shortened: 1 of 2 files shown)


hebelsan (approved):
/lgtm

PR #1448, opened 2025-08-18
Remove CPU requests for aws-extension components in Shoot and Seed

voelzmo:

/area auto-scaling
/kind enhancement
/platform aws

This PR removes the CPU requests from the Pod templates and switches the VPAs to control memory only.
Our analysis showed that these components are barely using the minimum of 10mCores CPU for most of the time, which means VPA isn't really helpful. Making things worse: VPA cannot recommend any value lower than 10mCores, so even if a container only uses 1mCore, it will still always request 10m.

For components with a CPU usage that low, it is better to switch off VPA entirely to save resources.

**Release note**:

```other operator
Remove CPU requests for aws-extension components in Shoot and Seed.
```


hebelsan:
Thanks for the PR @voelzmo!
Can you please run `make generate`.
Also, we should probably do a similar PR for the other providers as well right?

PR #1451, opened 2025-08-20
Add notes regarding quotas to dual-stack ingress docs.

axel7born:

/area docs
/kind networking
/platform aws

During the migration of load balancers to dual-stack, new load balancers are created while old load balancers still exist.
Added a section to the docs, to make users aware of that fact.

Fixes #

**Release note**:

```other operator
None
```


hebelsan (approved):
/lgtm

PR #1454, opened 2025-08-22
Add generation for `Extension` example

timuthy:

/area ops-productivity
/kind enhancement
/platform aws

This PR adds the generation for an `Extension` example manifest, similar to [`example/controller-registration.yaml`](https://github.com/gardener/gardener-extension-provider-aws/blob/master/example/controller-registration.yaml) which will be replaced on the long run.

**Release note**:

```other operator
An example `Extension` manifest for extension registration has been added. It can be found at [`example/extension.yaml`](https://github.com/gardener/gardener-extension-provider-aws/blob/master/example/extension.yaml)
```


hebelsan (approved):
/lgtm

PR #1458, opened 2025-08-28
[GEP-33] Capabilities based MachineClass creation

Roncossek:

/area control-plane
/kind enhancement
/platform aws

This PR is part of [GEP-33 Machine Image Capabilities](https://github.com/gardener/gardener/blob/master/docs/proposals/33-machine-image-capabilities.md)

To make use of the introduced machine `Capabilities`. This PR implements the AMI selection based on these `Capabilities`. The selection algorithm distinguishes if a shoot is created based on a `CloudProfile` with or without `Capabilities` to select the best matching AMI for an `MachineImageVersion`. It also ensures that the status of `Workers` is consistently using `Capabilities` or not to support switching existing `Workers` to a `Cloudprofile` with `Capabilities` and the other way round.

Part of: https://github.com/gardener/gardener/issues/11301

Image selection for bastions is not included as it depends on https://github.com/gardener/gardener/pull/12834
@hebelsan 

**Release note**:

```other operator
AMI selection for workers' `MachineClass` supports `Cloudprofiles` with `Capabilities`.
```

### DIFF
--- a/pkg/controller/worker/machines.go
+++ b/pkg/controller/worker/machines.go
@@ -18,6 +18,7 @@ import (
 	"github.com/gardener/gardener/extensions/pkg/controller"
 	"github.com/gardener/gardener/extensions/pkg/controller/worker"
 	genericworkeractuator "github.com/gardener/gardener/extensions/pkg/controller/worker/genericactuator"
+	"github.com/gardener/gardener/pkg/apis/core"
 	gardencorev1beta1 "github.com/gardener/gardener/pkg/apis/core/v1beta1"
 	v1beta1constants "github.com/gardener/gardener/pkg/apis/core/v1beta1/constants"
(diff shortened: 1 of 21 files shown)


pkg/controller/worker/machines.go
```diff
@@ -504,3 +511,68 @@ func isIPv6(c *controller.Cluster) bool {
...
+				// image is already legacy format
+				uniformMachineImages = appendMachineImage(uniformMachineImages, img, definitions)
+				continue
+			}
+			// transform to legacy format by using the Architecture capability if it exists
+			var architecture *string
+			if len(img.Capabilities[v1beta1constants.ArchitectureName]) > 0 {
+				architecture = &img.Capabilities[v1beta1constants.ArchitectureName][0]
+			} else {
+				architecture = nil
```
hebelsan, line 532: I think this is not necessary since the zero value of a pointer is nil

hebelsan (approved):

PR #1459, opened 2025-08-28
Add annotation to disable patching of load balancer services.

axel7born:

/area networking
/kind enhancement
/platform aws

In IPv6 and dual-stack clusters, LoadBalancer services are automatically annotated to deploy dual-stack network load balancers.

Users can opt out of this automatic behavior by setting the annotation `extensions.gardener.cloud/ignore-load-balancer: "true"`. This allows users to:
- Configure custom load balancer annotations
- Avoid issues during service migration
- Maintain full control over load balancer configuration when needed

Fixes #

**Release note**:

```other operator
LoadBalancer services in IPv6 and dual-stack clusters can now opt out of automatic dual-stack annotations using `extensions.gardener.cloud/ignore-load-balancer: "true"`.
```


ScheererJ (approved):
/lgtm

PR #1475, opened 2025-09-16
Revert "Allow deletion of duplicated zones"

hebelsan:

/area quality
/kind cleanup
/platform aws

This PR reverts the change that allowed deletion of duplicated zones (see https://github.com/gardener/gardener-extension-provider-aws/pull/1425).
The reason is that removing a duplicated zone may cause the cluster to break.

Fixes #

**Release note**:

```other operator
Disallow deletion of duplicated zones in infraConfig section
```


kon-angelo (approved):
/lgtm

PR #1488, opened 2025-09-24
Create/delete release branches automatically

hebelsan:

/area quality
/kind enhancement
/platform aws

This PR enhances the release workflow by introducing two changes:
- When a minor release is triggered, the action now also creates and prepares a corresponding release branch.
- A separate action monitors all release branches and automatically deletes any that are older than the last 3 minor versions. (In case of v1.6, v1.5, v1.4, v1.3, v1.2 -> v1.3 and v1.2 should be deleted)

Fixes #

**Release note**:

```other operator
NONE
```

### DIFF
--- a/.github/workflows/release-cleanup.yaml
+++ b/.github/workflows/release-cleanup.yaml
@@ -48,14 +48,14 @@ jobs:
           echo "Will delete all release branches <= v$major.$cutoff_minor"
 
           # List all branches and filter ones that match release-vX.Y
-          for branch in $(gh api repos/$REPO/branches --paginate --jq '.[].name'); do
+          for branch in $(gh api repos/$GH_REPO/branches --paginate --jq '.[].name'); do
             if [[ $branch =~ ^release-v([0-9]+)\.([0-9]+) ]]; then
               b_major=${BASH_REMATCH[1]}
               b_minor=${BASH_REMATCH[2]}
 
               if [[ $b_major -eq $major && $b_minor -le $cutoff_minor ]]; then
                 echo "Deleting old branch: $branch"
-                gh api -X DELETE "repos/$REPO/git/refs/heads/$branch" || true
+                gh api -X DELETE "repos/$GH_REPO/git/refs/heads/$branch" || true
               fi
             fi
           done


.github/workflows/create-branch.yaml
```diff
@@ -0,0 +1,51 @@
...
+      - name: Create new branch
+        run: |
+          # Validate input
+          if [ -z "${{ inputs.version }}" ]; then
+            echo "ERROR: version input is empty"
+            exit 1
+          fi
+          
+          VERSION="${{ inputs.version }}"
+          MINOR_VERSION="${VERSION::-2}"
```
kon-angelo, line 31: I think in the other script you used a different approach to strip everything past the last `.`. I feel that it is more robust or at least let's be consistent with techniques

.github/workflows/create-branch.yaml
```diff
@@ -0,0 +1,51 @@
...
+          BRANCH_NAME="release-${MINOR_VERSION}"
+          
+          git checkout -b "${BRANCH_NAME}"
+          
+          # Compute next development version
+          NEXT_VERSION="${MINOR_VERSION}.1"
+          DEV_VERSION="${NEXT_VERSION}-dev"
```
kon-angelo, line 38: I guess it can be condensed 🤷

.github/workflows/release-cleanup.yaml
```diff
@@ -0,0 +1,73 @@
...
+          if [[ -z "$latest_tag" ]]; then
+            echo "No valid releases found."
+            exit 0
+          fi
+
+          echo "Latest release: $latest_tag"
+
+          # Parse major and minor
+          IFS='.' read -r major minor patch <<<"${latest_tag#v}"
+          old_minor=$((minor - 3))
```
kon-angelo, line 43: I guess this can be moved after the check

.github/workflows/release-cleanup.yaml
```diff
@@ -0,0 +1,73 @@
...
+                      jq -r '.[].name' | grep "^release-$target_prefix")
+
+          for branch in $branches; do
+            echo "Deleting branch: $branch"
+            curl -s -X DELETE "${headers[@]}" "https://api.github.com/repos/$REPO/git/refs/heads/$branch"
+          done
+
+          # Delete (close) old milestones
+          milestones=$(curl -s "${headers[@]}" "https://api.github.com/repos/$REPO/milestones?state=all" | \
+                        jq -r '.[] | select(.title | startswith("'"$target_prefix"'")) | .number')
```
kon-angelo, line 65: isn't that what the other script also does ?

.github/workflows/release-cleanup.yaml
```diff
@@ -0,0 +1,73 @@
...
+          echo "Target version to clean: $target_prefix"
+
+          # Delete old release branches
+          branches=$(curl -s "${headers[@]}" "https://api.github.com/repos/$REPO/branches" | \
+                      jq -r '.[].name' | grep "^release-$target_prefix")
+
+          for branch in $branches; do
+            echo "Deleting branch: $branch"
+            curl -s -X DELETE "${headers[@]}" "https://api.github.com/repos/$REPO/git/refs/heads/$branch"
+          done
```
kon-angelo, line 61: Wouldn't there be only one branch with the target prefix ? Don't we want to compare all branches that start with the release- prefix and are less than target prefix to be closed ?

.github/workflows/create-branch.yaml
hebelsan: removed `DEV_VERSION` variable and use `NEXT_VERSION` instead:
```
NEXT_VERSION="${MINOR_VERSION}.1"
echo "${NEXT_VERSION}-dev" > VERSION
```

.github/workflows/release-cleanup.yaml
hebelsan: You are right, I removed the milestone deletion from this script since we will automatically close milestones after a release with another worklfow.

PR #1489, opened 2025-09-24
deny dual-stack migration if overlay is not disabled in all cases

DockToFuture:

/area networking
/kind enhancement
/platform aws

Deny dual-stack migration if overlay is not disabled in all cases.
Fixes a bug for old cluster if providerConfig was net set.

Fixes #

**Release note**:

```other operator
Deny dual-stack migration if overlay is not disabled in all cases.
```

### DIFF
--- a/pkg/apis/aws/validation/shoot.go
+++ b/pkg/apis/aws/validation/shoot.go
@@ -47,6 +47,8 @@ func ValidateNetworking(networking *core.Networking, fldPath *field.Path) field.
 					allErrs = append(allErrs, field.Invalid(fldPath.Child("providerConfig").Child(overlayKey).Child(enabledKey), enabled, "overlay is not supported in conjunction with IPv6"))
 				}
 			}
+		} else {
+			allErrs = append(allErrs, field.NotFound(fldPath.Child("providerConfig").Child(overlayKey).Child(enabledKey), false))
 		}
 	}
 


pkg/admission/validator/shoot_test.go
```diff
@@ -335,6 +338,9 @@ var _ = Describe("Shoot validator", func() {
 				c.EXPECT().Get(ctx, cloudProfileKey, &gardencorev1beta1.CloudProfile{}).SetArg(2, *cloudProfile)
 
 				shoot.Spec.Networking.IPFamilies = []core.IPFamily{core.IPFamilyIPv6, core.IPFamilyIPv4}
+				shoot.Spec.Networking.ProviderConfig = &runtime.RawExtension{
+					Raw: []byte(`{"overlay":{"enabled":false}}`),
+				}
```
ScheererJ, line 343: Should we add more tests checking that the validation yields errors if overlay is enabled regardless if it is done via `providerConfig` or implicitly?

ScheererJ (dismissed):
/lgtm

pkg/apis/aws/validation/shoot.go
```diff
@@ -47,6 +47,8 @@ func ValidateNetworking(networking *core.Networking, fldPath *field.Path) field.
 					allErrs = append(allErrs, field.Invalid(fldPath.Child("providerConfig").Child(overlayKey).Child(enabledKey), enabled, "overlay is not supported in conjunction with IPv6"))
 				}
 			}
+		} else {
```
hebelsan, line 50: From a logical perspective the code looks good.
That said, the section starting with the IPv6 check is a bit difficult to follow due to the nested `if/else` blocks. You could simplify it by returning early when IPv6 isn’t enabled, or by moving the IPv6-specific logic into a dedicated helper (e.g. `validateIPv6`). This would make the main function easier to read and maintain.

hebelsan (changes requested):

hebelsan (approved):
/lgtm

PR #1496, opened 2025-10-01
Support flow migration of shoots with duplicated zone

hebelsan:

/area robustness
/kind enhancement
/platform aws

This PR enables the migration of shoots with duplicated availability zone entries from Terraform to Flow. For each availability zone, only the first entry will be migrated and managed by the Flow reconciler.

Duplicated zone entries are not supported because the Cloud Controller Manager does not allow multiple subnets for a single zone.

Before migrating these shoots, the shoot tag must be removed from all duplicated subnets and private route tables to prevent the Flow reconciler from crashing.

After the migration, ELBv2 load balancers can be moved into the correct subnets, and all unused AWS resources associated with the duplicated subnets can be safely deleted.

Fixes #

**Release note**:

```feature operator
Support migration of shoots with duplicated zone entries from terraform to flow
```

### DIFF
--- a/pkg/controller/infrastructure/flow_reconciler.go
+++ b/pkg/controller/infrastructure/flow_reconciler.go
@@ -333,7 +333,14 @@ func migrateTerraformStateToFlowState(rawExtension *runtime.RawExtension, zones 
 
 	tfNamePrefixes := []string{"nodes_", "private_utility_", "public_utility_"}
 	flowNames := []string{infraflow.IdentifierZoneSubnetWorkers, infraflow.IdentifierZoneSubnetPrivate, infraflow.IdentifierZoneSubnetPublic}
+
+	processedZones := make(map[string]bool)
 	for i, zone := range zones {
+		if processedZones[zone.Name] {
+			continue
+		}
+		processedZones[zone.Name] = true
+
 		keyPrefix := infraflow.ChildIdZones + shared.Separator + zone.Name + shared.Separator
 		suffix := fmt.Sprintf("z%d", i)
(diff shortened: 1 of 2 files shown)


pkg/controller/infrastructure/flow_reconciler.go
```diff
@@ -333,7 +333,15 @@ func migrateTerraformStateToFlowState(rawExtension *runtime.RawExtension, zones
...
 	tfNamePrefixes := []string{"nodes_", "private_utility_", "public_utility_"}
 	flowNames := []string{infraflow.IdentifierZoneSubnetWorkers, infraflow.IdentifierZoneSubnetPrivate, infraflow.IdentifierZoneSubnetPublic}
+
+	// TODO: @hebelsan - remove processedZones after migration of shoots with duplicated zone name entries
+	processedZones := make(map[string]bool)
 	for i, zone := range zones {
+		if processedZones[zone.Name] {
+			continue
+		}
+		processedZones[zone.Name] = true
```
kon-angelo, line 343: can also be done with a string set

pkg/controller/infrastructure/infraflow/reconcile.go
```diff
@@ -571,7 +571,14 @@ func (c *FlowContext) ensureNodesSecurityGroup(ctx context.Context) error {
 		},
 	}
 
+	// TODO: @hebelsan - remove processedZones after migration of shoots with duplicated zone name entries
+	processedZones := make(map[string]bool)
 	for index, zone := range c.config.Networks.Zones {
+		if processedZones[zone.Name] {
+			continue
+		}
+		processedZones[zone.Name] = true
```
kon-angelo, line 580: same as above

PR #1513, opened 2025-10-13
Add support for Capacity Reservations

### LINKED ISSUE #1463: Support targeting of capacity reservations

/area control-plane
/kind enhancement
/platform aws

Support of targeting AWS capacity reservations, similar to the MCM for aws: https://github.com/gardener/machine-controller-manager-provider-aws/blob/7c3cd1f68ae26b3a7cd9f598aee2451729c6df43/kubernetes/machine-class.yaml#L28-L31

**Why is this needed**:
This would allow for individual reservations for different worker pools, more fine-granular control on reservation usage.

AndreasBurger:

/area usability
/kind enhancement
/platform aws

Adds the required fields and adapts MachineClass to use Capacity Reservation with MCM-AWS

Fixes #1463

**Release note**:

```feature user
Shoot worker nodes can now take advantage of AWS Capacity Reservations
```


AndreasBurger:
Un-drafting, the final issues here are with the tests (the actual values diverge from the expected because I only updated one set of those).

docs/usage/usage.md
```diff
@@ -811,6 +811,42 @@ spec:
       enabled: false
 ```
 
+## Use EC2 Capacity Reservations
+
+To have your workers launch into an existing Capacity Reservations, you need to add configuration about the Capacity Reservation into the workers provider-config:
```
wpross: into existing, without `an`

pkg/apis/aws/validation/filter.go
```diff
@@ -30,15 +30,21 @@ var (
 	ZoneNameRegex = `^[a-z0-9-]+$`
 	// TagKeyRegex matches Letters (a–z, A–Z), numbers (0–9), spaces, and the following symbols: + - = . _ : / @
 	TagKeyRegex = `^[\w +\-=\.:/@]+$`
+	// CapacityReservationIDRegex matches IDs of Capacity Reservations. Example: // cr-1234abcd56example
+	CapacityReservationIDRegex = `^cr-[a-z0-9]+$`
+	// CapacityReservationGroupRegex matches resource-group ARNs, e.g.: arn:aws:resource-groups:eu-west-2:123456789012:group/example-cr-group
```
wpross: same here:
... ARNs e.g. arn:...

pkg/apis/aws/validation/worker.go
```diff
@@ -95,6 +96,59 @@ func ValidateWorkerConfig(workerConfig *apisaws.WorkerConfig, volume *core.Volum
 		}
 	}
 
+	if workerConfig.CapacityReservation != nil {
+		childPath := fldPath.Child("capacityReservationOptions")
```
wpross: why `Options` in `capacityReservationOptions`?

pkg/apis/aws/v1alpha1/types_worker.go
```diff
@@ -155,3 +157,13 @@ type CpuOptions struct {
 	// ThreadsPerCore sets the number of threads per core. Must be either '1' (disable multi-threading) or '2'.
 	ThreadsPerCore *int64 `json:"threadsPerCore"`
 }
+
+// CapacityReservation contains configuration about the Capacity Reservation to use for the instance.
+type CapacityReservation struct {
+	// CapacityReservationPreference defines the instance's reservation preferences.
+	CapacityReservationPreference *string `json:"capacityReservationPreference,omitempty"`
```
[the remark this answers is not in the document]
AndreasBurger: Yeah, could work, I think. But as far as I understand Go "enums", this would actually provide no benefit other than necessitating more imports. I'd rather keep this exposed type as simple as possible.

Also I'm not sure if it is considered best practice for k8s to restrict oneself to strings for these versioned-type definitions, I haven't seen an example of using complex types here directly. There _may_ be a reason for that that I am unaware of.

PR #1545, opened 2025-10-29
VirtualCapacity now updated in MachineClass's NodeTemplate without change in MachineClass name

### LINKED ISSUE #1238: Support "Virtual"/"Pseudo" Extended Resources on a worker pool that do not trigger rollout

/area control-plane
/kind enhancement
/platform aws

"Virtual" or "Pseudo" extended resources specified on a worker pool - which do not represent "real" resources - should not cause a rollout of the machines of the worker pool.

### Option-1 (Naming Convention, No shoot API Change)
A "pseudo" or "virtual" resource can be disambiguated with the domain name containing the word "pseudo"/"virtual"

Example: `virtual.operator.namespace.com/memory: "2049008Ki"`

### Option-2 (Explicit delineation, Requires shoot API Change)

The `providerConfig.NodeTemplate` exposes an additional map that specify *virtual* resources of a Node that don't require rollout as opposed to *real* provisionable resources

```yaml
providerConfig:
    nodeTemplate:
        capacity:
            cpu: 123
        virtualCapacity: #PROPOSED API CHANGE
             hc.hana.com/memory: 1234567
```

**Why is this needed**:

Currently, specifying an extended resources for a worker pool in the shoot specification triggers rollout of all nodes of that worker pool.  This is not semantically in-correct since extended resources in the general case are used to represent provisionable "real" resources which should necessitate a rollout.  However, "virtual/pseudo" resources that are that represent some abstract/computational property of the Node should not trigger a rollout after the worker pool is updated.

elankath:

/area auto-scaling
/kind enhancement
/platform aws

Read #1238
After updating to latest MCM release, we need to be able to set the `MachineClass.NodeTemplate.VirtualCapacity` from the `ProviderConfig` and do so without changing the machine class hash. 

Fixes #1238

- Please check whether we need to change `ComputeAdditionalHashDataV1` also for old shoots ?

**Release note**:

```feature operator
providerConfig.nodeTemplate.virtualCapacity mapped to MachineClass NodeTemplate without triggering rollout
```

### DIFF
--- a/pkg/controller/worker/machines_test.go
+++ b/pkg/controller/worker/machines_test.go
@@ -1154,15 +1154,16 @@ var _ = Describe("Machines", func() {
 					ephemeralStorageQuant := resource.MustParse("30Gi")
 					dongleName := corev1.ResourceName("resources.com/dongle")
 					dongleQuant := resource.MustParse("4")
-					virtalResourceName := corev1.ResourceName("subdomain.domain.com/virtual-resource-name")
+					virtualResourceName := corev1.ResourceName("subdomain.domain.com/virtual-resource-name")
 					virtualResourceQuant := resource.MustParse("1024")
 					customResources := corev1.ResourceList{
 						corev1.ResourceEphemeralStorage: ephemeralStorageQuant,
 						dongleName:                      dongleQuant,
 					}
 					customVirtualResources := corev1.ResourceList{
-						virtalResourceName: virtualResourceQuant,
+						virtualResourceName: virtualResourceQuant,
 					}
+
 					w.Spec.Pools[0].ProviderConfig = &runtime.RawExtension{
 						Raw: encode(&api.WorkerConfig{
 							NodeTemplate: &extensionsv1alpha1.NodeTemplate{
@@ -1171,9 +1172,8 @@ var _ = Describe("Machines", func() {
 							},
 						}),
 					}
-
-					expectedCapacity := w.Spec.Pools[0].NodeTemplate.Capacity.DeepCopy()
-					maps.Copy(expectedCapacity, customResources)
+					expectedNodeTemplateCapacity := w.Spec.Pools[0].NodeTemplate.Capacity.DeepCopy()
+					maps.Copy(expectedNodeTemplateCapacity, customResources)
 
 					wd, err := NewWorkerDelegate(c, decoder, scheme, chartApplier, "", w, cluster)
 					Expect(err).NotTo(HaveOccurred())
@@ -1187,7 +1187,7 @@ var _ = Describe("Machines", func() {
 						if strings.Contains(className, namePool1) {
 							GinkgoWriter.Printf("Machine class name: %q\n", className)
 							nt := mClz["nodeTemplate"].(machinev1alpha1.NodeTemplate)
-							Expect(nt.Capacity).To(Equal(expectedCapacity))
+							Expect(nt.Capacity).To(Equal(expectedNodeTemplateCapacity))
 							Expect(nt.VirtualCapacity).To(Equal(customVirtualResources))
 						}
 					}
@@ -1196,7 +1196,6 @@ var _ = Describe("Machines", func() {
 			It("should generate machine classes with same name even when virtualCapacity is newly added or changed", Label("machineClass", "virtualCapacity"), func() {
 				capacityResources := corev1.ResourceList{
 					corev1.ResourceCPU:              resource.MustParse("1"),
-					"gpu":                           resource.MustParse("1"),
 					corev1.ResourceMemory:           resource.MustParse("1Gi"),
 					corev1.ResourceEphemeralStorage: resource.MustParse("10Gi"),
 				}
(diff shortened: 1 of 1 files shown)


pkg/controller/worker/machines.go
```diff
@@ -419,16 +429,82 @@ func ComputeAdditionalHashDataV1(pool extensionsv1alpha1.WorkerPool) []string {
...
-
-	// in the future, we may not calculate a hash for the whole ProviderConfig
-	// for example volume IOPS changes could be done in place
+	if pool.KubernetesVersion != nil && *pool.KubernetesVersion != "" {
+		poolVersion, err := version.Parse(*pool.KubernetesVersion)
+		if err != nil {
+			return nil, err
+		}
+		if poolVersion.AtLeast(version.MustParse("1.33.4")) && workerConfig != nil {
+			// new hash data only for newer clusters.
```
kon-angelo:
```
	k8sVersion, err := semver.NewVersion(pool.KubernetesVersioin)
	if err != nil {
		return err
	}
	if versionutils.ConstraintK8sGreaterEqual134.Check(k8sVersion) {
...

```

pkg/controller/worker/machines.go
```diff
@@ -419,16 +429,82 @@ func ComputeAdditionalHashDataV1(pool extensionsv1alpha1.WorkerPool) []string {
...
+// ComputeWorkerConfigHashData computes hash data strings for the given AWS WorkerConfig aka ProviderConfig
+func ComputeWorkerConfigHashData(workerConfig *awsapi.WorkerConfig) []string {
+	var hashData []string
+	if workerConfig.NodeTemplate != nil && workerConfig.NodeTemplate.Capacity != nil {
+		for n, q := range workerConfig.NodeTemplate.Capacity {
+			// Capacity is part of the hash data, VirtualCapacity is not
+			hashData = append(hashData, fmt.Sprintf("%s:%d", n, q.Value()))
+		}
+	}
+	if workerConfig.Volume != nil && workerConfig.Volume.IOPS != nil {
```
hebelsan: what about `volume.throughput` shouldn't it also be added to the hash calculation?

pkg/controller/worker/machines.go
```diff
@@ -419,16 +429,82 @@ func ComputeAdditionalHashDataV1(pool extensionsv1alpha1.WorkerPool) []string {
...
+			return nil, err
+		}
+		if poolVersion.AtLeast(version.MustParse("1.33.4")) && workerConfig != nil {
+			// new hash data only for newer clusters.
+			additionalData = append(additionalData, ComputeWorkerConfigHashData(workerConfig)...)
+			return additionalData, nil
+		}
+	}
+	// keep old hash for existing clusters to avoid triggering rollout
 	if pool.ProviderConfig != nil && pool.ProviderConfig.Raw != nil {
```
hebelsan: Can we add a TODO comment to remove this once the gating kubernets version from above gets removed?

pkg/controller/worker/machines.go
```diff
@@ -419,16 +429,82 @@ func ComputeAdditionalHashDataV1(pool extensionsv1alpha1.WorkerPool) []string {
...
 	var additionalData = ComputeAdditionalHashDataV1(pool)
-
-	// in the future, we may not calculate a hash for the whole ProviderConfig
-	// for example volume IOPS changes could be done in place
+	if pool.KubernetesVersion != nil && *pool.KubernetesVersion != "" {
+		poolVersion, err := version.Parse(*pool.KubernetesVersion)
+		if err != nil {
+			return nil, err
+		}
+		if poolVersion.AtLeast(version.MustParse("1.33.4")) && workerConfig != nil {
```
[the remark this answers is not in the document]
elankath: @hebelsan I am actually not sure about the best approach to take here - what do you recommend ?

pkg/controller/worker/machines.go
elankath: @hebelsan  OK, I checked and found out that EC2  permits dynamically changing the throughput and IOPS (Input/Output Operations Per Second) of an Amazon EBS volume attached to a running EC2 instance. Should I remove both throughput and IOPS from the hash or keep in both ? 

We will need to add support in the MCM for this though - we currently don't have this support. But after the `UpdateMachine` PR is merged (WIP by Andreas Fritzler),  we can technically add support for this in the MCM. 

So for now, I guess add both in and remove them from hash calculation later if someone wants to dynamically change IOPS and throughput ?

elankath: Now, simplified this, incorporating approach from the HANA meeting. `ComputeAdditionalHashDataV2` has been very simplified. (slightly hacky but this is what was suggested). Unit test for `VirtualCapacity` addition/change now enforces that there is zero machine class name change with addition or change of `VirtualCapacity` in any existing `ProviderConfig.NodeTemplate`. 

It was decided that the proper and smart approach to computing hash data in `ComputeAdditionalHashDataV2` can be introduced when we have the MCM support Hot-Update of Machines with change in the `ProviderConfig`. Until that happens, we stick to existing hash logic and only use a workaround for special cases like `VirtualCapacity`

pkg/controller/worker/machines.go
elankath: the logic regarding version stuff has been removed. Now we have a `TODO` like

```
TODO: once the MCM supports Machine Hot-Update from the WorkerConfig, this hash data logic can be made smarter
```

PR #1552, opened 2025-11-04
Fix tf subnets status

hebelsan:

/area quality
/kind enhancement
/platform aws

In case of duplicated zones multiple subnets with the same zone are written into the status.
This can lead to situations where after a reconcile a different subnet is used for the nodes because only one subnet per zone can be used in the machineclass.
With this PR we always use the the first zone entry as subnet in the status.

Fixes #

**Release note**:

```bugfix operator
Always use the same node subnet in case of duplicated zones
```

### DIFF
--- a/pkg/controller/infrastructure/tf_reconciler.go
+++ b/pkg/controller/infrastructure/tf_reconciler.go
@@ -587,7 +587,7 @@ func computeProviderStatusSubnets(logger logr.Logger, infrastructure *api.Infras
 			return s.Zone == zone && s.Purpose == purpose
 		})
 		if zoneExists {
-			logger.Info("Skipping subnet addition to status as zones is already present", "zone", zone, "purpose", purpose)
+			logger.Info("Skipping subnet addition to status as zone is already present", "zone", zone, "purpose", purpose)
 			continue
 		}
 


pkg/controller/infrastructure/tf_reconciler.go
```diff
@@ -578,11 +578,26 @@ func computeProviderStatusSubnets(infrastructure *api.InfrastructureConfig, valu
...
-		})
+
+		if zoneID < 0 || zoneID >= len(infrastructure.Networks.Zones) {
+			return nil, fmt.Errorf("zone index %d out of range", zoneID)
+		}
+		zone := infrastructure.Networks.Zones[zoneID].Name
+
+		exists := false
+		for _, subnet := range subnetsToReturn {
+			if subnet.Zone == zone && subnet.Purpose == purpose {
```
kon-angelo: probably better written with a `slices.ContainsFunc`

kon-angelo (changes requested):

pkg/controller/infrastructure/tf_reconciler.go
```diff
@@ -578,11 +578,26 @@ func computeProviderStatusSubnets(infrastructure *api.InfrastructureConfig, valu
...
+				exists = true
```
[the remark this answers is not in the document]
kon-angelo: I would agree with a log message. Otherwise the plan is to silently ignore them for now.

pkg/controller/infrastructure/tf_reconciler.go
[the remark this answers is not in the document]
kon-angelo: It should be:

https://github.com/gardener/gardener-extension-provider-aws/blob/6bffe9ce319884dd8c4115ee26be2ebf5efe918f/pkg/controller/infrastructure/tf_reconciler.go#L493

Is a slice of maps values, but it is a slice so order should be guaranteed. If it was not then TF would randomly destroy subnets because the name (by virtue of random ordering) changed.

pkg/controller/infrastructure/tf_reconciler.go
```diff
@@ -556,32 +556,51 @@ func generateTerraformInfraConfig(ctx context.Context, infrastructure *extension
...
 
-	for key, value := range values {
+	// iterate over keys in lexicographic order for deterministic ordering
+	keys := make([]string, 0, len(values))
+	for k := range values {
+		keys = append(keys, k)
+	}
+	slices.Sort(keys)
+
+	for _, key := range keys {
```
kon-angelo: can be replaced with `	for _, key := range slices.Sorted(maps.Keys(values)) {
`

kon-angelo (approved):
lgtm

pkg/controller/infrastructure/tf_reconciler.go
```diff
@@ -556,32 +557,44 @@ func generateTerraformInfraConfig(ctx context.Context, infrastructure *extension
...
+		if zoneID < 0 || zoneID >= len(infrastructure.Networks.Zones) {
+			return nil, fmt.Errorf("zone index %d out of range", zoneID)
+		}
+		zone := infrastructure.Networks.Zones[zoneID].Name
+
+		zoneExists := slices.ContainsFunc(subnetsToReturn, func(s v1alpha1.Subnet) bool {
+			return s.Zone == zone && s.Purpose == purpose
+		})
+		if zoneExists {
+			logger.Info("Skipping subnet addition to status as zones is already present", "zone", zone, "purpose", purpose)
```
wpross: nit: `zone` instead of `zones`.

PR #1573, opened 2025-11-13
Allow to specify an own IPv6 IPAM pool

### LINKED ISSUE #1048: dual stack ingress: support IPAM pools
/area networking
/kind enhancement
/platform aws

Support static IPv6 addresses for dualstack LBs. To achieve this, gardener may support allocation of IPv6 CIDRs to the VPC and public subnet from a user-managed IPAM pool. 

**Why is this needed**:
When a user-managed IPAM pool is used, the LB IPv6 addresses are allocated from a stable range that is decoupled from the cluster's lifecycle (not released in case of VPC deletion). Users may also select specific addresses from within these ranges, using the following annotation: `service.beta.kubernetes.io/aws-load-balancer-ipv6-addresses`. These ranges/addresses can be communicated to consumers for allowlisting purposes.

axel7born:

/area networking
/kind enhancement
/platform aws

Fixes #1048

**Release note**:

```feature user
It's possible to specify an own IPv6 IPAM pool now.
```

### DIFF
--- a/pkg/apis/aws/types_infrastructure.go
+++ b/pkg/apis/aws/types_infrastructure.go
@@ -116,7 +116,7 @@ type VPC struct {
 	// GatewayEndpoints service names to configure as gateway endpoints in the VPC.
 	GatewayEndpoints []string
 	// Ipv6IpamPool references an AWS IPv6 IPAM pool used to allocate the VPC's IPv6 CIDR block.
-	Ipv6IpamPool IPAMPool
+	Ipv6IpamPool *IPAMPool
 }
 
 // IPAMPool references an AWS IPAM pool that should be used to allocate the VPC's CIDR.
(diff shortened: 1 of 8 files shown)


pkg/apis/aws/types_infrastructure.go
```diff
@@ -115,6 +115,14 @@ type VPC struct {
 	CIDR *string
 	// GatewayEndpoints service names to configure as gateway endpoints in the VPC.
 	GatewayEndpoints []string
+	// Ipv6IpamPool references an AWS IPv6 IPAM pool used to allocate the VPC's IPv6 CIDR block.
+	Ipv6IpamPool IPAMPool
```
hebelsan: We have to decide here: Ipv6IpamPool is a value (not pointer). Future extension (additional optional fields) will make it harder to distinguish "unset" from "empty".

axel7born: Changed to pointer.

PR #1581, opened 2025-11-20
[GEP-28] Use technicalID from `Cluster` instead of `Worker.Namespace`

timebertt:

/area ipcei
/kind enhancement
/platform aws

To support self-hosted shoots with managed infrastructure, the `Worker` controller/delegate needs to use the technical ID from `Cluster.shoot.status.technicalID` for prefixing the names of machine-related objects. The `Worker` namespace is `kube-system` for self-hosted shoots.
This PR is similar to [gardener/gardener@`22f4295` (#13485)](https://github.com/gardener/gardener/pull/13485/commits/22f429593ba87fb08c3a37fb0ce3c1da914c0f21).

Part of https://github.com/gardener/gardener/pull/13485

**Release note**:

```feature operator
The `Worker` controller is prepared to support self-hosted shoot clusters with managed infrastructure (see [GEP-28](https://github.com/gardener/gardener/blob/master/docs/proposals/28-self-hosted-shoot-clusters.md#managed-infrastructure)).
```

### DIFF
--- a/pkg/controller/worker/machines.go
+++ b/pkg/controller/worker/machines.go
@@ -166,8 +166,8 @@ func (w *WorkerDelegate) generateMachineConfig(ctx context.Context) error {
 				},
 				"tags": utils.MergeStringMaps(
 					map[string]string{
-						fmt.Sprintf("kubernetes.io/cluster/%s", w.worker.Namespace): "1",
-						"kubernetes.io/role/node":                                   "1",
+						fmt.Sprintf("kubernetes.io/cluster/%s", w.cluster.Shoot.Status.TechnicalID): "1",
+						"kubernetes.io/role/node": "1",
 					},
 					pool.Labels,
 				),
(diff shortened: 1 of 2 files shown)


pkg/controller/bastion/options.go
```diff
@@ -47,7 +47,7 @@ type Options struct {
 
 // NewBaseOpts determines base opts that are required for creating and deleting a Bastion on AWS.
 func NewBaseOpts(ctx context.Context, bastion *extensionsv1alpha1.Bastion, cluster *controller.Cluster, awsClient *awsclient.Client) (BaseOptions, error) {
-	name := cluster.ObjectMeta.Name
+	name := cluster.Shoot.Status.TechnicalID
```
hebelsan: Shouldn't we adapt this bastion code change also in the other provider PRs (see: [azure](https://github.com/gardener/gardener-extension-provider-azure/pull/1378), [gcp](https://github.com/gardener/gardener-extension-provider-gcp/pull/1239), [open-stack](https://github.com/gardener/gardener-extension-provider-openstack/pull/1211))?

timebertt: You're right. We should adapt the other provider extensions as well in a similar manner. 
However, I figured, that we also need to adapt the other controllers (`Infrastructure` and `ControlPlane`) in some places. 

Hence, I dropped the `Bastion` commit in this PR to focus on the `Worker` controller for now – as done in the other extensions.
We can have dedicated PRs for adapting the remaining controllers in all extensions.

PR #1585, opened 2025-11-24
Switch VPAs to control memory only

voelzmo:

/area auto-scaling
/kind enhancement
/platform aws

Switch VPAs to control memory only, such that we don't get any recommendations and evictions for CPU changes. VPA can only recommend values > =10m Cores, so we're not really interested in changes for these low-load components.

Fixes #

Seems like this was missed in https://github.com/gardener/gardener-extension-provider-aws/pull/1448

**Release note**:

```other operator
Switch VPAs to control memory only
```


hebelsan (approved):
/lgtm

PR #1589, opened 2025-11-28
fix: unchanged worker pool hash and hot-update with virtualCapacity

### LINKED ISSUE #1586: Nodes are rolled for a workerGroup change to only providerConfig.nodeTemplate.virtualCapacity
/area auto-scaling
/kind bug
/platform aws

**What happened**:
I added a non-empty `virtualCapacity` to a `providerConfig.nodeTemplate` of a `workerGroup` definition and an existing node of that `workerGroup` was rolled.

That existing nodes are not rolled by any changes to `virtualCapacity`.

Have a node for a given workerGroup deployed, then update its workerGroup configuration with a change to only `providerConfig.nodeTemplate.virtualCapacity` and observe, if the existing node is replaced by a new node.

**Environment**:

- Gardener version (if relevant): 1.132.1
- Extension version: 1.66.0
- Kubernetes version (use `kubectl version`): 1.32.7
- Cloud provider or hardware configuration: AWS
- Others:

elankath:

/area auto-scaling
/kind bug
/platform aws

Covered in #1586 .
This PR also now addresses #1598 for k8s versions > 1.34 as requested by maintainers

We need to ensure that no rollout occurs for `virtualCapacity` add/removal with both the cases of no previous provider config and also with previous provider config present and `virtualCapacity` values are correctly hot-updated into the `MachineClass`'s `NodeTemplate` (and thus consequently hot-updated into the `Node.Status.Capacity` by the `machine-controller`.

Fixes #1586, #1598

**Release note**:

```bugfix operator
no rollout hot-update of ProviderConfig.NodeTemplate.VirtualCapacity with/without already existing ProviderConfig.
new hash strategy adopted for ProviderConfig for k8s versions >= 1.34
```

### DIFF
--- a/pkg/controller/worker/machines_test.go
+++ b/pkg/controller/worker/machines_test.go
@@ -1512,6 +1512,7 @@ var _ = Describe("Machines", func() {
 							"cpu=8",
 							"gpu=1",
 							"memory=137438953472",
+							"name1",
 							"arn",
 							"required",
 							"1",


pkg/controller/worker/helper.go
```diff
@@ -46,3 +47,39 @@ func (w *WorkerDelegate) updateWorkerProviderStatus(ctx context.Context, workerS
...
+		workerConfigCopy.NodeTemplate.VirtualCapacity = nil
+	}
+
+	if workerConfigCopy.NodeTemplate != nil && workerConfigCopy.NodeTemplate.Capacity == nil {
+		// Need the same hash if WorkerConfig was present, but nodeTemplate was NOT set and subsequently nodeTemplate.virtualCapacity was just added.
+		workerConfigCopy.NodeTemplate = nil
+	}
+
+	// Step 2: wrap and inject apiVersion & kind
+	// needs an explicit set of APIVersion and Kind in exact order so we don't to differ from the previous `string(pool.ProviderConfig.Raw)`
```
hebelsan:
```suggestion
	// needs an explicit set of APIVersion and Kind in exact order so we don't differ from the previous `string(pool.ProviderConfig.Raw)`
```

hebelsan (changes requested):
Thanks for the update!

elankath:
@hebelsan  rebased to latest master and adopted review comments

charts/internal/seed-controlplane/charts/csi-driver-controller/templates/csi-driver-controller-deployment.yaml
```diff
@@ -136,6 +136,7 @@ spec:
         - --kube-api-burst=200
         - --worker-threads=100
         - --retry-interval-max=10m
+        - --strict-topology
```
kon-angelo: this doesn't seem to belong here right ?

PR #1635, opened 2026-01-09
Add health check condition `RouteControllerActive` to control plane

DockToFuture:

/area networking
/kind enhancement
/platform aws

This PR adds a new health check condition `RouteControllerActive` that monitors whether the AWS custom route controller is actively running in the control plane. RouteControllerActive (new): Returns `False` when the route controller is disabled (replicas=0) and `True` only when it's actively running with replicas > 0. 

The health check condition looks like:

```
  {
    "lastTransitionTime": "2026-01-09T14:10:08Z",
    "lastUpdateTime": "2026-01-09T14:10:08Z",
    "message": "All health checks successful",
    "reason": "HealthCheckSuccessful",
    "status": "True",
    "type": "RouteControllerActive"
  }
```

The `RouteControllerActive` condition is required by the [calico extension](https://github.com/gardener/gardener-extension-networking-calico) to know when a transition from overlay network to native routing can happen seamlessly.
For more information see: https://github.com/gardener/gardener-extension-networking-calico/pull/765
Fixes #

**Release note**:

```other operator
Add a new health check condition `RouteControllerActive` that monitors whether the AWS custom route controller is actively running in the control plane.
```

### DIFF
--- a/pkg/controller/healthcheck/add.go
+++ b/pkg/controller/healthcheck/add.go
@@ -69,6 +69,11 @@ func RegisterHealthChecks(_ context.Context, mgr manager.Manager, opts healthche
 				HealthCheck:   newCustomRouteControllerHealthCheck(general.NewSeedDeploymentHealthChecker(aws.AWSCustomRouteControllerName)),
 				// no precheck needed, as the deployment is always created (with replicas=0 if not enabled, see valuesprovider.go)
 			},
+			{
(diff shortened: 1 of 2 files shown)


kon-angelo (commented):
This change would help you detect that RC is enabled, but not when it has finished setting up routes. That would lead to downtime right ? Or do you see this as non-issue

PR #1636, opened 2026-01-12
Improve routing table association management

hebelsan:

/area robustness
/kind enhancement
/platform aws

This PR improves the routing table association management by checking if the route table association stored in the state really exist in AWS.
Otherwise it deletes the entry from the state.
To perform this action a new client function `GetRouteTableAssociationIDs` was added.

Fixes #

**Release note**:

```other operator
Improve routing table association management
```

### DIFF
--- a/pkg/controller/infrastructure/infraflow/reconcile.go
+++ b/pkg/controller/infrastructure/infraflow/reconcile.go
@@ -1476,7 +1476,7 @@ func (c *FlowContext) deletePrivateRoutingTable(zoneName string) flow.TaskFn {
 	}
 }
 
-func (c *FlowContext) routingAssocSpecs() []routeTableAssociationSpec {
+func (c *FlowContext) routingAssociationSpecs() []routeTableAssociationSpec {
 	return []routeTableAssociationSpec{
 		{IdentifierZoneSubnetPublic, IdentifierZoneSubnetPublicRouteTableAssoc, false},
 		{IdentifierZoneSubnetPrivate, IdentifierZoneSubnetPrivateRouteTableAssoc, true},
@@ -1526,7 +1526,7 @@ func (c *FlowContext) validateAndPruneRoutingTableAssocState(ctx context.Context
 
 func (c *FlowContext) ensureRoutingTableAssociations(zoneName string) flow.TaskFn {
 	return func(ctx context.Context) error {
-		specs := c.routingAssocSpecs()
+		specs := c.routingAssociationSpecs()
 
 		err := c.validateAndPruneRoutingTableAssocState(ctx, zoneName, specs)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/controller/infrastructure/infraflow/reconcile.go
```diff
@@ -1476,18 +1476,74 @@ func (c *FlowContext) deletePrivateRoutingTable(zoneName string) flow.TaskFn {
...
+	shouldValidatePresence := false
+	for _, spec := range specs {
+		if assocID := child.Get(spec.assocKey); assocID != nil {
+			shouldValidatePresence = true
+			break
+		}
+	}
+	if !shouldValidatePresence {
+		return nil
+	}
```
AndreasBurger: Consider moving this to a helper function, but the only thing this would effectively do is shorten this func a _little_ bit and get rid of the break. Not sure if its worth it, to be honest.

AndreasBurger (dismissed):
lgtm

pkg/controller/infrastructure/infraflow/reconcile.go
hebelsan: I like your idea, this will improve the readability and reduce complexity in reconcile.go.
I've extracted the functionality into a helper function in utils.go.

AndreasBurger (approved):

pkg/aws/client/client.go
```diff
@@ -2085,6 +2086,36 @@ func (c *Client) DeleteRouteTableAssociation(ctx context.Context, associationId
...
+	if ignoreNotFound(err) != nil {
+		return nil, err
+	}
+
+	for _, routeTable := range routeTablesOutput.RouteTables {
+		for _, assoc := range routeTable.Associations {
+			if len(subnetIDs) == 0 {
+				associationIDs = append(associationIDs, aws.ToString(assoc.RouteTableAssociationId))
+			} else if assoc.SubnetId != nil && slices.Contains(subnetIDs, *assoc.SubnetId) {
+				associationIDs = append(associationIDs, aws.ToString(assoc.RouteTableAssociationId))
```
kon-angelo: Is it not possible to have the filtering done server-side ?

pkg/controller/infrastructure/infraflow/reconcile.go
```diff
@@ -1479,18 +1476,67 @@ func (c *FlowContext) deletePrivateRoutingTable(zoneName string) flow.TaskFn {
...
+}
+
+// validateAndPruneRoutingTableAssocState checks whether the routing table associations stored in the state
+// still exist in AWS. If not, it removes them from the state.
+func (c *FlowContext) validateAndPruneRoutingTableAssocState(ctx context.Context, zoneName string, specs []routeTableAssociationSpec) error {
+	child := c.getSubnetZoneChild(zoneName)
+
+	// should validate only if at least one association ID is present in state
+	if !hasRouteTableAssociationInState(func(assocKey string) *string { return child.Get(assocKey) }, specs) {
```
kon-angelo: does this need to be wrapped again in a func ? Can't the first argument simply be `child.Get` ?

hebelsan: good catch, thank you!

pkg/controller/infrastructure/infraflow/reconcile.go
```diff
@@ -1479,18 +1476,70 @@ func (c *FlowContext) deletePrivateRoutingTable(zoneName string) flow.TaskFn {
 	}
 }
 
+func (c *FlowContext) routingAssocSpecs() []routeTableAssociationSpec {
```
kon-angelo: let's be consistent at least with top level names. No need to cut the assoc-iation 😉

kon-angelo (commented):
/lgtm

(with the small nit about name cutting)

Issue #1470, opened 2025-09-15
Enhance Security: Add support for AMD SEV-SNP secure enclaves on AWS based Clusters


philipphombergersap:
**How to categorize this issue?**

/area security
/area os
/kind enhancement
/platform aws

**What would you like to be added**:

The ability to configure additional CPU options for EC2 worker nodes beyond the currently supported `coreCount` and `threadsPerCore`. Specifically, we need to be able to set the `AmdSevSnp` option to `enabled` for supported instance types.

This would involve extending the `cpuOptions` field within the `WorkerConfig` of the AWS provider extension (`providerConfig`).

An ideal implementation in the shoot manifest might look like this:

```yaml
...
provider:
  type: aws
  workers:
  - name: cpu-worker-sev
    machine:
      type: m6a.large
      # ...
    providerConfig:
      apiVersion: aws.provider.extensions.gardener.cloud/v1alpha1
      kind: WorkerConfig
      # ...
      cpuOptions:
        coreCount: 2
        threadsPerCore: 1
        amdSevSnp: "enabled" # <-- This is the proposed addition
...
```

This change would allow Gardener to provision worker nodes that are capable of running workloads inside AMD SEV-SNP secure enclaves.

Here some AWS Docs:
https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/sev-snp.html
https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/snp-attestation.html

**Why is this needed**:

We need to run workloads with highly sensitive data (e.g., in-memory crypto keys) securely within Gardener clusters on AWS. Standard VMs do not protect memory from potential access by the cloud provider or a compromised hypervisor.

Hardware-based Confidential Computing technologies like AMD SEV-SNP solve this by providing CPU-enforced memory encryption and isolation. A crucial feature is **remote attestation**, which allows us to cryptographically verify that a workload is running in a genuine secure enclave. This provides verifiable proof of security, rather than forcing us to simply trust the cloud provider.

To use this technology on AWS, a specific CPU option (`AmdSevSnp=enable`) must be set when launching an EC2 instance. The Gardener AWS provider currently does not allow setting this parameter, which completely blocks our ability to leverage these essential hardware security features.

Enabling this option is a necessary first step to support Confidential Computing in Gardener, allowing users to build and run truly secure and attestable applications.

PR #1642, opened 2026-01-14
feature: add support for AMD SEV-SNP in CPU options

hebelsan:

/area quality
/kind enhancement
/platform aws

This PR adds a new options in the `cpuOptions` namely `amdSevSnp` which can be enabled to protect memory from potential access by the cloud provider or a compromised hypervisor.
This feature is only supported though on certain machine types. 

Fixes https://github.com/gardener/gardener-extension-provider-aws/issues/1470

There was an issue in gardenlinux that prevented the validating AMD SEV-SNP: https://github.com/gardenlinux/package-linux/issues/201
I tested AMD SEV-SNP successfully with Gardenlinux 1877.9.0.
For the AMD SEV-SNP attestation follow this guide: https://docs.aws.amazon.com/de_de/AWSEC2/latest/UserGuide/snp-attestation.html.

**Release note**:

```other operator
Add support for AMD SEV-SNP in CPU options
```

### DIFF
--- a/pkg/apis/aws/validation/worker_test.go
+++ b/pkg/apis/aws/validation/worker_test.go
@@ -6,6 +6,7 @@ package validation_test
 
 import (
 	"fmt"
+	"strings"
 
 	ec2types "github.com/aws/aws-sdk-go-v2/service/ec2/types"
 	"github.com/gardener/gardener/pkg/apis/core"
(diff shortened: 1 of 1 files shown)


charts/internal/machineclass/templates/machineclass.yaml
```diff
@@ -73,10 +73,15 @@ providerSpec:
     httpTokens: {{ $machineClass.instanceMetadataOptions.httpTokens }}
     {{- end }}
 {{- end }}
-{{- if $machineClass.cpuOptions }}
+{{- with $machineClass.cpuOptions }}
```
kon-angelo: like this 👍

pkg/apis/aws/v1alpha1/types_worker.go
```diff
@@ -156,6 +156,11 @@ type CpuOptions struct {
 	CoreCount *int64 `json:"coreCount"`
 	// ThreadsPerCore sets the number of threads per core. Must be either '1' (disable multi-threading) or '2'.
 	ThreadsPerCore *int64 `json:"threadsPerCore"`
+	// AmdSevSnp indicates whether AMD SEV-SNP is enabled.
+	// Currently, this option is only supported on M6a, R6a, and C6a instance types.
+	// Valid options are "enabled" and "disabled".
+	// +optional
+	AmdSevSnp *string `json:"amdSevSnp"`
```
AndreasBurger:
```suggestion
	AMDSEVSNP *string `json:"amdSevSnp"`
```
If we want to be consistent. I am not sure we should be here :-|.
Same for the internal repr

hebelsan: I agree that we use uppercase for abbreviations in this file, but writing it in all caps makes it look like a constant...

PR #1696, opened 2026-02-16
Improve opt-in mechanism for dual-stack to avoid automatic load balancer recreation in dual-stack clusters

axel7born:

/area networking
/kind enhancement
/platform aws

Prevents automatic recreation of load balancers when enabling IPv6/dual-stack on existing clusters, giving users control over when to migrate services to dual-stack networking.

Fixes #

**Release note**:

```other operator
Improved opt-in mechanism for dual-stack to avoid automatic load balancer recreation in dual-stack clusters
```


axel7born:
> /approve
> 
> Just a very optional improvement. I think that the dual-stack when enabled will rename the hostname of the loadbalancer. Do we want to have another annotation added in that case to show to the user that there will be an undeleted "invisible" loadbalancer that he should manually clean in this case. Something simple that outlines the old hostname. WDYT?
> 
> If I am wrong and the LB is actually deleted by CCM with the transition to dual-stack, then please ignore.

Yes, that's a good idea! I will add such an annotation.

PR #1715, opened 2026-02-26
Proposal for BYO-subnet

kon-angelo:

/area control-plane
/kind discussion
/platform aws

Fixes #

**Release note**:

```other operator
NONE
```


hebelsan (commented):
Nice job, this is a really solid proposal.
I've added some comments and thoughts.

docs/usage/flexible-network-configuration.md
```diff
@@ -0,0 +1,821 @@
...
+    Public   *string  // +optional, forbidden when WorkersSubnetID is set
+
+    // BYO worker subnet
+    WorkersSubnetID *string  // +optional, mutually exclusive with Workers
+
+    // BYO load balancer subnets (only allowed when WorkersSubnetID is set)
+    // Gardener auto-tags these with the cluster tag and the role tag on reconcile.
+    // On delete, only the cluster tag is removed (the role tag is shared infrastructure).
+    PublicSubnetID   *string  // +optional, BYO only
+    InternalSubnetID *string  // +optional, BYO only
```
rfranzke: Should we create a dedicated structure to make this more clear?

```yaml
existingSubnetIDs:
  workers: ...
  public: ...
  internal: ...
```

rfranzke (commented):
Generally OK.

1. I'm wondering whether you thought about the ops perspective. If we open it up to that degree, A LOT can and will go wrong. How do we ensure (e.g., via automated detection mechanisms) that operators are not distracted by user-caused issues (probably misconfigurations)?
2. Similarly, how do we help end-users to fix and find their issues autonomously to prevent bothering operators with too many demands for help?

docs/proposals/flexible-network-configuration.md
```diff
@@ -0,0 +1,821 @@
...
+      internalSubnetID: subnet-internal-1b
+```
+
+**What Gardener creates:** IAM role, instance profile, SSH key pair only.
+
+**User responsibilities:**
+- Worker subnet routing (NAT/Transit Gateway/VPC endpoints for connectivity)
+- Security group with appropriate rules (see [Security Group Requirements](#security-group-requirements))
+- LB subnets: provide `publicSubnetID`/`internalSubnetID` per zone (recommended — Gardener auto-tags them), or pre-tag subnets manually with `kubernetes.io/role/internal-elb=1`, `kubernetes.io/role/elb=1`, and `kubernetes.io/cluster/<name>=shared`
+- Minimum 8 available IPs per LB subnet; at least 2 AZs for ALBs
```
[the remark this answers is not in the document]
kon-angelo: Not so far. I think the issue with subnet free IPs and worker routing is that we can only verify it by testing: e.g. deploying a bastion to test connectivity.
The LB requirements are necessary if they want to create LBs - and AFAIK the free IPs is an issue every time they would try to create an LB. We can verify the ALB AZs

kon-angelo:
> I'm wondering whether you thought about the ops perspective. If we open it up to that degree, A LOT can and will go wrong. How do we ensure (e.g., via automated detection mechanisms) that operators are not distracted by user-caused issues (probably misconfigurations)?

I agree. I was also sceptical about the implications on SRE teams. However I don't think we can avoid these changes any more. Any large company/LOB will have their own network requirements that currently they fulfil by working around our managed infrastructure. 
If anything, achieving parity with EKS or similar offerings should be our target. 

Also to a certain degree, a BYO-setup is what I would consider a power-user. We still allow and support our minimal managed-infra cluster. If you choose to use this setup because you want to forward traffic through custom NAT devices etc then you should be able as a user to troubleshoot.

> Similarly, how do we help end-users to fix and find their issues autonomously to prevent bothering operators with too many demands for help?

Generally both our SREs and users have "similar" tools to use. The issues that we expect in their majority network issues which we can deal with a couple ways. One could be to use claude skills to automate this. An operator would search through the NSG, Routing rules, test with a bastion etc. That can be codified in a skill. Of course in addition with our reasonable pre-flight checks should check most scenarios.

docs/proposals/flexible-network-configuration.md
```diff
@@ -0,0 +1,821 @@
...
+    Public   *string  // +optional, forbidden when WorkersSubnetID is set
+
+    // BYO worker subnet
+    WorkersSubnetID *string  // +optional, mutually exclusive with Workers
+
+    // BYO load balancer subnets (only allowed when WorkersSubnetID is set)
+    // Gardener auto-tags these with the cluster tag and the role tag on reconcile.
+    // On delete, only the cluster tag is removed (the role tag is shared infrastructure).
+    PublicSubnetID   *string  // +optional, BYO only
+    InternalSubnetID *string  // +optional, BYO only
```
kon-angelo: I find this more confusing for the user than a flat structure actually 🤔. But I am open to it.

hebelsan: For the subnet IDs, I like the idea of introducing a dedicated structure, as it helps avoid confusion with the existing CIDR fields.
Additionally, we could consider making the worker subnet ID a non-pointer, since it is required when working with existing subnets.

docs/proposals/flexible-network-configuration.md
```diff
@@ -0,0 +1,821 @@
...
+type Networks struct {
+    VPC   VPC
+    Zones []Zone
+
+    // NodesSecurityGroupID optionally specifies an existing security group for worker nodes.
+    // When provided, Gardener will not create a nodes security group.
+    // The security group must exist in the same VPC.
+    // Requires VPC.ID to be set.
+    // +optional
+    NodesSecurityGroupID *string
```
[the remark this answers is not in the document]
hebelsan: I’m not sure we need a separate structure for just one field.

docs/proposals/flexible-network-configuration.md
kon-angelo: @rfranzke  and @hebelsan how about a counterproposal to make existing fields aware of the options e.g.:

```
workers: 10.0.0.0/24
```
and
```
workers: subnet-foo
```

be valid. And we will make a distinction in the code. AWS uses specific prefixes for IDs

docs/proposals/flexible-network-configuration.md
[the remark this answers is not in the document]
rfranzke: Sure - I was just thinking, maybe you want to introduce a new configuration option for the nodes security group tomorrow. Now that you touch the API, you could already design it in a future-proof way.

rfranzke:
I'm not arguing against it - I agree, that this flexibility is needed. I was pointing out that we might want to establish ways to not overload SRE teams with complex issue analysis demands from end-users that lever this new flexibility. In other words, how can we protect the SRE teams and push the responsibility to end-users when something goes wrong (at least, initially, until they can show that something Gardener-related is wrong and not their configuration as such)?

docs/proposals/flexible-network-configuration.md
rfranzke: I think that's somewhat against the Kubernetes API conventions. Cramming two distinct semantics into one field makes the API harder to validate, document, and evolve. What happens when a third option appears? 
Generally, we can also live with flat values. How would you design the API in a green field? We can also discuss this and introduce a new API version (`v1alpha2` or `v1beta1`) to make it very clean from the beginning.

kon-angelo: Ok then we avoid the cramming of distinct semantics. The flat values I think are more in line with current user expectations. They are also more future proof in case we allow a sort of mixed setting (which we explicitly exclude for now in this proposal) e.g:

```
  workers: ...
  publicSubnetID: ...
  internalSubnetID: ...
```
I would also like to proceed in the near future with a `v1alpha2` API but I do appreciate that I can use the current `v1alpha1` API to hone the requirements before deciding on the new API. I would have probably gone for something like these are somewhat shallow thoughts for now and I tried to optimize for local API consistency 😅 :
```
type Subnet struct{
id *string // XOR
cidr *string // XOR 
}
```

hebelsan:
I agree that overloading the SRE teams is something we should avoid.

As an initial step, we could enhance the existing error message by including an informational note indicating that this shoot’s infrastructure is not managed by Gardener.

docs/proposals/flexible-network-configuration.md
```diff
@@ -0,0 +1,821 @@
...
+      internalSubnetID: subnet-internal-1a   # auto-tagged by Gardener
+    - name: eu-west-1b
+      workersSubnetID: subnet-workers-1b
+      publicSubnetID: subnet-public-1b
+      internalSubnetID: subnet-internal-1b
+```
+
+**What Gardener creates:** IAM role, instance profile, SSH key pair only.
+
+**User responsibilities:**
```
hebelsan, line 303: Can we additionally to User responsibilities also add a section that gives examples on how to create the resources in AWS for different cluster types.
While testing the setups it took me some time to figure out what I need for a basic byo cluster.
Maybe something along these lines:

```
IPv4 Cluster:
1. Create a new VPC with an IPv4 CIDR block
2. Go to VPC → Edit DNS settings: Enable DNS resolution, Enable DNS hostnames
3. Create worker Subnet in this VPC, choose the correct az-zone (optional create public and internal subnet)
4. Create securityGroup -> inbound: All traffic, (select "Custom" and paste the SG ID you're creating, or do it after saving) outbound: All traffic to 0.0.0.0/0
5. Create  an Internet Gateway and attach it to your vpc
6. Create a NAT gateway
7. Create a internal Route Table for workers and optional internal subnet with route: Destination: 0.0.0.0/0 and Target: your created NAT Gateway. Create a subnet associations with your worker subnet (and optionally with internal subnets)
8. Optionally create a public Route Table (for public LB subnets) with routes: Destination: 0.0.0.0/0 -> Target: your created Internet Gateway. Create a subnet association with your public subnets

IPv6:
1. Create a VPC with IPv6 cidr and DNS settings: Enable DNS resolution + Enable DNS hostnames
2. create an EgressOnlyInternetGateway for the VPC
3. Create worker subnets with a IPv6 CIDR blocks and DNS64 enabled
OPTIONALLY: create subnets with IPv6 CIDR blocks for internal an public subnet. These subnets need some IPv4 IPs since AWS Load Balancers require IPv4
4. Create a private route table for workers and OPTIONAL for internal lb with: 
 --> route ::/0 to the egress only gateway
 --> route 64:ff9b::/96 to the NAT Gateway
5. OPTIONAL create an Internet Gateway and attach it to the VPC (if you want public lb)
6. OPTIONAL create a public route table for public lb:
 --> route ::/0 to internet gateway
 --> route 0.0.0.0/0 to internet gateway
7. Create a SecurityGroup with inbound: All traffic, (select "Custom" and paste the SG ID you're creating, or do it after saving) outbound: All traffic, ::/0
8. Create a Nat Gateway and add it to your VPC

Dualstack:
Same as IPv6 with these differences:
1. Worker subnets should not be IPv6-native — nodes get both IPv4 and IPv6 addresses 
2. No DNS64 needed for nodes subnet since they have IPv4 and can reach IPv4 endpoints directly
3. Worker route table needs 0.0.0.0/0 → NAT GW for IPv4 egress
4. Security group needs an additional outbound rule for all traffic to 0.0.0.0/0
```

kon-angelo: @hebelsan the steps you suggested would fit in the implementation PR and is an excellent suggestion. However, do we want to add them to this PR that contains only the proposal ?

hebelsan: If the plan is to merge this PR and with it the usage guide ‎(_docs/usage/flexible-network-configuration.md_) I'd add them here.

hebelsan (approved):
/lgtm

PR #1731, opened 2026-03-11
Replace `ChartApplier` with controller-runtime's client to deploy `MachineClass`es and their `Secret`s

plkokanov:

/area control-plane
/kind enhancement
/platform aws

This PR replaces the [`github.com/gardener/gardener/pkg/client/kubernetes.ChartApplier`](https://github.com/gardener/gardener/blob/b5d62d852baa2256c8dab11451cc1f8d2692bc6f/pkg/client/kubernetes/chartapplier.go#L18-L24) and the `machineclass` chart used to deploy `MachineClass`es and their `Secret`s with controller runtime's client.
With this we can parallelise the requests to the API server and use `Patch` instead of `Update` to modify resources.

This is mostly necessary to reduce the deployment time when dealing with a large number of `MachineClass` resources and a latency to the kube-apiserver of around 200ms, which can happen if the `kube-apiserver` is in a different region.
- With the ChartApplier, it takes around 8 minutes to deploy 300 `MachineClass`es when latency is around 200ms.
- With controller-runtime's client and parallelisation across 50 go routines, it takes around 10 seconds.

Additionally, the concrete [`AWSProviderSpec`](https://github.com/gardener/machine-controller-manager-provider-aws/blob/1a6cbb8dbf025e56245c764084a82e665a629faa/pkg/aws/apis/aws_provider_spec.go#L54) type is now used from [machine-controller-manager-provider-aws](https://github.com/gardener/machine-controller-manager-provider-aws) instead of a `map[string]interface{}`.

Part of https://github.com/gardener/gardener/issues/14137

**Release note**:

```other operator
`MachineClass`es and their `Secret`s are now deployed with the controller-runtime's client instead of using the [`github.com/gardener/gardener/pkg/client/kubernetes.ChartApplier`](https://github.com/gardener/gardener/blob/b5d62d852baa2256c8dab11451cc1f8d2692bc6f/pkg/client/kubernetes/chartapplier.go#L18-L24) and the `machineclass` chart.
Additionally, the deployments are done in parallel, which speeds up the whole operation. This is important for cases when there is high latency to the `kube-apiserver` and a large number of resources have to be deployed.
```


plkokanov:
/unhold

I decided to change the int values in the aws worker extension provider type to `int32` to keep them in line with what was consumed so far by `mcm-provider-aws`.

pkg/controller/worker/machines.go
```diff
@@ -61,12 +72,38 @@ func (w *WorkerDelegate) MachineClass() client.Object {
...
+	}
+	if err := flow.ParallelN(maxConcurrentMachineTasks, secretsTaskFns...)(ctx); err != nil {
+		return err
+	}
+
+	var machineClassesTaskFns = make([]flow.TaskFn, 0, len(w.machineClassToMutateFuncMap))
+	for machineClass, mutateFn := range w.machineClassToMutateFuncMap {
+		machineClassesTaskFns = append(machineClassesTaskFns, func(ctx context.Context) error {
+			if _, err := controllerutils.GetAndCreateOrMergePatch(ctx, w.client, machineClass, mutateFn); err != nil {
+				return fmt.Errorf("could not deploy MachineClass '%s':  %w", client.ObjectKeyFromObject(machineClass), err)
```
hebelsan: Very small minor:
```suggestion
				return fmt.Errorf("could not deploy MachineClass '%s': %w", client.ObjectKeyFromObject(machineClass), err)
```

pkg/controller/worker/machines.go
```diff
@@ -61,12 +72,38 @@ func (w *WorkerDelegate) MachineClass() client.Object {
...
 func (w *WorkerDelegate) DeployMachineClasses(ctx context.Context) error {
-	if w.machineClasses == nil {
+	if len(w.machineClassToMutateFuncMap) == 0 || len(w.machineClassSecretToMutateFuncMap) == 0 {
 		if err := w.generateMachineConfig(ctx); err != nil {
 			return err
 		}
 	}
-	return w.seedChartApplier.ApplyFromEmbeddedFS(ctx, charts.InternalChart, filepath.Join(charts.InternalChartsPath, "machineclass"), w.worker.Namespace, "machineclass", kubernetes.Values(map[string]interface{}{"machineClasses": w.machineClasses}))
+
+	var secretsTaskFns = make([]flow.TaskFn, 0, len(w.machineClassSecretToMutateFuncMap))
```
hebelsan:
```suggestion
	secretsTaskFns := make([]flow.TaskFn, 0, len(w.machineClassSecretToMutateFuncMap))
```

pkg/controller/worker/actuator.go
```diff
@@ -122,17 +117,26 @@ func NewWorkerDelegate(
...
+		serverVersion: serverVersion,
 
 		cloudProfileConfig: config,
 		cluster:            cluster,
 		worker:             worker,
 	}, nil
 }
 
-// GetMachineClasses returns the islice  of machine classes contained inside the worker delegate.
+// GetMachineClasses returns the islice of machine classes contained inside the worker delegate.
```
hebelsan:
```suggestion
// GetMachineClasses returns the slice of machine classes contained inside the worker delegate.
```

pkg/controller/worker/machines.go
```diff
@@ -217,41 +259,24 @@ func (w *WorkerDelegate) generateMachineConfig(ctx context.Context) error {
...
 			if workerConfig.CapacityReservation != nil {
 				capacityReservationOpts := *workerConfig.CapacityReservation
-				capacityReserverationCfg := map[string]string{}
-
-				if capacityReservationOpts.CapacityReservationPreference != nil {
-					capacityReserverationCfg["capacityReservationPreference"] = *capacityReservationOpts.CapacityReservationPreference
-				}
-				if capacityReservationOpts.CapacityReservationID != nil {
-					capacityReserverationCfg["capacityReservationId"] = *capacityReservationOpts.CapacityReservationID
+				capacityReserverationCfg := &awsmachineapi.AWSCapacityReservationTargetSpec{
```
hebelsan: Don't introduced by you but let's fix the typo in `capacityReserverationCfg`

pkg/controller/worker/machines.go
```diff
@@ -159,56 +197,60 @@ func (w *WorkerDelegate) generateMachineConfig(ctx context.Context) error {
...
+				Architecture: ptr.To("amd64"),
+				Capacity: corev1.ResourceList{
+					corev1.ResourceCPU:    resource.MustParse("4"),
+					"GPU":                 resource.MustParse("0"),
+					corev1.ResourceMemory: resource.MustParse("16Gi"),
+				},
+				InstanceType: "m4.xlarge",
+				Region:       "eu-west-1",
+				Zone:         "eu-west-1a",
+			}
```
hebelsan: The nodeTemplate should be empty here, I guess this is a leftover from tests?

pkg/controller/worker/machines.go
```diff
@@ -217,41 +259,24 @@ func (w *WorkerDelegate) generateMachineConfig(ctx context.Context) error {
...
 				capacityReservationOpts := *workerConfig.CapacityReservation
-				capacityReserverationCfg := map[string]string{}
-
-				if capacityReservationOpts.CapacityReservationPreference != nil {
-					capacityReserverationCfg["capacityReservationPreference"] = *capacityReservationOpts.CapacityReservationPreference
-				}
-				if capacityReservationOpts.CapacityReservationID != nil {
-					capacityReserverationCfg["capacityReservationId"] = *capacityReservationOpts.CapacityReservationID
+				capacityReserverationCfg := &awsmachineapi.AWSCapacityReservationTargetSpec{
+					CapacityReservationPreference:       ptr.Deref(capacityReservationOpts.CapacityReservationPreference, ""),
```
hebelsan, line 265: The original code only set this field when the pointer was non-nil, did you double check that this is not an issue?

hebelsan (commented):
Thanks for the contribution!
I've left some comments.

pkg/controller/worker/machines.go
plkokanov: I saw that in the chart it was defined with these fields in the `values.yaml` file and kept it here as a default:
https://github.com/gardener/gardener-extension-provider-aws/blob/7b00a07af2e5301d1e6a4310f54bf67c8b19df9b/charts/internal/machineclass/values.yaml#L6-L14

pkg/controller/worker/machines.go
plkokanov: I will test it out again, but if we do not set it, it will anyways just get defaulted to an empty string.

pkg/controller/worker/machines.go
plkokanov: Seems like this field was actually part an element in the machineclasses array in the `valyes.yaml` file so it should not be taken into account when generating the charts.

I've also removed the default values from the code, introduced in this PR.

pkg/controller/worker/machines.go
plkokanov: `CapacityReservationPreference` has the `omitempty` json tag, so it should not matter if it's set to empty string here.

hebelsan (commented):
/lgtm
/approve

PR #1732, opened 2026-03-11
Introduce MTUCustomizer feature gate to phase out MTU customization

axel7born:

/area networking
/kind enhancement
/platform aws

* Adds a MTUCustomizer feature gate to the controller configuration — this controls whether the mtu-customizer DaemonSet is deployed on the seed
* Adds an enableMTUCustomizer boolean field to InfrastructureConfig — this controls whether the custom-mtu.service systemd unit and /opt/bin/mtu-customizer.sh script are deployed to the shoot worker nodes (per-shoot opt-in, default: true)

Fixes #

**Release note**:

```other operator
[DEPRECATION] MTU customization on AWS nodes is being phased out. Previously, a DaemonSet on
seed nodes and a systemd unit on shoot nodes manually reduced network interface MTU to 1460.
With Calico and Cilium supporting MTU auto-detection, this is no longer required. A new
`MTUCustomizer` feature gate (default=true) has been introduced to allow gradual rollout of
the removal. Operators can proactively disable it by setting `featureGates.MTUCustomizer: false`
in the provider-aws controller configuration. The feature gate and all related code will be
removed in a future release.
```

```feature user
`enableMTUCustomizer` field has been added to the `InfrastructureConfig` (default=true) to
preserve the existing behavior for all shoots. Users who no longer require the MTU to be set
to 1460 on worker nodes can opt out by setting `enableMTUCustomizer: false` in their shoot's `InfrastructureConfig`.
```


hebelsan (commented):
Thanks for the contribution!

In your PR text under **What this PR does / why we need it:** you wrote:
*...Adds an enableMTUCustomizer boolean field to InfrastructureConfig — this controls whether the custom-mtu.service systemd unit and /opt/bin/mtu-customizer.sh script are deployed to the shoot worker nodes (per-shoot opt-in, **default: false**)*
It should be **default: true** no?

hebelsan (commented):
/lgtm
/approve

PR #1741, opened 2026-03-25
Better support for BYO Infrastructure

### LINKED ISSUE #1: Specify additional security groups for nodes.
It should be possible to deploy 2 or more Shoots in the same network (in AWS - VPC) and configure the secure groups on the nodes, so the pods / nodes from the different clusters are routeable to each other.

This is a requirement for Istio multicluster:

```
The usage of an RFC1918 network, VPN, or alternative more advanced network techniques to meet the following requirements:
- Individual cluster Pod CIDR ranges and service CIDR ranges must be unique across the multicluster environment and may not overlap.
- All pod CIDRs in every cluster must be routable to each other.
- All Kubernetes control plane API servers must be routable to each other.
```

### LINKED ISSUE #3: [AWS] Allow using existing subnets
End-users are interested in not only using an existing VPC but also existing subnets for their AWS shoot clusters. Today, the AWS infrastructure extension always creates new, dedicated subnets (even if an existing VPC is used).

We could allow the following `InfrastructureConfig`:

```yaml
kind: InfrastructureConfig
networks:
  vpc:
    id: vpc-123456 # re-use existing VPC
  zones:
  - name: eu-west-1a
    internalID: subnet-123456 # re-use existing subnet
    publicID: subnet-7890ab # re-use existing subnet
    workersID: subnet-cdefgh # re-use existing subnet
```

We could even allow a mixed setup:

```yaml
kind: InfrastructureConfig
networks:
  vpc:
    id: vpc-123456 # re-use existing VPC
  zones:
  - name: eu-west-1a
    internal: 10.250.112.0/22 # create a new, dedicated subnet for internal LBs
    publicID: subnet-7890ab # re-use existing subnet
    workersID: subnet-cdefgh # re-use existing subnet
```

Constraints:

* If an existing subnet for public load balancers is used then we require that a NAT gateway + elastic IP is attached to this subnet (similar to how we require that an internet gateway is attached if an existing VPC is used)
* We keep creating dedicated route tables between the subnets.

### LINKED ISSUE #377: Support for custom route to internet gateway in AWS
/area networking
/kind enhancement
/priority 3
/platform aws

Brief summary:

We are using Gardener provisioned cluster in AWS. As per our network design we use Transit gateways and attach the VPC hosting Gardener cluster to a transit gateway in another AWS account which acts as a traffic hub.
We are creating a VPC on our own and supplying this VPC id while provisioning Gardener cluster. Whereas Gardener creates subnet, routes etc.
 

Limitations we are facing:

We cannot use our transit account for outbound connectivity and instead we have to attach an internet gateway to the same VPC hosting the gardener cluster.
We are adding routes to transit gateway manually after the cluster is provisioned for intranet connectivity.
 

What we would like to achieve?

We want the outbound and intranet connectivity via the transit gateway. So we would like to hear from you what is the best solution here. We can manage subnets, routes in our terraform code if you could guide us on the requirement of gardener cluster and then while provisioning the gardener cluster we can provide the subnet, route ids as inputs along with VPC id.
Another possibility we can think of is to have a way to supply additional routes information during provisioning. This way gardener can manage the subnet, route etc but we can provide the additional route which we would want to add for intranet connectivity or outbound connectivity.

### LINKED ISSUE #1008: Possibility of additing custom routes to routing table (Traffic routing control for egress)
/area networking
/kind enhancement
/platform aws

When a shoot cluster is created, user shall have possibility to add custom routes (e.g default route) so that one can divert egress traffic for cluster to another VPC/VPG/Transit Gateway in another VPC or firewall instances in same VPC. When custom route for default route is provided then Gardener shall not create any NATGW as it shall rely on existing routing to provide internet connectivity to Gardener seed. 

**Why is this needed**:
By default Gardener creates NAT GW and default routes point to NATGW which allow unrestricted internet access to shoot clusters. This is security risk for many types of deployments where user wants to egress traffic via predefined firewall instances or other VPCs.

kon-angelo:

/area control-plane
/kind enhancement
/platform aws

Is an first implementation of the proposal in https://github.com/gardener/gardener-extension-provider-aws/pull/1715

Fixes https://github.com/gardener/gardener-extension-provider-aws/issues/1008
Fixes https://github.com/gardener/gardener-extension-provider-aws/issues/3
Fixes https://github.com/gardener/gardener-extension-provider-aws/issues/1
Fixes https://github.com/gardener/gardener-extension-provider-aws/issues/377

**Release note**:

```other operator
Adds capability to deploy shoot in existing subnet and provides better support for BYO infrastructure scenarios such user route-tables, security groups, ingress and egress traffic control.
```


pkg/controller/infrastructure/infraflow/reconcile.go
```diff
@@ -1354,7 +1687,7 @@ func (c *FlowContext) deleteNATGateway(zoneName string) flow.TaskFn {
 				return err
 			}
 		}
-		child.Delete(IdentifierZoneNATGateway)
```
hebelsan, line 1369: Why did you remove the state deletion?

pkg/apis/aws/validation/infrastructure.go
```diff
@@ -192,6 +358,13 @@ func ValidateInfrastructureConfigUpdate(oldConfig, newConfig *apisaws.Infrastruc
 	allErrs = append(allErrs, apivalidation.ValidateImmutableField(newVPC.ID, oldVPC.ID, vpcPath.Child("id"))...)
 	allErrs = append(allErrs, apivalidation.ValidateImmutableField(newVPC.CIDR, oldVPC.CIDR, vpcPath.Child("cidr"))...)
 
+	// NodesSecurityGroupID is immutable once set
+	allErrs = append(allErrs, apivalidation.ValidateImmutableField(
+		newConfig.Networks.NodesSecurityGroupID,
+		oldConfig.Networks.NodesSecurityGroupID,
+		field.NewPath("networks.nodesSecurityGroupID"),
```
hebelsan: this should be `field.NewPath("networks").Child("nodesSecurityGroupID")`

pkg/controller/infrastructure/infraflow/reconcile.go
```diff
@@ -710,12 +747,17 @@ func (c *FlowContext) ensureNodesSecurityGroup(ctx context.Context) error {
 	return nil
 }
 
+// hasCIDRs returns true if the security group rule has at least one IPv4 or IPv6 CIDR block.
+func hasCIDRs(rule *awsclient.SecurityGroupRule) bool {
```
hebelsan: the helper function `hasCIDRs()` might be better placed in utils.go

pkg/controller/infrastructure/infraflow/context.go
```diff
@@ -278,9 +287,25 @@ func (c *FlowContext) tagKeyCluster() string {
...
+// isOwnedClusterResource checks if a resource's cluster tag value indicates ownership.
+// Accepts both the legacy value ("1") and the standard value ("owned").
+func isOwnedClusterResource(tagValue string) bool {
+	return tagValue == TagValueCluster || tagValue == TagValueClusterOwned
+}
+
+// clusterTagKeyExistsFilter returns an EC2 filter that matches resources having the cluster
+// tag key regardless of value. Used for discovery where we want to find all resources
+// associated with this cluster (owned, shared, or legacy).
+func (c *FlowContext) clusterTagKeyExistsFilter() ec2types.Filter {
```
hebelsan: In my view it would make more sense to add a `WithTagKeyExists(key string)` method on the filterBuilder instead.

docs/usage/flexible-network-configuration.md
hebelsan, line 1: I'd remove this file here since it's already part of your proposal PR.

charts/internal/seed-controlplane/charts/cloud-controller-manager/templates/deployment.yaml
```diff
@@ -46,6 +46,9 @@ spec:
         - --cluster-name={{ .Values.clusterName }}
         - --concurrent-service-syncs=10
         - --configure-cloud-routes=false
+        {{- if .Values.disableServiceController }}
```
hebelsan, line 49: I would also add a default value (`false`) in values.yaml

pkg/controller/infrastructure/infraflow/reconcile.go
```diff
@@ -1103,6 +1428,10 @@ func (c *FlowContext) ensureSubnetCidrReservation(ctx context.Context) error {
 		}
 
 		if key == IdentifierZoneSubnetWorkers {
+			// BYO subnets may not have IPv6 CIDR blocks.
+			if len(subnet.Ipv6CidrBlocks) == 0 {
+				continue
```
hebelsan, line 1728: Shouldn't we return an error here?
The function `ensureSubnetCidrReservation` already gates on `ContainsIPv6(c.getIpFamilies())` at the top, so it only runs when IPv6 is actually enabled. 
If the user has configured a BYO IPv6 cluster, their worker subnets must have IPv6 CIDR blocks no?
I think in case of BYO we should only validate and set the state of the ipv6 subnet cidr here.

pkg/controller/infrastructure/infraflow/reconcile.go
```diff
@@ -1129,6 +1458,10 @@ func (c *FlowContext) ensureSubnetCidrReservation(ctx context.Context) error {
 		}
 
 		if key == IdentifierZoneSubnetWorkers {
+			// BYO subnets may not have IPv6 CIDR blocks.
+			if len(subnet.Ipv6CidrBlocks) == 0 {
```
hebelsan, line 1763: Again, this check seems not correct.
If we return an error above for the same check it becomes redundant here.

pkg/controller/controlplane/valuesprovider.go
```diff
@@ -623,6 +649,15 @@ func getCCMChartValues(
...
 	}
 
+	// Disable the CCM service controller when no public or internal subnets are available.
+	// Without LB subnets, the service controller cannot create load balancers and would
+	// error on every type: LoadBalancer Service. The CCM is still needed for node lifecycle.
+	_, hasPublic := helper.FindSubnetForPurpose(infraStatus.VPC.Subnets, apisaws.PurposePublic)
+	_, hasInternal := helper.FindSubnetForPurpose(infraStatus.VPC.Subnets, apisaws.PurposeInternal)
+	if hasPublic != nil && hasInternal != nil {
+		values["disableServiceController"] = true
+	}
```
hebelsan, line 672: The documentation gives the idea that if I pre tag my subnets with the correct tags, then I would not need to add them to the shoot config. but that will not work because the service controller is disabled.

The infra reconciler only stores subnets in state (and thus in infraStatus.VPC.Subnets) when they appear in the zone config — IdentifierZoneSubnetPublic is only set when zone.PublicSubnetID != nil, and similarly for internal. If neither is set in the shoot config, those entries are never written to state, so infraStatus.VPC.Subnets contains only the workers subnet with PurposeNodes.

pkg/apis/aws/validation/infrastructure.go
hebelsan, line 1: Currently in BYO it's allowed to create one zone **with** public/internal subnets and the others without.
Does this configuration make sense?
To me this would look like a misconfiguration and we should prevent it.

example/30-infrastructure.yaml
```diff
@@ -57,6 +57,28 @@ spec:
         public: 10.250.96.0/22
         workers: 10.250.0.0/19
       # elasticIPAllocationID: eipalloc-123456 # Allocation ID of the Elastic IP that will be attached to the NAT gateway in this zone
+
+    ## BYO Infrastructure example:
+    ## Use an existing VPC with pre-provisioned worker subnets and security group.
+    ## Gardener does not create any network resources. The user-provided security group
+    ## replaces Gardener's nodes SG entirely (see docs/usage/flexible-network-configuration.md).
+    ## LB subnets can be specified explicitly via publicSubnetID/internalSubnetID (Gardener
+    ## auto-tags them), or discovered by CCM/LBC via standard AWS tags if pre-tagged by user.
```
hebelsan, line 66: Should we remove the tags on public and internal subnets if we added them?
We could create an additional tag that hints that they weren't present and that we added them to distinguish if they were already present.

pkg/controller/infrastructure/configvalidator.go
```diff
@@ -189,13 +225,111 @@ func (c *configValidator) validateVPC(
...
+		workerErrs := c.validateBYOSubnet(ctx, awsClient, zone.WorkersSubnetID, fldPath, vpcID, zone.Name)
+		allErrs = append(allErrs, workerErrs...)
+		if len(workerErrs) > 0 {
+			continue
+		}
+
+		// When DualStack/IPv6 is enabled, BYO worker subnets must have an IPv6 CIDR block
+		// from the VPC's IPv6 pool. Gardener uses this for CIDR reservations (services).
+		dualStack := config.DualStack != nil && config.DualStack.Enabled
+		if dualStack {
```
hebelsan: This is not enough.
One can also create a real dual-stack or IPv6 cluster and we should check that they have a IPv6 CIDRs as well.

pkg/controller/infrastructure/configvalidator.go
hebelsan, line 1: Can we add some more prerequisite checks in the configvalidator?
I was thinking about:
- the VPC must have DNS hostnames enabled
- LB subnets must have min 8 available IPs per LB subnet
- in case of IPv4 clusters we should add a check that verifies that the network nodes range lays inside the CIDR of the provided worker subnet

Also, what do you think about verifying whether a NAT Gateway, Transit Gateway, or VPC endpoint is present? Without at least one of these, the connection won’t be possible.

hebelsan (commented):
During my tests I've encountered an issue when creating a dualstack cluster with BYO.
I've created a fix PR here: https://github.com/kon-angelo/gardener-extension-provider-aws/pull/1

pkg/controller/infrastructure/configvalidator.go
kon-angelo: > LB subnets must have min 8 available IPs per LB subnet

where does the 8 ips (as a magic number come from)? I am also not sure that this needs to be a hard limit. If the subnets are shared, then maybe IPs can be reserved by other clusters before us creating an LB. I am not sure if we improve much with this check, so I would be interested to hear more about your thoughts.

kon-angelo: > Also, what do you think about verifying whether a NAT Gateway, Transit Gateway, or VPC endpoint is present? Without at least one of these, the connection won’t be possible.

Because it's not necessary that one of these exist. You can NAT through an EC2 instance with a public IP should you wish or some other solution. This would make it really only applicable to specific and known AWS solution. There is also peering etc. I will not completely exclude it, but i would defer this in the future as additional validation if that's okay with you.

pkg/controller/controlplane/valuesprovider.go
kon-angelo: I thought taht his was addressed but let me double check 👍

kon-angelo: I thought taht his was addressed but let me double check 👍

pkg/controller/infrastructure/configvalidator.go
hebelsan: I just remembered running into this during my testing.
The 8-IP requirement appears to come from the aws-load-balancer-controller:
https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/af746dc597f92b6e7602ea42586d5a17d4c2c9ed/pkg/networking/subnet_resolver.go#L43
AWS requires every subnet used by an Application Load Balancer to have at least 8 available IP addresses (https://docs.aws.amazon.com/elasticloadbalancing/latest/application/application-load-balancers.html#availability-zones:~:text=at%20least%20eight%20free%20IP%20addresses%20per%20subnet).
That said, it looks like defaultMinimalAvailableIPAddressCount is only enforced during auto-discovery (tag-based subnet discovery). If subnet IDs are provided explicitly, this check isn't performed.
Documenting the recommendation to have at least 8 available IPs per load balancer subnet in the usage guide should be sufficient.

PR #1744, opened 2026-03-27
Document route table entries limit

### LINKED ISSUE #691: Add documentation about route table quota limits with regards to overlay free networking

/area documentation
/area networking
/kind enhancement
/platform aws

Add documentation about route table quota limits with regards to overlay free networking.

**Why is this needed**:
In case the overlay network is disabled, the routes required by the pod network are configured in the virtual network (VPC). The [aws-custom-route-controller](https://github.com/gardener/aws-custom-route-controller) performs this if required.

Unfortunately, the [default quotas for route table entries per route table](https://docs.aws.amazon.com/vpc/latest/userguide/amazon-vpc-limits.html#vpc-limits-route-tables) are fairly limited with only `50` entries. This means that a cluster more than `50` nodes will run into the default limit resulting in a partially unusable pod network.

Apart from the default limit (`50`) it is also important to mention the hard limit of `1000` route table entries resulting in a hard stop regarding the amount of cluster nodes that can be used in an overlay free pod network.

hebelsan:

/area quality
/kind enhancement
/platform aws

Document route table entries limit if network overlay is disabled.

Fixes https://github.com/gardener/gardener-extension-provider-aws/issues/691

**Release note**:

```other operator
Document route table entries limit if network overlay is disabled
```


kon-angelo (approved):
/lgtm

PR #1745, opened 2026-03-30
Add pvca webhook

Kostov6:

/area storage
/kind enhancement
/platform aws

This PR introduces the `seed-pvca` webhook that ensurers that there is at least 6 Hours between pvc upscales on aws seeds. It is needed because a volume can be modified up to four times within a rolling 24-hour period ([ref](https://docs.aws.amazon.com/ebs/latest/userguide/ebs-modify-volume.html)).

Part of https://github.com/gardener/gardener/issues/14383

You can verify using ` helm template test charts/gardener-extension-provider-aws > foo && echo OK ` and playing around different values to validate correctness

**Release note**:

```feature developer
A new webhook `seed-pvca` is introduced that ensures that there is at least 6 Hours between pvc upscales on aws seeds
```


plkokanov (changes requested):
The `provider-aws` extension can be installed on all types of seeds (gcp, azure, etc) depending on the required `DNSRecord` type and the cloud provider of the hosted shoots. We have to make sure that the aws-specific PVCA mutation does not happen in such cases.
You can check how this is done [here](https://github.com/gardener/gardener-extension-provider-azure/blob/master/pkg/webhook/highavailability/namespace/mutator.go#L70-L74), although I'm not sure if there's a better way to do it. For instance, the [seedprovider](https://github.com/gardener/gardener-extension-provider-gcp/tree/master/pkg/webhook/seedprovider) webhook has a namespace selector that only selects shoot control plane namespaces labeled with `seed.gardener.cloud/provider=aws`. One thing to note is that the PVCA webhook also has to select resources from the garden namespace.

pkg/webhook/pvca/mutator.go
```diff
@@ -0,0 +1,47 @@
...
+	if !ok {
+		return fmt.Errorf("could not mutate: object is not of type PersistentVolumeClaimAutoscaler")
+	}
+
+	extensionswebhook.LogMutation(m.logger, "PersistentVolumeClaimAutoscaler", pvca.GetNamespace(), pvca.GetName())
+
+	for i := range pvca.Spec.VolumePolicies {
+		if pvca.Spec.VolumePolicies[i].ScaleUp != nil &&
+			(pvca.Spec.VolumePolicies[i].ScaleUp.CooldownDuration == nil || pvca.Spec.VolumePolicies[i].ScaleUp.CooldownDuration.Duration < minCooldownDuration) {
+			pvca.Spec.VolumePolicies[i].ScaleUp.CooldownDuration = &metav1.Duration{Duration: minCooldownDuration}
```
plkokanov, line 43: The `scaleUp.minStepAbsolute` should also be some larger value so that there is enough room for data aggregation within the 6 hour window.
I'm not entirely sure what the value should be though.

pkg/webhook/pvca/add.go
```diff
@@ -0,0 +1,60 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+
+	"github.com/gardener/gardener-extension-provider-aws/pkg/aws"
+)
+
+const (
+	// WebhookName is the name of the webhook.
+	WebhookName = "shoot-pvca"
```
plkokanov:
```suggestion
	WebhookName = "seed-pvca"
```

charts/gardener-extension-provider-aws/templates/_helpers.tpl
```diff
@@ -61,3 +61,11 @@ true
 true
 {{- end }}
 {{- end -}}
+
+{{- define "disable.webhooks" -}}
+{{- $disableWebhooks := .Values.disableWebhooks -}}
+{{- if ne (include "seed.provider" . | trim) "aws" -}}
+{{- $disableWebhooks = append $disableWebhooks "shoot-pvca" -}}
```
plkokanov:
```suggestion
{{- $disableWebhooks = append $disableWebhooks "seed-pvca" -}}
```

pkg/webhook/pvca/add.go
```diff
@@ -0,0 +1,60 @@
...
+	// DefaultAddOptions are the default AddOptions for AddToManager.
+	DefaultAddOptions = AddOptions{}
+)
+
+// AddOptions are options to apply when adding the AWS shoot webhook to the manager.
+type AddOptions struct{}
+
+// AddToManagerWithOptions creates a webhook with the given options and adds it to the manager.
+func AddToManagerWithOptions(mgr manager.Manager, _ AddOptions) (*extensionswebhook.Webhook, error) {
+	logger := log.Log.WithName("aws-shoot-pvca-webhook")
```
plkokanov:
```suggestion
	logger := log.Log.WithName("aws-seed-pvca-webhook")
```

plkokanov (dismissed):

pkg/webhook/pvca/mutator.go
Kostov6: Let's introduce this via another PR

RadaBDimitrova (commented):
/lgtm
After validating with @Kostov6 on real infrastructure alongside the new release of pvc-autoscaler, it works as expected.

pkg/webhook/pvca/mutator.go
```diff
@@ -0,0 +1,48 @@
...
+}
+
+// NewMutator creates a new Mutator that mutates resources in the shoot cluster.
+func NewMutator(logger logr.Logger) extensionswebhook.Mutator {
+	return &mutator{logger: logger}
+}
+
+// Mutate mutates PersistentVolumeClaimAutoscaler resources by defaulting cooldownDuration.
+func (m *mutator) Mutate(_ context.Context, newObj, _ client.Object) error {
+	m.logger.Info("mutating PersistentVolumeClaimAutoscaler", "namespace", newObj.GetNamespace(), "name", newObj.GetName())
```
wpross: that's redundant I guess as a few lines down `extensionswebhook.LogMutation()` is called.

pkg/webhook/pvca/mutator.go
```diff
@@ -0,0 +1,48 @@
...
+	pvca, ok := newObj.(*pvcautoscalingv1alpha1.PersistentVolumeClaimAutoscaler)
+	if !ok {
+		return fmt.Errorf("could not mutate: object is not of type PersistentVolumeClaimAutoscaler")
+	}
+
+	extensionswebhook.LogMutation(m.logger, "PersistentVolumeClaimAutoscaler", pvca.GetNamespace(), pvca.GetName())
+
+	for i := range pvca.Spec.VolumePolicies {
+		if pvca.Spec.VolumePolicies[i].ScaleUp != nil &&
```
wpross, line 41: what happens if `ScaleUp` is nil? Or is that maybe not possible?

pkg/webhook/pvca/mutator.go
```diff
@@ -0,0 +1,48 @@
...
+	"time"
+
+	extensionswebhook "github.com/gardener/gardener/extensions/pkg/webhook"
+	pvcautoscalingv1alpha1 "github.com/gardener/pvc-autoscaler/api/autoscaling/v1alpha1"
+	"github.com/go-logr/logr"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const minCooldownDuration = 6 * time.Hour
```
wpross, line 20: I would add a comment above adding the link to https://docs.aws.amazon.com/ebs/latest/userguide/ebs-modify-volume.html and quoting:
> You can modify a volume up to four times within a rolling 24-hour period

so that maintainers understand where that magic number is coming from.

pkg/webhook/pvca/mutator_test.go
```diff
@@ -0,0 +1,106 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+var _ = Describe("Mutator", func() {
+	var (
+		mutator = NewMutator(log.Log.WithName("test"))
+	)
+
+	It("should default cooldownDuration to 6h when missing", func() {
```
wpross, line 24: maybe add another TC with multiple VolumePolicies of which only a part of them required mutation.

pkg/webhook/pvca/add.go
```diff
@@ -0,0 +1,57 @@
...
+		Types:             types,
+		Webhook:           &admission.Webhook{Handler: handler, RecoverPanic: ptr.To(true)},
+		Path:              WebhookName,
+		NamespaceSelector: nil,
+	}, nil
+}
+
+// AddToManager creates a webhook with the default options and adds it to the manager.
+func AddToManager(mgr manager.Manager) (*extensionswebhook.Webhook, error) {
+	return AddToManagerWithOptions(mgr, DefaultAddOptions)
```
wpross: Should the `AddToManagerWithOptions` exist for possible future options? if not, the logic could be added directly into the `AddToManager` fctn since no options are used.

charts/gardener-extension-provider-aws/templates/_helpers.tpl
```diff
@@ -61,3 +61,11 @@ true
 true
 {{- end }}
 {{- end -}}
+
+{{- define "disable.webhooks" -}}
+{{- $disableWebhooks := .Values.disableWebhooks -}}
+{{- if ne (include "seed.provider" . | trim) "aws" -}}
```
wpross: Is the `trim` really required? As e.g. https://github.com/gardener/gardener-extension-provider-aws/blob/master/charts/gardener-extension-provider-aws/templates/storageclass.yaml#L1 also works without it. 
Anyways, I guess the `seed.provider` helper is a bit broken and that's probably why you added the `trim` . Should probably be something like:
```
  {{- define "seed.provider" -}}
  {{- if .Values.gardener.seed -}}
  {{- .Values.gardener.seed.provider -}}
  {{- end -}}
  {{- end -}}
```

wpross (changes requested):
* small typo in the Release note (and description): `ensurers` -> `ensures`
* I would also add a link to https://github.com/gardener/gardener/issues/14383 in the description so that it's easier later to track down to the bigger story
* a link to https://docs.aws.amazon.com/ebs/latest/userguide/ebs-modify-volume.html in the PR description also helps to understand why the 6 hours need to be enforced as a min

pkg/webhook/pvca/mutator.go
Kostov6: It defaults to an empty object so it can not be nil. We have made it a pointer for defaulting purposes

charts/gardener-extension-provider-aws/templates/_helpers.tpl
Kostov6: So your proposal is to replace
```
{{- define "seed.provider" -}}
  {{- if .Values.gardener.seed }}
{{- .Values.gardener.seed.provider }}
  {{- else -}}
""
  {{- end }}
{{- end -}}
```
with
```
  {{- define "seed.provider" -}}
  {{- if .Values.gardener.seed -}}
  {{- .Values.gardener.seed.provider -}}
  {{- end -}}
  {{- end -}}
```
and remove the trim?

wpross: exactly

plkokanov (commented):
/lgtm

charts/gardener-extension-provider-aws/templates/_helpers.tpl
```diff
@@ -49,15 +49,27 @@ true
...
 {{- end -}}
 
 {{- define "runtimeCluster.enabled" -}}
 {{- if and .Values.gardener.runtimeCluster .Values.gardener.runtimeCluster.enabled }}
 true
 {{- end }}
 {{- end -}}
+
+{{- define "persistentVolumeClaimAutoscaler.enabled" -}}
+{{- if and .Values.gardener.seed .Values.gardener.seed.spec.settings.persistentVolumeClaimAutoscaler .Values.gardener.seed.spec.settings.persistentVolumeClaimAutoscaler.enabled }}
```
wpross: same as in case of the `seed.provider`:
to be consistent, add a `-` at the end and then get rid of the `| trim` in
`{{- if or (ne (include "seed.provider" . ) "aws") (ne (include "persistentVolumeClaimAutoscaler.enabled" . | trim ) "true") -}}`

pkg/webhook/pvca/add.go
```diff
@@ -0,0 +1,52 @@
...
+)
+
+const (
+	// WebhookName is the name of the webhook.
+	WebhookName = "seed-pvca"
+)
+
+var (
+	// DefaultAddOptions are the default AddOptions for AddToManager.
+	DefaultAddOptions = AddOptions{}
```
wpross: that seems to be dead code

wpross (changes requested):

plkokanov:
/ping @Kostov6 

As there are some unresolved comments.

pkg/webhook/pvca/mutator.go
```diff
@@ -0,0 +1,48 @@
...
+)
+
+// A volume can be modified up to four times within a rolling 24-hour period (https://docs.aws.amazon.com/ebs/latest/userguide/ebs-modify-volume.html)
+const minCooldownDuration = 6 * time.Hour
+
+type mutator struct {
+	logger logr.Logger
+}
+
+// NewMutator creates a new Mutator that mutates resources in the shoot cluster.
```
wpross: resource are not mutated in the shoot cluster, only in the seed cluster. Just use e.g.
```
// NewMutator creates a new Mutator
```

pkg/webhook/pvca/shoot_pvca_suite_test.go
```diff
@@ -0,0 +1,17 @@
...
+import (
+	"testing"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+func TestPVCA(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Shoot PVCA Webhook Suite")
```
wpross: same here, it's not the shoot, but the seed.

wpross (commented):
sry, the last review was so long ago that I had to work into the PR again and thus found 2 (very little) things 🙈. So best address them faster than last time  😅

wpross:
@Kostov6 Please rebase to current master so that the Build step works.

wpross (approved):
/lgtm

PR #1747, opened 2026-03-31
Enhance CIDR validation for vpcs with additional CIDR associations

AndreasBurger:

/area quality
/kind enhancement
/platform aws

**What this PR does / why we need it**: TBD

Fixes #

**Release note**:

```other operator
VPC validation now also takes additional CIDR associations into account
```

### DIFF
--- a/pkg/aws/client/client.go
+++ b/pkg/aws/client/client.go
@@ -1009,16 +1009,13 @@ func (c *Client) fromVpc(ctx context.Context, item ec2types.Vpc, withAttributes 
 		CidrBlock: aws.ToString(item.CidrBlock),
 		// CidrBlockAssociationSet: item.CidrBlockAssociationSet,
 		CidrBlockAssociationSet: func() []string {
-			if len(item.CidrBlockAssociationSet) > 0 {
-				cidrBlocks := make([]string, 0, len(item.CidrBlockAssociationSet))
-				for _, i := range item.CidrBlockAssociationSet {
-					if i.CidrBlockState.State == ec2types.VpcCidrBlockStateCodeAssociated {
-						cidrBlocks = append(cidrBlocks, aws.ToString(i.CidrBlock))
-					}
+			cidrBlocks := make([]string, 0, len(item.CidrBlockAssociationSet))
+			for _, i := range item.CidrBlockAssociationSet {
+				if i.CidrBlockState.State == ec2types.VpcCidrBlockStateCodeAssociated {
+					cidrBlocks = append(cidrBlocks, aws.ToString(i.CidrBlock))
 				}
-				return cidrBlocks
 			}
-			return []string{}
+			return cidrBlocks
 		}(),
 		IPv6CidrBlock: func() string {
 			if len(item.Ipv6CidrBlockAssociationSet) > 0 && item.Ipv6CidrBlockAssociationSet[0].Ipv6CidrBlockState.State == ec2types.VpcCidrBlockStateCodeAssociated {
--- a/pkg/aws/client/client.go
+++ b/pkg/aws/client/client.go
@@ -1004,12 +1004,24 @@ func (c *Client) describeVpcs(ctx context.Context, input *ec2.DescribeVpcsInput)
 
 func (c *Client) fromVpc(ctx context.Context, item ec2types.Vpc, withAttributes bool) (*VPC, error) {
 	vpc := &VPC{
-		VpcId:                   aws.ToString(item.VpcId),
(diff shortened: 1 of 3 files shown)


pkg/aws/client/client.go
```diff
@@ -1004,9 +1004,10 @@ func (c *Client) describeVpcs(ctx context.Context, input *ec2.DescribeVpcsInput)
 
 func (c *Client) fromVpc(ctx context.Context, item ec2types.Vpc, withAttributes bool) (*VPC, error) {
 	vpc := &VPC{
-		VpcId:     aws.ToString(item.VpcId),
-		Tags:      FromTags(item.Tags),
-		CidrBlock: aws.ToString(item.CidrBlock),
+		VpcId:                   aws.ToString(item.VpcId),
+		Tags:                    FromTags(item.Tags),
+		CidrBlock:               aws.ToString(item.CidrBlock),
+		CidrBlockAssociationSet: item.CidrBlockAssociationSet,
```
hebelsan: As of now we do not check the state of the CidrBlockAssociationSet which could be `disassociated` or `failed`.
I guess we should only add a CidrBlockAssociationSet if the state is `associated` or `associating`.

pkg/aws/client/types.go
```diff
@@ -232,6 +232,7 @@ type VPC struct {
 	Tags
 	VpcId                        string
 	CidrBlock                    string
+	CidrBlockAssociationSet      []ec2types.VpcCidrBlockAssociation
```
hebelsan: Since we only use the CIDRs, we could replace the AWS-specific type with a slice of strings, which would make future SDK upgrades easier.

hebelsan (commented):
Thanks for the PR!

pkg/aws/client/client.go
```diff
@@ -1007,8 +1007,21 @@ func (c *Client) fromVpc(ctx context.Context, item ec2types.Vpc, withAttributes
 		VpcId:     aws.ToString(item.VpcId),
 		Tags:      FromTags(item.Tags),
 		CidrBlock: aws.ToString(item.CidrBlock),
+		// CidrBlockAssociationSet: item.CidrBlockAssociationSet,
+		CidrBlockAssociationSet: func() []string {
+			if item.CidrBlockAssociationSet != nil {
+				cidrBlocks := make([]string, len(item.CidrBlockAssociationSet))
+				for _, i := range item.CidrBlockAssociationSet {
+					if i.CidrBlockState.State == ec2types.VpcCidrBlockStateCodeAssociated {
+						cidrBlocks = append(cidrBlocks, aws.ToString(i.CidrBlock))
```
hebelsan: make([]string, len(...)) creates a slice of length N already filled with empty strings. Then append adds more elements after them. The result will contain leading empty strings for every CIDR in the association set, followed by only the ones that passed the filter.

pkg/aws/client/client.go
```diff
@@ -1007,8 +1007,21 @@ func (c *Client) fromVpc(ctx context.Context, item ec2types.Vpc, withAttributes
...
+					}
+				}
+				return cidrBlocks
+			}
+			return []string{}
+		}(),
 		IPv6CidrBlock: func() string {
-			if item.Ipv6CidrBlockAssociationSet != nil {
+			if item.Ipv6CidrBlockAssociationSet != nil && item.Ipv6CidrBlockAssociationSet[0].Ipv6CidrBlockState.State == ec2types.VpcCidrBlockStateCodeAssociated {
```
hebelsan: `Ipv6CidrBlockAssociationSet` could be non-nil but empty (len == 0), in which case [0] panics.
A better check would be ` len(item.Ipv6CidrBlockAssociationSet) > 0`.
(This may be pre-existing but we should still fix it).

pkg/aws/client/client.go
```diff
@@ -1007,8 +1007,21 @@ func (c *Client) fromVpc(ctx context.Context, item ec2types.Vpc, withAttributes
 		VpcId:     aws.ToString(item.VpcId),
 		Tags:      FromTags(item.Tags),
 		CidrBlock: aws.ToString(item.CidrBlock),
+		// CidrBlockAssociationSet: item.CidrBlockAssociationSet,
+		CidrBlockAssociationSet: func() []string {
+			if len(item.CidrBlockAssociationSet) > 0 {
```
hebelsan: I don't think the length check is even needed here.
If the slice is nil or empty the loop will not iterate and we return an empty slice.

hebelsan (commented):
Thanks for the changes!
/lgtm
/approve

PR #1764, opened 2026-04-10
Modernize etcd storageclass to out-of-tree provisioner

### LINKED ISSUE #1716: Modernize etcd storageclass to use out-of-tree csi provisioner

/area storage
/kind enhancement
/platform aws

Modernize etcd storageclass `gardener.cloud-fast` to use out-of-tree csi provisioner, similar to what we already did for azure in https://github.com/gardener/gardener-extension-provider-azure/pull/1223

This gets rid of some technical debt and ensures we don't pile up more volumes which we will have to migrate eventually.

**Why is this needed**:

voelzmo:

/area storage
/kind enhancement
/platform aws

Modernizes the etcd `StorageClass` for AWS:

1. Replaces the deprecated in-tree provisioner `kubernetes.io/aws-ebs` with the EBS CSI driver `ebs.csi.aws.com`
2. Extracts the hardcoded `type: gp3` disk parameter from the `StorageClass` template into `values.yaml` under `config.etcd.storage.parameters`, making it configurable by operators.

The existing `config.etcd.storage.encrypted` boolean is preserved for backwards compatibility.

Fixes #1716

The `StorageClass` template now renders arbitrary key/value pairs from `config.etcd.storage.parameters`, so operators can pass any EBS
  CSI driver parameter (e.g. `kmsKeyId`, `throughput`) without chart changes.

The `encrypted` boolean in values is kept for backwards compatibility: if set, `encrypted: "true"` is still emitted unless the operator has already included `encrypted` in the `parameters` map, avoiding duplicate keys.

**Release note**:
```noteworthy operator
The etcd StorageClass provisioner has been switched from in-tree `kubernetes.io/aws-ebs` to the EBS CSI driver `ebs.csi.aws.com`. The disk type is now configurable via `config.etcd.storage.parameters` in the Helm values (default: `type: gp3`).
The old way of configuring it via `.Values.config.etcd.storage.encrypted` is now deprecated and will be removed with a future release.
```


voelzmo:
Especially not sure about the need for backwards compatibility regarding the `encrypted` configuration of the volumes. WDYT?

hebelsan:
> Especially not sure about the need for backwards compatibility regarding the encrypted configuration of the volumes. WDYT?

Hm, I’d prefer having the encryption under the parameter field only... 
Could we introduce this with a breaking change announcement?

AndreasBurger (approved):
lgtm

nvm, did not see the latest comments.

AndreasBurger:
And it was merged by prow. Splendid.

I'll follow up on the latest comment from Alex and do the work

AndreasBurger:
We talked about it, and this is good for now. We'll move towards the new scheme (i.e. using `.Values.config.etcd.storage.parameters` for all config) at a future point. I've put a deprecation notice for the old way into the release notes.

PR #1773, opened 2026-04-17
enable and scrape CSI driver metrics

### LINKED ISSUE #1568: Missing metric in Gardener setup
Hello everyone,
I’m trying to find the metric aws_ebs_csi_ec2_detach_pending_seconds in our Gardener setup. According to the [AWS EBS CSI driver documentation](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/docs/metrics.md), this metric was introduced in version v1.43.
In our cluster(aws canary), I see the CSI driver version is v1.47.0 already present:

```
kc describe daemonset csi-driver-node -n kube-system | grep Image
->   Image:      europe-docker.pkg.dev/sap-se-gcp-k8s-l-delivery/releases-live-public/registry_k8s_io/provider-aws/aws-ebs-csi-driver@sha256:dfdae45f7ff9c89c14bc7a0da8b81a0978f680fddeab331f8440a643909d72c6

docker run --rm europe-docker.pkg.dev/sap-se-gcp-k8s-l-delivery/releases-live-public/registry_k8s_io/provider-aws/aws-ebs-csi-driver@sha256:dfdae45f7ff9c89c14bc7a0da8b81a0978f680fddeab331f8440a643909d72c6 --version
-> {
  "driverVersion": "v1.47.0",
  "gitCommit": "",
  "buildDate": "2025-08-06T22:32:28+00:00",
  "goVersion": "go1.24.5",
  "compiler": "gc",
  "platform": "linux/arm64"
}
```
However, I don’t see this metric in Plutono.
Could anyone help me understand why it’s missing or if there’s something I might be overlooking?

/kind enhancement
/platform aws

Metrics from [AWS EBS CSI driver documentation](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/docs/metrics.md#aws-api-metrics-ebs-csi-controller)

**Why is this needed**:
We occasionally experience issues with volume mounts, especially when the hardware is degraded. The AWS CSI driver team has added these metrics to the driver which can be used in HANA Cloud. These metrics are included in the official AWS version and can be scraped in the usual way.

We want to use these metrics in collaboration with AWS to identify the root cause of these volume mount issues.

matthias-horne:

/area monitoring
/kind enhancement
/platform aws

This PR enables metrics for the CSI driver and adds a `ServiceMonitor` to scrape them.

Fixes #1568 

**Release note**:

```feature user
Metrics for the CSI driver controller are now enabled
```

### DIFF
--- a/charts/internal/seed-controlplane/charts/csi-driver-controller/templates/csi-driver-controller-deployment.yaml
+++ b/charts/internal/seed-controlplane/charts/csi-driver-controller/templates/csi-driver-controller-deployment.yaml
@@ -47,6 +47,7 @@ spec:
         - --logtostderr
         - --v=5
         - --batching=true
+        - --http-endpoint=0.0.0.0:3301
         env:
         - name: CSI_ENDPOINT
           value: unix://{{ .Values.socketPath }}/csi.sock
@@ -69,6 +70,9 @@ spec:
         ports:
         - name: healthz
           containerPort: 9808
+          protocol: TCP
+        - name: metrics
+          containerPort: 3301
           protocol: TCP
         livenessProbe:
           httpGet:
--- a/charts/internal/seed-controlplane/charts/csi-driver-controller/templates/csi-driver-controller-service.yaml
+++ b/charts/internal/seed-controlplane/charts/csi-driver-controller/templates/csi-driver-controller-service.yaml
@@ -1,0 +1,21 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: csi-driver-controller
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app: csi
+    role: controller
+  annotations:
+    networking.resources.gardener.cloud/from-all-scrape-targets-allowed-ports: '[{"port":3301,"protocol":"TCP"}]'
+    networking.resources.gardener.cloud/from-world-to-ports: '[{"port":"3301","protocol":"TCP"}]'
+spec:
+  type: ClusterIP
+  clusterIP: None
+  ports:
+  - name: metrics
+    port: 3301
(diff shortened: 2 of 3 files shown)


hebelsan:
In the issue, you mentioned that we should consider putting this feature behind a flag so it can be enabled on demand, given that it could scale into an unbounded number of metrics.
What led to the decision to enable it by default now?

matthias-horne:
> In the issue, you mentioned that we should consider putting this feature behind a flag so it can be enabled on demand, given that it could scale into an unbounded number of metrics. What led to the decision to enable it by default now?

I did take a detailed look at the exposed metrics and discussed with @rickardsjp. The following table shows the number of individual metrics that will be generated.

| Metric | Calculation | Count |
| ------ | ----------- | ------- |
| aws_ebs_csi_api_request_duration_seconds | 16 request types * 14 metrics/type | 224 |
| aws_ebs_csi_api_request_errors_total | 16 request types * 1 counter | max 16 |
| aws_ebs_csi_api_request_throttles_total | 16 request types * 1 counter | max 16 |                                                                                                                                   |
| aws_ebs_csi_ec2_detach_pending_seconds | 1 counter * number of volumes pending detachment [^1] | max number of PVC |

[^1]: only created when volume is not detached in the first attempt. Deleted as soon as volume is successfully detached. If this happens between two scrapes, the metric might never show up in Prometheus.

We concluded that the number of additional metrics should not be too much for Prometheus. Worst case the retention time might be reduced. The low risk combined with a need to timely deliver the additional metrics justifies delivering this feature without an additional flag.

hebelsan (commented):
/lgtm
/approve

PR #247, opened 2021-01-04
Add option to manage the default storage class

mvladev:
**How to categorize this PR?**

/area storage
/kind enhancement
/priority normal
/platform aws

**What this PR does / why we need it**:

New `storage.managedDefaultClass` option in the `ControlPlaneConfig` can be used enable/disable marking the `default` Storage/Volume Snapshot classes as default. This allows end-users to set the storage class to another one.

By default its set to `true` to keep the old behaviour.

**Which issue(s) this PR fixes**:

Fixes #122

**Special notes for your reviewer**:

This PR is the worked version of #123. The old PR is kept as reference.

**Release note**:

```feature user
New `storage.managedDefaultClass` in `ControlPlaneConfig` controls if the `default` storage / volume snapshot classes are marked as default by Gardener. Set it to `false` to [mark another storage / volume snapshot class as default](https://kubernetes.io/docs/tasks/administer-cluster/change-default-storage-class/) without Gardener overwriting this change. If unset, this field defaults to `true`.
```

PR #245, opened 2020-12-25
doc: add s3 permissions for aws provider

zjj2wry:
**How to categorize this PR?**

/area TODO
/kind TODO
/priority normal
/platform aws

**What this PR does / why we need it**:
fixes https://github.com/gardener/gardener-extension-provider-aws/issues/242
**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Release note**:

```other operator

```

PR #1791, opened 2026-05-05
Add EFA network interface, placement, and ML capacity block support

### LINKED ISSUE #1790: Add EFA network interface, placement group, and ML capacity block support to WorkerConfig API
/area control-plane
/kind api-change
/platform aws

Add support for configuring multiple EFA (Elastic Fabric Adapter) network interfaces, placement groups, and ML capacity blocks on AWS worker pools via the WorkerConfig API.

New API fields:
- `networkInterfaces[]`: Configure multiple network interfaces with EFA type, network card index ranges, and device index ranges
- `placement`: First-class placement group, tenancy, and dedicated host configuration
- `capacityReservation.isMLCapacityBlock`: Flag for ML capacity block market type

**Why is this needed**:

GPU instances (p4d.24xlarge, p5.48xlarge, g4dn.12xlarge, etc.) require EFA interfaces for high-bandwidth GPU-to-GPU RDMA communication. Currently, Gardener creates a single ENI per worker node with no way to configure EFA type, multiple network cards, or placement groups. This blocks AI/ML workloads that need multi-node GPU training and inference with NCCL over EFA.

shreyas-s-rao:
/area control-plane
/kind api-change
/platform aws

Adds support for configuring multiple EFA network interfaces, placement groups, and ML capacity blocks on AWS worker pools. This enables GPU instances (p4d.24xlarge, p5.48xlarge) to be provisioned with the correct EFA configuration for high-bandwidth inter-node RDMA communication.

Changes:
- API: Add IndexRange struct, DeviceIndexRange field, Placement struct, InstanceMarketOptions struct
- Validation: Type enum, range mutual exclusivity, EFA ordering, placement field checks
- Worker controller: Expand network interface ranges into MachineClass providerSpec
- Helm chart: Render placement and isMLCapacityBlock in machineclass template
- Admission webhook: Validate new fields on shoot create/update

<details>
  <summary>Release notes for machine-controller-manager-provider-aws v0.28.0</summary>
  <p><em>Sourced from <a href="https://github.com/gardener/machine-controller-manager-provider-aws/releases">github.com/gardener/machine-controller-manager-provider-aws's
  releases</a>.</em></p>
  <blockquote>
  <h2>v0.28.0</h2>
  <h1>[github.com/gardener/machine-controller-manager-provider-aws:v0.28.0]</h1>
  <h2>📰 Noteworthy</h2>
  <ul>
  <li><code>[USER]</code> <code>SpotPrice</code> is now deprecated, and will be removed in a future release. Please use <code>InstanceMarketOptions.MarketType: spot</code> instead. by <a
  href="https://github.com/shreyas-s-rao"><code>@shreyas-s-rao</code></a> [<a href="https://redirect.github.com/gardener/machine-controller-manager-provider-aws/pull/247">#247</a>]</li>
  </ul>
  <h2>✨ New Features</h2>
  <ul>
  <li><code>[USER]</code> Add <code>InstanceMarketOptions.MarketType</code> to specify market-type options such as <code>spot</code>, <code>capacity-block</code> (ML capacity block
  reservations), etc. by <a href="https://github.com/shreyas-s-rao"><code>@shreyas-s-rao</code></a> [<a
  href="https://redirect.github.com/gardener/machine-controller-manager-provider-aws/pull/247">#247</a>]</li>
  <li><code>[USER]</code> Add support for EFA network interfaces (InterfaceType, NetworkCardIndex, DeviceIndex) in EC2 instance creation by <a
  href="https://github.com/shreyas-s-rao"><code>@shreyas-s-rao</code></a> [<a href="https://redirect.github.com/gardener/machine-controller-manager-provider-aws/pull/247">#247</a>]</li>
  <li><code>[USER]</code> Add placement group, tenancy, and dedicated host configuration support via providerSpec by <a
  href="https://github.com/shreyas-s-rao"><code>@shreyas-s-rao</code></a> [<a href="https://redirect.github.com/gardener/machine-controller-manager-provider-aws/pull/247">#247</a>]</li>
  </ul>
  <h2>🐛 Bug Fixes</h2>
  <ul>
  <li><code>[USER]</code> Fix SourceDestCheck validation for EFA instances causing machines to get stuck in Uninitialized state by <a
  href="https://github.com/shreyas-s-rao"><code>@shreyas-s-rao</code></a> [<a href="https://redirect.github.com/gardener/machine-controller-manager-provider-aws/pull/247">#247</a>]</li>
  </ul>
  <h2>🏃 Others</h2>
  <ul>
  <li><code>[OPERATOR]</code> Add support for enabling instance IPv6 IMDS by <a href="https://github.com/hebelsan"><code>@hebelsan</code></a> [<a
  href="https://redirect.github.com/gardener/machine-controller-manager-provider-aws/pull/234">#234</a>]</li>
  <li><code>[OPERATOR]</code> Golang version used is now upgraded to <code>1.26.2</code> by <a href="https://github.com/takoverflow"><code>@takoverflow</code></a> [<a
  href="https://redirect.github.com/gardener/machine-controller-manager-provider-aws/pull/244">#244</a>]</li>
  <li><code>[OPERATOR]</code> Updated MCM dependency version to <code>v0.62.0</code> to enable support for machine preservation [<a
  href="https://redirect.github.com/gardener/machine-controller-manager-provider-aws/pull/245">#245</a>]</li>
  </ul>
  </blockquote>
  </details>

Fixes #1790

- Requires PRs gardener/machine-controller-manager-provider-aws#247 and gardener/aws-custom-route-controller#484 to be merged and released first
- E2E tested on g4dn.12xlarge and p4d.24xlarge with gardener remote setup

**Locally tested cases**:

- [x] g4dn.12xlarge single EFA interface creation
- [x] p4d.24xlarge multi-NIC EFA (4 cards, range expansion)
- [x] Placement group applied to EC2 instances
- [x] Single-node NCCL all-reduce (4 GPUs, g4dn)
- [x] Single-node NCCL all-reduce (8 GPUs, p4d)
- [x] Multi-node NCCL (16 GPUs, 2 p4d nodes, TCP)
- [x] Multi-node NCCL over EFA RDMA (13 GB/s peak bus bandwidth)
- [x] Custom route controller multi-NIC routing (NetworkInterfaceId)
- [x] Admission webhook negative tests:
  - [x] Invalid type rejected
  - [x] efa-only as first interface rejected
  - [x] NetworkCardIndex + NetworkCardIndexRange mutual exclusivity
  - [x] Negative networkCardIndex rejected
  - [x] NetworkCardIndexRange from < 0 rejected
  - [x] NetworkCardIndexRange from > to rejected
  - [x] DeviceIndex + DeviceIndexRange mutual exclusivity
  - [x] Negative deviceIndex rejected
  - [x] DeviceIndexRange without NetworkCardIndexRange rejected
  - [x] DeviceIndexRange from < 0 rejected
  - [x] DeviceIndexRange from > to rejected
  - [x] Mismatched range lengths rejected
  - [x] Negative ipv6AddressCount rejected
  - [x] Invalid InstanceMarketOptions.MarketType rejected
  - [x] Invalid tenancy rejected
  - [x] Invalid affinity rejected
  - [x] hostId without tenancy "host" rejected
  - [x] partitionNumber < 1 rejected

**Release note**:

```feature user
Add `networkInterfaces` field to WorkerConfig API for configuring multiple EFA network interfaces with network card index ranges and device index ranges on GPU worker pools
```

```feature user
Add `placement` field to WorkerConfig API for configuring placement groups, tenancy, dedicated hosts, and partition numbers on worker pools
```

```feature user
Add `instanceMarketOptions` field to WorkerConfig API for configuring options like `spot` instances, `capacity-block` (for targeting ML capacity block reservations), `interruptible-capacity-reservation`.
```

```other operator
Bump `machine-controller-manager-provider-aws` to `v0.28.0`.
```


MartinWeindel:
> /hold until [gardener/machine-controller-manager-provider-aws#247](https://github.com/gardener/machine-controller-manager-provider-aws/pull/247) and [gardener/aws-custom-route-controller#484](https://github.com/gardener/aws-custom-route-controller/pull/484) are merged and released.

New aws-custom-route-controller release containing required changes available with PR #1794

shreyas-s-rao:
Rebased and updated PR inline with new changes in https://github.com/gardener/machine-controller-manager-provider-aws/pull/247. 
/hold since the extension now depends directly on mcm-provider-aws module, so I had to add a replace directive in go.mod to ensure that changes/tests in this PR pass. Once mcm-provider-aws is released with the changes, I'll remove the replace directive and unhold the PR.

kon-angelo:
@shreyas-s-rao could you include some examples and documentation on how to use the new fields ?

pkg/apis/aws/validation/worker.go
```diff
@@ -353,3 +357,156 @@ func validateCpuOptions(cpuOptions *apisaws.CpuOptions, fldPath *field.Path) fie
 
 	return allErrs
 }
+
+var validNetworkInterfaceTypes = []string{
+	"",
```
kon-angelo: is that really valid ?

shreyas-s-rao: empty NI type would default to `interface` (non-EFA), so that's actually ok.

shreyas-s-rao: on second thought, you're right, it makes sense to explicitly block empty string for NI type. I'll add validation for it, thanks for pointing out.

shreyas-s-rao: Addressed in https://github.com/gardener/gardener-extension-provider-aws/pull/1791/commits/e106c7fa5d3ebaff219b338820ec07b91e39d728

shreyas-s-rao:
@kon-angelo thanks for your comments. I have now addressed them in https://github.com/gardener/gardener-extension-provider-aws/pull/1791/commits/e106c7fa5d3ebaff219b338820ec07b91e39d728 and https://github.com/gardener/gardener-extension-provider-aws/pull/1791/commits/af22ea49b279dea993b59320831c85e18acad3f8. PTAL, thanks

hebelsan:
@shreyas-s-rao, the AWS MCM just got released: https://github.com/gardener/machine-controller-manager-provider-aws/releases/tag/v0.28.0.
Could you please adjust the go.mod and images.yaml files?

pkg/apis/aws/validation/worker.go
```diff
@@ -353,3 +357,155 @@ func validateCpuOptions(cpuOptions *apisaws.CpuOptions, fldPath *field.Path) fie
...
+}
+
+func validateNetworkInterfaces(networkInterfaces []apisaws.NetworkInterface, fldPath *field.Path) field.ErrorList {
+	var allErrs field.ErrorList
+
+	for i, ni := range networkInterfaces {
+		allErrs = append(allErrs, validateNetworkInterface(ni, fldPath.Index(i))...)
+	}
+
+	if len(networkInterfaces) > 0 && networkInterfaces[0].Type != nil && *networkInterfaces[0].Type == string(ec2types.NetworkInterfaceTypeEfaOnly) {
```
hebelsan: this validation would not catch the following:
```
 networkInterfaces:
  - networkCardIndex: 1
    type: efa           # first entry, valid type → check passes
  - networkCardIndex: 0
    type: efa-only      # primary NIC is efa-only → should be rejected, but isn't
```

shreyas-s-rao: Thanks for pointing it out, done in https://github.com/gardener/gardener-extension-provider-aws/pull/1791/commits/10b0835b27f2444131490389eab8456470f4f3fc

shreyas-s-rao:
@hebelsan thanks for your review. I've addressed your comments in https://github.com/gardener/gardener-extension-provider-aws/pull/1791/commits/b6b8ab0ab3f6872e78ebfc2136e430fcd38ae8d1 and https://github.com/gardener/gardener-extension-provider-aws/pull/1791/commits/10b0835b27f2444131490389eab8456470f4f3fc, and updated the PR description and release notes. PTAL, thanks.

pkg/apis/aws/v1alpha1/types_worker.go
```diff
@@ -172,3 +182,85 @@ type CapacityReservation struct {
...
+	Type *string `json:"type,omitempty"`
+	// Description is a description for the network interface.
+	// +optional
+	Description *string `json:"description,omitempty"`
+	// SubnetID is the ID of the subnet to which the network interface should be attached.
+	// +optional
+	SubnetID *string `json:"subnetID,omitempty"`
+	// SecurityGroupIDs is a list of security group IDs to associate with the network interface.
+	// +optional
+	SecurityGroupIDs []string `json:"securityGroupIDs,omitempty"`
```
kon-angelo: ~~Do the security groups here replace completely the gardener managed ones ?~~ They do

pkg/apis/aws/v1alpha1/types_worker.go
```diff
@@ -172,3 +182,85 @@ type CapacityReservation struct {
...
+	NetworkCardIndexRange *IndexRange `json:"networkCardIndexRange,omitempty"`
+	// DeviceIndex is the device index for the network interface attachment.
+	// Mutually exclusive with DeviceIndexRange.
+	// +optional
+	DeviceIndex *int64 `json:"deviceIndex,omitempty"`
+	// DeviceIndexRange is the range of device indices. Iterates in lockstep with NetworkCardIndexRange.
+	// Must have the same length as NetworkCardIndexRange. Mutually exclusive with DeviceIndex.
+	// Can only be specified when NetworkCardIndexRange is set.
+	// +optional
+	DeviceIndexRange *IndexRange `json:"deviceIndexRange,omitempty"`
```
kon-angelo: I am not sold out on the idea of an index range. Why do you think that on top level gardener resource this is an acceptable abstraction but not for the MCM ? I mean why delegate a templating feature on this object level (something that can be trivially solved by helm on customer side with actual guarantee about what is produced)

pkg/apis/aws/v1alpha1/types_worker.go
```diff
@@ -172,3 +182,85 @@ type CapacityReservation struct {
...
+	NetworkCardIndex *int64 `json:"networkCardIndex,omitempty"`
+	// NetworkCardIndexRange is the range of network card indices.
+	// Mutually exclusive with NetworkCardIndex.
+	// +optional
+	NetworkCardIndexRange *IndexRange `json:"networkCardIndexRange,omitempty"`
+	// DeviceIndex is the device index for the network interface attachment.
+	// Mutually exclusive with DeviceIndexRange.
+	// +optional
+	DeviceIndex *int64 `json:"deviceIndex,omitempty"`
+	// DeviceIndexRange is the range of device indices. Iterates in lockstep with NetworkCardIndexRange.
```
kon-angelo: If they are ineed in lockstep why do we need both a device index and a nic index range ? Why not group them in a struct ?

docs/usage/usage.md
```diff
@@ -865,6 +865,99 @@ The exact behaviour depends on the combination of the targeting and the reservat
...
+      machine:
+        type: p4d.24xlarge
+      ...
+      providerConfig:
+        apiVersion: aws.provider.extensions.gardener.cloud/v1alpha1
+        kind: WorkerConfig
+        networkInterfaces:
+        - networkCardIndex: 0
+          deviceIndex: 0
+          type: efa                    # primary NIC on card 0 (must be interface or efa)
```
kon-angelo: Should we really allow customization of the primary NIC ? Does this not break basic cluster functionality e.g. internet connectivity and so on ?

pkg/apis/aws/v1alpha1/types_worker.go
shreyas-s-rao: Yes. Would you prefer that users not be able to specify `SecurityGroupIDs` at all, so that it always uses the gardener-managed one?

pkg/apis/aws/v1alpha1/types_worker.go
shreyas-s-rao: There are certain GPU instances that can go up to 68 network interfaces per instance ([ref](https://docs.aws.amazon.com/ec2/latest/instancetypes/ac.html#ac_network)), and we wouldn't want users to have to specify every single one of them in the shoot spec (that would blow up the shoot spec size, plus it's inconvenient). Hence the range approach. When the NI specification is largely deterministic (continuously increasing indices for device and NIC), so such range structs would actually help.

docs/usage/usage.md
shreyas-s-rao: [AWS EFA setup guide](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/efa-start.html#efa-start-instances) itself specifies setting the first network interface to `EFA` to actually benefit from underlying EFA network fabric instead of the regular ENI.

pkg/apis/aws/v1alpha1/types_worker.go
shreyas-s-rao: You mean something like:
```yaml
indexRange:
  networkCardIndexFrom: 1
  deviceIndexFrom: 2
  count: 30
```

This would still provide users to specify a skew for the indices, while ensuring the count of both ranges would be the same.

pkg/apis/aws/v1alpha1/types_worker.go
kon-angelo: I actually don't know enough about the use case and if these NICs or the additional NICs need to have special NSGs. There is a reason that the NSG ids were not exposed before and that is to prevent misconfigurations. In this case, you enforce essentially that the primary NIC has to have correct rules for subnet traffic, LB traffic etc. However none of this is checked or even documented currently.

See my proposal for BYO-infrastructure: https://github.com/gardener/gardener-extension-provider-aws/pull/1715 where I talk about it and you essentially introduce the same issue with regards to NSGs (at least for primary nic).

pkg/apis/aws/v1alpha1/types_worker.go
kon-angelo: What does the user need to know to calculate the skew ? Say I have additional nics and additional data disks and so on. How is this calculated and if it can be calculated why can't we calculate it for the user.

pkg/apis/aws/v1alpha1/types_worker.go
shreyas-s-rao: Fair point. Even I'm not sure how the SGs might be used by users, because in all my tests, using the gardener-supplied SG was sufficient (it would also make sense, in order to allow inter-node communication within the same gardener-provided SG rather than users having to manually configure their new SG with rules allowing traffic in and out of their SG to and from the gardener-managed SG).

So I can essentially remove this for now, and later if any user raises this concern for their usecase, then we can see how to introduce it in a limited way. WDYT?

shreyas-s-rao: Or I could enforce gardener-managed SG for primary NIC (non `efa-only`), and allow configuring the SG for `efa-only` NICs. I'm fine wither way.

pkg/apis/aws/v1alpha1/types_worker.go
shreyas-s-rao: Traditionally the deviceIndex = NICIndex. I honestly don't know if users would want to configure it differently at all. That's the reason I allowed for the skew. But again, we can introduce this without a skew, ie, enforce deviceIndex = NICIndex, and later if users have an explicit need for skew or different kind of configurability, then we can introduce it.  WDYT?

So that would look like:
```yaml
indexRange: # applies to deviceIndices and NICIndices
  from: 1
  to: 3
```

kon-angelo: I would go simple before there is a need for something more complex. At least the machines are created deterministically by us hence I am not sure what the user could have decided on his own here. Unless there are very special machine types that require that esoteric knowledge.


even if there is an offset in the lockstep, wouldn't we know about it when we create the machineclass ? What cases would we now know about it?

pkg/apis/aws/v1alpha1/types_worker.go
kon-angelo: Okay, I can see the reasoning now. Still handling ranges in these APIs doesn't feel great.

I am not sure if a NICTemplate, approach would be better to reduce the boilerplate. At least having the range as part of the NIC struct feels not correct from an API perspective. Still I lack a better suggestion at the time

pkg/apis/aws/v1alpha1/types_worker.go
```diff
@@ -172,3 +182,85 @@ type CapacityReservation struct {
...
+	AssociatePublicIPAddress *bool `json:"associatePublicIPAddress,omitempty"`
+	// DeleteOnTermination indicates whether the network interface should be deleted when the instance is terminated.
+	// +optional
+	DeleteOnTermination *bool `json:"deleteOnTermination,omitempty"`
+	// Ipv6AddressCount is the number of IPv6 addresses to assign to the network interface.
+	// +optional
+	Ipv6AddressCount *int64 `json:"ipv6AddressCount,omitempty"`
+	// PrimaryIpv6 indicates whether the first IPv6 address will be the primary IPv6 address.
+	// +optional
+	PrimaryIpv6 *bool `json:"primaryIpv6,omitempty"`
```
kon-angelo: Which of these fields are core to your scenario and are some of them added just because they are available as options ?

- For example in what cases do we not want to delete on termination ? 
- Why expose `AssociatePublicIPAddress` which is generally a security risk ? 
- How does `Ipv6AddressCount` work if specifying ranges different (less then the pod mask) for the primary NIC

pkg/apis/aws/v1alpha1/types_worker.go
shreyas-s-rao: Hmm, then we can stick to the suggestion [here](https://github.com/gardener/gardener-extension-provider-aws/pull/1791#discussion_r3282275362), so we don't allow skew to begin with (deviceIndex = NICIndex for every NI). Simple enough to begin with, can be enhanced later if at all there's a concrete need from users. WDYT?

pkg/apis/aws/v1alpha1/types_worker.go
shreyas-s-rao: Right now, I'm honestly not sure when we would actually need an offset, or whether it would even be necessary. I'll switch to the simpler API that I described above then, to begin with.

pkg/apis/aws/v1alpha1/types_worker.go
shreyas-s-rao: Good points:
- `AssociatePublicIPAddress` can be set to false by default since we actually don't want instances to be accessible publicly but rather through bastion host. I'll remove this field and always set it to `false`
- `DeleteOnTermination`: I think this can be set to true be default to tie the NI lifecycle to the instance lifecycle, to avoid orphaned resources.
- `Ipv6AddressCount` and `PrimaryIpv6`: I haven't tested this properly, so I'm not sure whether this would be a necessity for shoots with dual-stack. I can remove this field for now, and introduce it later if users have a concrete need for it. WDYT?

The other fields are necessary, but these 4 can be removed for now.

pkg/apis/aws/v1alpha1/types_worker.go
shreyas-s-rao: Done in https://github.com/gardener/gardener-extension-provider-aws/pull/1791/commits/002b55779ef86f04273ead8539305c8c51d6a38c. PTAL.

pkg/apis/aws/v1alpha1/types_worker.go
shreyas-s-rao: After checking https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/efa-acc-inst-types.html again, I realised that lockstep is not really used for EFA, and only deviceIndex 0 or 1 are used. So I've now gotten rid of `DeviceIndexRange` entirely in https://github.com/gardener/gardener-extension-provider-aws/pull/1791/commits/41b6755c4730cc71cf426e94ee9e341e492ff177. This simplifies it vastly, since we only allow range for `networkCardIndex` (which allows for easy expansion), while `deviceIndex` is fixed for the entire range, which is what AWS also recommends.

pkg/apis/aws/v1alpha1/types_worker.go
shreyas-s-rao: Check https://github.com/gardener/gardener-extension-provider-aws/pull/1791#discussion_r3289746644

docs/usage/usage.md
shreyas-s-rao: Two guides from AWS recommend setting either ENA or EFA for the first NIC, so I'm a bit confused here. I'd rather leave this configurable, since AWS has allowed setting EFA for the first NIC via the API, and also recommends it in the above-mentioned guide.

pkg/apis/aws/v1alpha1/types_worker.go
shreyas-s-rao: Done in https://github.com/gardener/gardener-extension-provider-aws/pull/1791/commits/db598bfc293dfeded999b2681cfb069012d494ec. PTAL, thanks.

shreyas-s-rao:
@kon-angelo thanks for your extensive review. I've now addressed them and pushed the changes. PTAL, thanks.

Ultimately, the goal is to allow configuring the network interfaces to look like the ones described in this [example guide](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/efa-acc-inst-types.html).

kon-angelo (commented):
/lgtm

shreyas-s-rao:
Thanks for the LGTM @kon-angelo @hebelsan . Can the hold on this PR be removed?

PR #1799, opened 2026-05-06
feat(vpc): support Ipv6CidrBlock

### LINKED ISSUE #1798: Support `Ipv6CidrBlock`
/area networking
/kind enhancement
/platform aws

Support selecting specific IPv6 CIDR from the referenced `IPv6IPAMPool`.  This can be achieved by setting the `Ipv6CidrBlock` option instead of `Ipv6NetmaskLength`. 

**Why is this needed**:
This is a follow-up to #1048. 

Users may want to use a single IPAM pool for multiple VPCs, for example deploying multiple regional clusters that all use the same top-level contiguous address space. 

When clusters are deleted and recreated (e.g. due to maintenance, changes in immutable fields, etc.), there's a risk that the original CIDR would not be re-allocated to the recreated cluster, disrupting static Ingress IPs and (for dual stack clusters) static egress pod IP ranges. 

Users can specify the required CIDR, rather than let AWS determine it, to allow consistent allocation of CIDRs from a shared pool. 

This feature is aligned with other options that exist in Gardener today, such as using static NAT IPv4 addresses.

guydc:
/area networking
/kind enhancement
/platform aws

Fixes #1798 

**Release note**:
```feature user
Add ability to specify IPv6CIDR from an IPAM pool
```

### DIFF
--- a/docs/usage/ipv6.md
+++ b/docs/usage/ipv6.md
@@ -184,11 +184,11 @@ Use this option when you operate centralized IP address management and need dete
 
 #### Pinning a Specific IPv6 CIDR
 
-When replacing a cluster (e.g. blue/green rollout or disaster recovery), AWS Network Load Balancers derive their IPv6 addresses from the VPC's `/56` CIDR. If a new cluster gets a different `/56`, the NLB IPv6 addresses change — breaking DNS records, firewall rules, and client allowlists that reference those addresses.
+When replacing a cluster, AWS Network Load Balancers derive their IPv6 addresses from the VPC's `/56` CIDR. If a new cluster gets a different `/56`, the NLB IPv6 addresses change and any DNS records, firewall rules, or client allowlists that reference those addresses will need to be updated.
 
 To keep NLB IPv6 addresses stable across cluster replacements, you can pin a specific `/56` block from the IPAM pool using the `infrastructureConfig.networks.vpc.ipv6CidrBlock` field. When the replacement cluster is provisioned with the same `/56` (once the previous cluster's VPC has been deleted and returned the block to the pool), the NLBs receive the same IPv6 addresses.
 
-Pinning the `/56` also makes pod egress IP ranges deterministic: in IPv6-only and dual-stack clusters, pods egress to the internet using addresses from the VPC's `/56` via the egress-only internet gateway. A stable `/56` means external firewall rules, partner allowlists, and compliance policies that reference pod source IPs by range remain valid across cluster replacements.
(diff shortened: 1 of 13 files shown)


docs/usage/ipv6.md
```diff
@@ -177,11 +177,38 @@ provider:
...
+provider:
+  type: aws
+  infrastructureConfig:
+    apiVersion: aws.provider.extensions.gardener.cloud/v1alpha1
+    kind: InfrastructureConfig
+    networks:
+      vpc:
+        ipv6IpamPool:
+          id: ipam-pool-0123456789abcdef0
+        ipv6CidrBlock: "2600:1f13:abcd::/56"
```
hebelsan: Wouldn't it make sense to have the new cidr field `ipv6CidrBlock` under `ipv6IpamPool` and just title it `cidrBlock`

docs/usage/ipv6.md
```diff
@@ -177,11 +177,38 @@ provider:
...
+...
+```
+
+Requirements & notes:
+* `ipv6IpamPool` must also be set; `ipv6CidrBlock` cannot be used without it.
```
hebelsan:
```suggestion
* `ipv6IpamPool.id` must also be set; `ipv6CidrBlock` cannot be used without it.
```

pkg/controller/infrastructure/infraflow/reconcile.go
```diff
@@ -209,7 +209,11 @@ func (c *FlowContext) ensureManagedVpc(ctx context.Context) error {
 		if c.config.Networks.VPC.Ipv6IpamPool != nil && c.config.Networks.VPC.Ipv6IpamPool.ID != nil {
 			desired.AssignGeneratedIPv6CidrBlock = false
 			desired.Ipv6IpamPoolId = c.config.Networks.VPC.Ipv6IpamPool.ID
-			desired.Ipv6NetmaskLength = ptr.To(int32(defaultIPv6NetmaskSize))
+			if c.config.Networks.VPC.Ipv6CidrBlock != nil {
+				desired.Ipv6CidrBlock = c.config.Networks.VPC.Ipv6CidrBlock
+			} else {
+				desired.Ipv6NetmaskLength = ptr.To(int32(defaultIPv6NetmaskSize))
```
hebelsan: I know this wasn't introduced by you but let's improve it now:
```suggestion
				desired.Ipv6NetmaskLength = ptr.To[int32](defaultIPv6NetmaskSize)
```

hebelsan (commented):
Thanks for the contribution!
I've some small comments.
Could you please also add an example to the file 30-infrastructure.yaml?

hebelsan (commented):
/lgtm
/approve

PR #1801, opened 2026-05-07
Remove controllers flag from deployment configuration

hebelsan:

/area quality
/kind enhancement
/platform aws

In the context of https://github.com/gardener/gardener/pull/14628, we need to enable the BackupEntry Controller in provider extensions. Currently, it's disabled, so we have to introduce a feature gate (with https://github.com/gardener/gardener/pull/14749), otherwise this won't work and the BackupEntry never gets reconciled.

Fixes #

**Release note**:

```other operator
Remove controllers flag from deployment configuration
```


matthias-horne (commented):
/lgtm

matthias-horne:
/lgtm cancel

Enabling all controllers in the runtime cluster will fail for several reasons:
- The service account does not have RBAC permissions to watch the resources for the newly enabled controllers in the runtime cluster
- The resources might not even exist, if the runtime cluster is not also a seed

hebelsan:
While the reconciliations are scoped, the controllers still register a watch on all resources. 
This fails either because the CRDs don't even exist in the runtime cluster or because the service account does not have permissions.
The controller pods are continuously crashing in the runtime cluster.

PR #1802, opened 2026-05-08
fix EFFECTIVE_VERSION in Makefile to only depend on repo status

matthias-horne:

/area dev-productivity
/kind bug
/platform aws

The new calculation of `EFFECTIVE_VERSION` does not depend on time and only on the state of the files in the repo. This is required to allow running `make helm-charts` and `make helm-push` in subsequent commands.

Fixes #

**Release note**:

```other operator

```


hebelsan (approved):
/lgtm

PR #1805, opened 2026-05-11
Improve wording for enableMTUCustomizer flag default behavior

benedictweis:

/area documentation
/kind enhancement
/platform aws

Fixes an inconsistency in the enableMTUCustomizer documentation: https://github.com/gardener/gardener-extension-provider-aws/blob/0bc056fd8d1238e2576337a4496b651cd7020473/docs/usage/usage.md?plain=1#L291

The sentence states that the flag is set to `false` for newly created shoots but `true` for newly created shoots. The way the sentence is worded makes me think that it should be `true` in both cases. This would also align with https://github.com/gardener/gardener-extension-provider-aws/pull/1732.

None

None

**Release note**:

```other operator
NONE
```

### DIFF
--- a/docs/usage/usage.md
+++ b/docs/usage/usage.md
@@ -291,7 +291,7 @@ It only applies for those worker pools whose `iamInstanceProfile` is not set.
 The `enableMTUCustomizer` flag controls whether a systemd unit and script are deployed to the shoot worker nodes that set the MTU of all non-virtual network interfaces to `1460`.
 This is a legacy mechanism from a time when CNIs lacked automatic MTU detection. Modern CNIs detect and configure the MTU automatically, so new clusters should explicitly set this flag to `false`.
 It may still be useful in environments where the default AWS MTU of `9001` causes connectivity issues with peers that have a lower MTU (e.g. `1500`) and the CNI does not handle this automatically.
-If the flag is not provided, the admission webhook defaults it to `false` for newly created shoots. For existing shoots the field remains unset, which also defaults to `true` for backwards compatibility.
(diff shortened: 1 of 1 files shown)


benedictweis:
@axel7born, you might want to take a look at this, since you introduced the documentation with https://github.com/gardener/gardener-extension-provider-gcp/pull/1399. Also let me know if my line of thinking is wrong and the flag is indeed set to `false` for newly created shoots.

voelzmo:
🤔 So I think this might be worded a bit complicated, but my understanding is the following
* if you create a new Shoot now, the MTU customizer is `disabled`
* if you have an existing Shoot (which doesn't know/have this field in its spec), the MTU customizer remains `enabled`, although the flag isn't explicitly set

This seems to be supported by the two test cases for the defaulting behavior: https://github.com/gardener/gardener-extension-provider-aws/pull/1732/changes/21afbe32df04965fc535c678c8bf9f482b71b132#diff-3faf384dc3bdf082a5ff6d4d0a8b590f2faeb78f636f8b41d0eefe57d099d8c0R372-R402

So, if you want the MTU customizer to be switched off on all Shoots, you
* For existing Shoots: need to set the property `enableMTUCustomizer: false` explicitly
* For new Shoots: don't need to do anything, as the webhook takes care to set `enableMTUCustomizer: false` for new Shoots automatically

benedictweis:
Alright, this behavior makes sense to me. I have adjusted the PR to change the wording to make this part more clear. Do you think that provides any benefit or do we just leave the wording as is?

PR #1813, opened 2026-05-15
Fix byo vpc validation

### LINKED ISSUE #1811: VPC CIDR validation incorrectly requires single CIDR to cover all subnets (multi-CIDR VPC)
/area control-plane
/kind bug
/platform aws

**What happened**:

Infrastructure reconciliation fails with:

```
networks.vpc.id: Invalid value: "vpc-xxx": provided VPC has no CIDR block matching the clusters configured networks. There must be a CIDR block that is a superset of all subnet ranges.
```

This occurs on a shoot cluster using a bring-your-own-VPC with multiple CIDR associations where subnets are intentionally spread across different VPC CIDRs:
- VPC primary CIDR: `10.238.180.128/27` → covers workers (`10.238.180.144/28`) and internal (`10.238.180.128/28`)
- VPC secondary CIDR: `10.62.192.0/28` → covers public (`10.62.192.0/28`)

Both CIDR associations are in `associated` state. The cluster ran successfully with this configuration for over a year.

Validation should pass because each subnet is covered by at least one VPC CIDR block. The check should verify per-subnet coverage across all CIDRs, not require a single CIDR to be a superset of all subnets.

1. Create a VPC with two CIDR blocks (e.g., `10.238.180.128/27` and `10.62.192.0/28`)
2. Configure a shoot with BYO-VPC referencing that VPC ID
3. Assign zone subnets across both CIDRs (workers/internal in first CIDR, public in second)
4. Trigger infrastructure reconciliation

The bug is in `pkg/controller/infrastructure/configvalidator.go` lines 160-184, introduced by PR #1747.

The current logic:
```go
haveValidCidr := slices.ContainsFunc(cidrsToConsider, func(cidr string) bool {
    return validateZoneSubnetCIDRs(networkingPath, cidr, infraConfig).ToAggregate() == nil
})
```

`validateZoneSubnetCIDRs` checks if ALL subnets fit within a single CIDR. The outer loop asks "is there at least one CIDR where all subnets pass?" — this fails when subnets are intentionally distributed across multiple VPC CIDRs.

**Suggested fix**: Instead of checking if one CIDR covers everything, check that each individual subnet is covered by at least one VPC CIDR:

```go
for _, zone := range infraConfig.Networks.Zones {
    for _, subnet := range []cidrvalidation.CIDR{...} {
        covered := slices.ContainsFunc(cidrsToConsider, func(cidr string) bool {
            vpcCIDR := cidrvalidation.NewCIDR(cidr, fldPath)
            return vpcCIDR.ValidateSubset(subnet).ToAggregate() == nil
        })
        if !covered {
            // error for this specific subnet
        }
    }
}
```

**Environment**:

- Gardener version: v1.34.6
- Extension version: version deployed on canary landscape (post PR #1747 merge of 2026-04-14)
- Kubernetes version: N/A (validation issue, not runtime)
- Cloud provider: AWS (eu-central-1, BYO-VPC with multiple CIDR associations)

AndreasBurger:

/area control-plane
/kind bug
/platform aws

**What this PR does / why we need it**: Previous validation assumed that all subnets need to be covered by one associated CIDR of the provided VPC. With this PR, subnets are verified independent from each other in the BYO case.
I've also updated some of the field-paths to be more precise.

Fixes #1811 🤞

**Special notes for your reviewer**: Not yet tested

**Release note**:

```bugfix user
Validation no longer assumes all subnets of a zone must be covered by a single VPC cidr
```

### DIFF
--- a/pkg/controller/infrastructure/configvalidator.go
+++ b/pkg/controller/infrastructure/configvalidator.go
@@ -70,7 +70,7 @@ func (c *configValidator) Validate(ctx context.Context, infra *extensionsv1alpha
 	// Validate infrastructure config
 	if config.Networks.VPC.ID != nil {
 		logger.Info("Validating infrastructure networks.vpc.id")
-		allErrs = append(allErrs, c.validateVPC(ctx, field.NewPath("networks", "vpc", "id"),
+		allErrs = append(allErrs, c.validateVPC(ctx, field.NewPath("networks"),
 			awsClient, *config, *config.Networks.VPC.ID, infra.Spec.Region)...)
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/controller/infrastructure/configvalidator.go
```diff
@@ -102,100 +102,103 @@ func (c *configValidator) validateVPC(
...
+		for _, subnetCIDR := range []cidrvalidation.CIDR{
+			cidrvalidation.NewCIDR(zone.Workers, zonesPath.Index(i).Child("nodes")),
+			cidrvalidation.NewCIDR(zone.Public, zonesPath.Index(i).Child("public")),
+			cidrvalidation.NewCIDR(zone.Internal, zonesPath.Index(i).Child("internal")),
+		} {
+			if !isSubnetCIDRContainedInAnyCIDR(vpcCIDRs, subnetCIDR) {
+				allErrs = append(allErrs, field.Invalid(
+					subnetCIDR.GetFieldPath(),
+					subnetCIDR.GetCIDR(),
+					fmt.Sprintf("Subnet %q not contained in any of the associated CIDRs of the given VPC", subnetCIDR.GetFieldPath()),
```
hebelsan:
```suggestion
					fmt.Sprintf("subnet CIDR %q is not contained in any of the VPC's associated CIDR blocks", subnetCIDR.GetCIDR()),
```

hebelsan (commented):
/lgtm
/approve

PR #1815, opened 2026-05-18
Bring extension main Pod back under VPA control for CPU

voelzmo:
/area control-plane
/kind bug
/platform aws

PR #1448 accidentally removed CPU from the VPA `controlledResources` for the main extension pod, leaving it only managing memory. In practice, the default CPU requests are too low compared to real-world usage, so this restores CPU autoscaling by adding `cpu` back to `controlledResources`.

Additionally, the wildcard `containerName: '*'` policy is replaced with an explicit policy for the `gardener-extension-provider-aws`  container plus a catch-all `'*'` policy with `mode: "Off"`. This follows the Gardener project's VPA best practices ([defining a VPA  resource](https://gardener.cloud/docs/guides/applications/shoot-pod-autoscaling-best-practices/#defining-a-vpa-resource)), which  recommend explicitly disabling VPA for all containers not under active control to prevent unintended scaling of injected sidecars.

Fixes #

**Release note**:
```bugfix operator
VPA for the extension pod now controls CPU requests again (accidentally removed in #1448). The VPA container policy also explicitly  disables autoscaling for any non-main containers following Gardener best practices.```


voelzmo:
fyi @oliver-goetz

hebelsan (commented):
/lgtm
/approve

PR #1819, opened 2026-05-22
[GEP-26] Cleanup usage of the deprecated `shoot.spec.dns.providers[].secretName`

vpnachev:
/area security ipcei
/kind cleanup
/label ipcei/workload-identity
/platform aws

Cleanup usage of the deprecated `shoot.spec.dns.providers[].secretName`

Fixes #

Follow-up on https://github.com/gardener/gardener-extension-provider-aws/pull/1730
Part of https://github.com/gardener/gardener/issues/9586

**Release note**:

```breaking operator
⚠️ This extension no longer support Gardener installation running with `github.com/gardener/gardener < v1.135.0`, kindly update `github.com/gardener/gardener` to version `>= v1.135.0` before updating the extension.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ The following lists known compatibility issues of this extension controller with
 | AWS Extension | Gardener | Action | Notes |
 | ------------- | -------- | ------ | ----- |
 | `<= v1.15.0` | `> v1.10.0` | Please update the provider version to `> v1.15.0` or disable the feature gate `MountHostCADirectories` in the Gardenlet. | Applies if feature flag `MountHostCADirectories` in the Gardenlet is enabled. Shoots with CSI enabled (Kubernetes version >= 1.18) miss a mount to the directory `/etc/ssl` in the Shoot API Server. This can lead to not trusting external Root CAs when the API Server makes requests via webhooks or OIDC. |
-| `>= v1.70.0` | `< v1.135.0` | Please update Gardener to version `>= v1.135.0` | The shoot API field `shoot.spec.dns.providers[].secretName` has been deprecated in favor of `shoot.spec.dns.providers[].credentialsRef` which is available in Gardener `>= v1.135.0`, the provider extension is migrated to use the new field only in `v1.70.0`. |
+| `>= v1.71.0` | `< v1.135.0` | Please update Gardener to version `>= v1.135.0` | The shoot API field `shoot.spec.dns.providers[].secretName` has been deprecated in favor of `shoot.spec.dns.providers[].credentialsRef` which is available in Gardener `>= v1.135.0`, the provider extension is migrated to use the new field only in `v1.71.0`. |
 
 ----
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -35,8 +35,10 @@ Please take a look [here](https://github.com/gardener/gardener/blob/master/docs/
...
+| ------------- | -------- | ------ | ----- |
+| `<= v1.15.0` | `> v1.10.0` | Please update the provider version to `> v1.15.0` or disable the feature gate `MountHostCADirectories` in the Gardenlet. | Applies if feature flag `MountHostCADirectories` in the Gardenlet is enabled. Shoots with CSI enabled (Kubernetes version >= 1.18) miss a mount to the directory `/etc/ssl` in the Shoot API Server. This can lead to not trusting external Root CAs when the API Server makes requests via webhooks or OIDC. |
+| `>= v1.71.0` | `< v1.135.0` | Please update Gardener to version `>= v1.135.0` | The shoot API field `shoot.spec.dns.providers[].secretName` has been deprecated in favor of `shoot.spec.dns.providers[].credentialsRef` which is available in Gardener `>= v1.135.0`, the provider extension is migrated to use the new field only in `v1.71.0`. |
```
wpross: nit: maybe better use "as of" or "from ... onwards" instead of "only in".

wpross (changes requested):

README.md
wpross: I would also add the secretName to credentialsRef migration step in the Release notes, something like

```
Before updating:
  1. Update Gardener to `>= v1.135.0`
  2. Migrate all Shoot DNS configurations to use `spec.dns.providers[].credentialsRef` instead of
  `secretName`
```

vpnachev: On the one side end users does not control gardener or provider-aws versions, on the other side gardener operators should not migrate end-user shoot specs. I mean, adding "migrate all shoot DNS configurations ..." in the release note is not correct because such migration would be responsibility of the end user, while extension update is outside of their control. 

Anyway, Gardener API server is taking care to keep both fields in sync when the referred credential is of type Secret, so there is no immediate action item to the shoot owners. They will be getting warnings from the API server until they are setting the `secretName` field, and it will be disallowed to update to k8s 1.36 if the field is still set.

tl;dr- I will not change the release note

wpross (approved):
/lgtm

wpross:
I'm wondering if we should do this cleanup only once k8s `v1.34` will not be supported anymore? So basically shouldn't it be possible to still use `secretName` for g/g versions <= `v1.134`?
Sry for asking this question after approving the PR, the thoughts just came to me.

vpnachev:
g/g v1.134 is not supported since v1.137 got released which happened on February 26, operators using the extension are expected to update their g/g regularly, not only the provider extension. So, I do not think the extension should keep compatibility with so old g/g versions. 


refs: 
- https://github.com/gardener/gardener/blob/master/docs/deployment/version_skew_policy.md#supported-gardener-versions
- https://github.com/gardener/gardener/releases/v1.137.0

PR #1833, opened 2026-06-05
Add self-referencing egress rule for EFA-enabled worker pools

### LINKED ISSUE #1832: Add self-referencing security group egress rule when EFA network interfaces are configured

/area networking
/area security
/kind enhancement
/platform aws

The infrastructure controller should add a self-referencing security group egress rule to allow all traffic to itself whenever any worker pool in the shoot has at least one EFA-enabled network interface (`networkInterfaces[].type` is `efa` or `efa-only`).

**Why is this needed**:

EFA workloads do not function without this rule. #1791 (in response to #1790) added the API for EFA NICs, but configuring a worker pool with `networkInterfaces[].type: efa` (or `efa-only`) is currently insufficient on its own, since every NCCL/MPI workload that exercises EFA RDMA fails with `Error 4126: Unresponsive receiver (reachable by EFA device but handshake failed)` until the self-referencing egress rule is added.

Since AWS evaluates CIDR-based and security-group-id-based rules at different layers, and EFA traffic is only matched by the latter, this new security group rule is necessary, even though the security group already has `egress 0.0.0.0/0`, and that's a superset of `egress to self`.

Investigation into this error yielded the following observations:

1. AWS explicitly mandates this rule:
    [Get started with EFA, Step 1: "Prepare an EFA-enabled security group"](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/efa-start.html#efa-start-security):
    
    > An EFA requires a security group that allows all inbound and outbound traffic `to and from the security group itself`.
    
    The procedure walks the user through adding both an inbound and an outbound rule whose source/destination is set to the security group's own ID. It is the only documented way to allow EFA traffic (which was initially non-obvious from the documentation).
    
    The same recommendation appears in:
    - [AWS Parallel Computing Service: Create a security group to support EFA communications](https://docs.aws.amazon.com/pcs/latest/userguide/working-with_networking_efa_create-sg.html): uses `--source-group <sg-id>` for both ingress and egress, with `--protocol -1` (all traffic).
    - [AWS Deep Learning AMI: Prepare an EFA-Enabled Security Group](https://docs.aws.amazon.com/dlami/latest/devguide/tutorial-efa-launching.html#tutorial-efa-security-group): "EFA requires a security group that allows all inbound and outbound traffic to and from the security group itself."
    
2. EFA traffic bypasses the IP/CIDR enforcement path:
    EFA uses the **Scalable Reliable Datagram (SRD)** protocol, which is non-IP. From [Elastic Fabric Adapter basics](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/efa.html#efa-basics):
    > "The Libfabric API **bypasses the operating system kernel** and communicates directly with the EFA device to put packets on the network."
    > "The EFA device provides capabilities like built-in **OS-bypass** and congestion control through the Scalable Reliable Datagram (SRD) protocol."

    And from [EFA limitations](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/efa.html#efa-limits):
    > "EFA traffic is **not routable**. Normal IP traffic from the ENA device of an EFA interface remains routable."

    The consequence is that SRD packets are not IP packets and are not processed by the VPC's IP routing/CIDR-matching layer. The only way the VPC fabric can authorize EFA egress is by matching on the destination ENI's identity (i.e., "the destination ENI is a member of security group `sg-xxxxx`"). That match is expressed through a security-group-id-referenced rule instead of a CIDR rule. So `egress 0.0.0.0/0` rule is apparently inapplicable to SRD frames.

3. Empirical confirmation:
    With only `egress 0.0.0.0/0` on the worker SG (without self-referencing egress):
    
    ```
    NCCL WARN NET/OFI Request ... completed with error. RC: 103. Flags: 2050.
    Error: 4126 (Unresponsive receiver (reachable by EFA device but handshake failed)
    My EFA addr: fi_addr_efa://[fe80::...]:5:...  My host id: i-...
    Peer EFA addr: fi_addr_efa://[fe80::...]:5:... Peer host id: N/A).
    ```
    
    So EFA address discovery (which goes through link-local IPv6 over the ENA path) succeeded because that traffic is covered by the `0.0.0.0/0` egress, but the SRD-layer handshake (which goes through the EFA device) seems to be dropped at the ENI boundary.
    
    After adding `egress IpProtocol=-1 UserIdGroupPairs=[{GroupId=<self>}]` on the worker SG, the handshake completed and 16-GPU NCCL all-reduce test ran successfully.

In conclusion, after careful testing and deliberation, the intuition that `egress 0.0.0.0/0` is a superset turned out to be incorrect. As described above, CIDR rules and SG-id rules are evaluated at different layers in the AWS network stack and EFA traffic is only authorized by the latter. This is non-obvious from the AWS documentation, and was missed in #1791 since I had manually added this rule during testing and totally forgot to make the change in the reconciler code. So it needs to be added now.

**Sources**:

- AWS EFA - Get started: [Step 1: Prepare an EFA-enabled security group](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/efa-start.html#efa-start-security)
- AWS EFA - Limitations and basics: [Elastic Fabric Adapter for AI/ML and HPC workloads](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/efa.html)
- AWS Parallel Computing Service: [Create a security group to support EFA communications](https://docs.aws.amazon.com/pcs/latest/userguide/working-with_networking_efa_create-sg.html)
- AWS Deep Learning AMI: [Prepare an EFA-Enabled Security Group](https://docs.aws.amazon.com/dlami/latest/devguide/tutorial-efa-launching.html#tutorial-efa-security-group)

shreyas-s-rao:

/area networking
/area security
/kind bug
/platform aws

The infrastructure controller now adds a self-referencing security group egress rule (`protocol: -1`, source/destination = the worker security group itself) to the shoot's worker security group whenever any worker pool has an EFA-enabled network interface (`networkInterfaces[].type` of `efa` or `efa-only`). More details at #1832 .

The rule is only added when at least one worker pool actually requests EFA, so non-AI/HPC shoots are unaffected.

Fixes #1832 

Continuation (fix) for #1790 

**Release note**:

```bugfix operator
The AWS infrastructure controller now adds a self-referencing security group egress rule to the worker security group when any worker pool has an EFA-enabled network interface (`networkInterfaces[].type: efa` or `efa-only`). This rule is required for EFA's SRD traffic, which is not covered by the existing `egress 0.0.0.0/0` rule.
```

### DIFF
--- a/docs/usage/usage.md
+++ b/docs/usage/usage.md
@@ -912,6 +912,10 @@ This expands to 33 NICs total: 1 primary `interface` on card 0 device 0, 1 `efa-
 
 For more details on EFA, see the [AWS EFA documentation](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/efa.html). For instance-specific recommended NIC layouts, see [Maximize network bandwidth on EC2 instances with multiple network cards](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/efa-acc-inst-types.html).
 
+### Security group
+
+When at least one worker pool has an EFA-enabled network interface (`type: efa` or `type: efa-only`), the infrastructure controller automatically adds a self-referencing egress rule (`protocol: -1`, source/destination = the worker security group itself) to the shoot's worker security group. This is required because EFA traffic uses the Scalable Reliable Datagram (SRD) protocol, which is not authorized by CIDR-based rules. AWS evaluates CIDR rules and security-group-id rules at different layers, and EFA traffic is only matched by the latter, even though the security group already allows `egress 0.0.0.0/0`. See AWS's [Get started with EFA, Step 1: Prepare an EFA-enabled security group](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/efa-start.html#efa-start-security) for details.
+
 
 ## Placement
 
--- a/pkg/apis/aws/helper/scheme.go
+++ b/pkg/apis/aws/helper/scheme.go
@@ -9,8 +9,10 @@ import (
 	"errors"
 	"fmt"
 
+	ec2types "github.com/aws/aws-sdk-go-v2/service/ec2/types"
 	"github.com/gardener/gardener/extensions/pkg/controller"
 	"github.com/gardener/gardener/extensions/pkg/util"
+	gardencorev1beta1 "github.com/gardener/gardener/pkg/apis/core/v1beta1"
 	extensionsv1alpha1 "github.com/gardener/gardener/pkg/apis/extensions/v1alpha1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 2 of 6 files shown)


AndreasBurger (approved):
lgtm

PR #1770, opened 2026-04-16
Use the correct API version for MutatingAdmissionPolicy

DockToFuture:
**How to categorize this PR?**

/area networking
/kind enhancement
/platform aws

**What this PR does / why we need it**:
MutatingAdmissionPolicy graduates from v1alpha1 → v1beta1 in Kubernetes 1.34 and to v1 (GA) in 1.36. This PR makes the Calico mutating admission policy chart use the correct API version based on the shoot's Kubernetes version, and aligns the enable/disable logic accordingly:

  - `< 1.34`: v1alpha1, requires explicit feature gate + RuntimeConfig opt-in
  - `>= 1.34`: v1beta1, enabled by default (can be opted out)
  - `>= 1.36`: v1, always enabled (feature gate locked on)
  

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Release note**:

```other operator
Calico mutating admission policy chart uses the correct API version based on the shoot's Kubernetes version, and aligns the enable/disable logic accordingly.
```

PR #1841, opened 2026-06-15
Broaden MAP to suppress aws-custom-route-controller NetworkUnavailable=true cold-start race

DockToFuture:
## Why

When the overlay is disabled on a long-running shoot, [`aws-custom-route-controller`](https://github.com/gardener/aws-custom-route-controller) restarts with an empty in-memory route cache and patches every node it has not yet observed a VPC route for to `NetworkUnavailable=True`, even though the routes already exist. On a recent migration of a large prod shoot this caused **~5 minutes of user-visible downtime** even though the calico-scoped MutatingAdmissionPolicy was deployed and active — the MAP had no `matchCondition` for the route controller's ServiceAccount, so its 183 cold-start patches went straight through admission.

Detailed RCA: https://github.tools.sap/kubernetes-live/issues-live/issues/9754

## What

Broaden `block-calico-network-unavailable` to also intercept node/status writes from `system:serviceaccount:kube-system:aws-custom-route-controller`, but only when the request tries to **flip** `NetworkUnavailable` from `False` (or unset) to `True`.

This narrowly catches the cold-start race during overlay disablement without permanently silencing the route controller's authority over the condition:

- heartbeat refreshes that keep `NetworkUnavailable=True` on a genuinely broken node pass through unchanged
- `=False` writes always pass through, so nodes can still be marked network-ready once routes are in place
- `False`/unset → `True` flips from calico-node or aws-custom-route-controller are stripped (the condition is restored to its previous value)

## Tests

`charts/internal/shoot-system-components/charts/calico-mutating-admission-policy/chart_test.go` renders the chart and asserts:

- `enabled=false` produces no manifest
- `apiVersion` value flows into both `MutatingAdmissionPolicy` and `MutatingAdmissionPolicyBinding` for `v1alpha1`, `v1beta1`, `v1`
- Binding's `policyName` matches policy's `metadata.name`
- The first `matchCondition` references both `calico-node` and `aws-custom-route-controller` SAs
- The flip `matchCondition` requires the new condition to be `=True` AND the old condition to NOT already be `=True` (so heartbeat refreshes pass through)
- The mutation replaces `/status/conditions` with `finalConditions = conditionsWithoutNetworkUnavailable + oldNetworkUnavailableCondition`

A negative-test verification (reverting the chart to the old single-SA form) confirms the SA-coverage assertion correctly fails on regression.

## Caveats / future work

- The MAP is permanently active once deployed, not scoped to the migration window. Long-term the right shape is for the AWS provider extension to deploy the MAP only while overlay disablement is in flight, and remove it after. A `TODO` in the chart points at this.
- A new node legitimately starts with no route in AWS yet — the route controller will try to set `NetworkUnavailable=True`, the MAP will block it. That signal is lost during the brief new-node-bring-up window. Trade accepted: the alternative (5-min outage for every overlay disablement) is much worse.

## Test plan

- [ ] Existing unit tests continue to pass.
- [ ] On a test shoot with overlay enabled, set `MutatingAdmissionPolicy=true` + `runtimeConfig admissionregistration.k8s.io/v1beta1=true`, then disable overlay. Confirm the MAP is created and that the rolling calico DS restart does **not** see mass `NetworkUnavailable=true` flips on nodes.
- [ ] Confirm `aws-custom-route-controller` can still flip a node from `=True` to `=False` after creating its VPC route (i.e., recovery still works).
- [ ] Confirm a new node added during steady state still becomes Ready (its `=True` from the route controller is suppressed; the route controller's subsequent `=False` lands and the node is healthy).

### DIFF
--- a/charts/internal/shoot-system-components/charts/calico-mutating-admission-policy/chart_test.go
+++ b/charts/internal/shoot-system-components/charts/calico-mutating-admission-policy/chart_test.go
@@ -1,0 +1,319 @@
+// SPDX-FileCopyrightText: SAP SE or an SAP affiliate company and Gardener contributors
+//
+// SPDX-License-Identifier: Apache-2.0
+
+// Package calicomutatingadmissionpolicy contains tests for the calico
+// MutatingAdmissionPolicy chart. The tests render the chart for each
+// supported apiVersion and assert behavioural properties of the produced
+// MutatingAdmissionPolicy: that it targets both the calico-node and the
+// aws-custom-route-controller service accounts, that it only intercepts
+// flips of NetworkUnavailable from False (or unset) to True, and that it
+// strips NetworkUnavailable from the new conditions while restoring it
+// from the old object.
+package calicomutatingadmissionpolicy_test
+
+import (
+	"bytes"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+	"text/template"
+
+	"gopkg.in/yaml.v3"
+)
+
+const (
+	calicoNodeUser              = "system:serviceaccount:kube-system:calico-node"
+	awsCustomRouteControllerUser = "system:serviceaccount:kube-system:aws-custom-route-controller"
+)
+
+// renderCalicoMAPChart renders the chart's mutating-admission-policy.yaml for
+// the given values. It uses Go's text/template (which is the engine helm uses
+// under the hood) and intentionally does not pull in the helm engine for this
+// test, since the chart only references {{ .Values.enabled }} and
+// {{ .Values.apiVersion }} and uses no sprig helpers or chart partials.
+func renderCalicoMAPChart(t *testing.T, values map[string]interface{}) string {
+	t.Helper()
+
+	path := filepath.Join("templates", "mutating-admission-policy.yaml")
+	raw, err := os.ReadFile(path)
+	if err != nil {
+		t.Fatalf("read chart template: %v", err)
+	}
+
+	// helm's whitespace-trim syntax {{- ... -}} is identical to Go's
+	// text/template trim markers. The chart additionally uses sprig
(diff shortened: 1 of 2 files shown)


DockToFuture:
Closing this PR. The premise — that aws-custom-route-controller was mass-patching nodes to NetworkUnavailable=True during overlay disablement on prod-haas — turned out to be wrong on closer inspection of the controller logs.

The log line we read as "183 nodes flipped to NetworkUnavailable=True":

```
Patching node status <node> with true previous condition was: nil
```

is from [reconciler.go:259](https://github.com/gardener/aws-custom-route-controller/blob/v0.15.0/pkg/controller/reconciler.go#L259), where the second %v is the `routesCreated` bool. `routesCreated=true` actually triggers the [reconciler.go:268-275](https://github.com/gardener/aws-custom-route-controller/blob/v0.15.0/pkg/controller/reconciler.go#L268-L275) branch that writes `NetworkUnavailable=False, Reason=RouteCreated`. So those 183 writes were *=False recovery writes*, not =True flips.

The current node state in the cluster confirms this: all 192 nodes show `NetworkUnavailable=False, Reason=RouteCreated, lastTransitionTime ≈ 12:17:05–12:17:39` — single transition, exactly when the controller patched.

The only writer of `NetworkUnavailable=True` on prod-haas was calico-node itself on 7 nodes (Reason=CalicoIsDown) during pod teardown. The existing calico-only MAP (once it actually deploys, fixed in v1.70.0 by #1770) handles that case. Broadening the MAP's matchConditions to also intercept aws-custom-route-controller is solving a non-problem.

The actual outage on prod-haas (genuine NPD probe failures 12:14:48–12:17:48) appears to be data-plane disruption during the calico DS rollout in non-overlay mode, not anything the MAP can prevent. That's a separate investigation.

PR #1842, opened 2026-06-16
Restrict MutatingAdmissionPolicy to v1alpha1 for K8s < 1.34

DockToFuture:
## Summary
- For Kubernetes versions below 1.34, `MutatingAdmissionPolicy` is only served via `admissionregistration.k8s.io/v1alpha1`.
- `isMutatingAdmissionPolicyEnabled` no longer treats `v1beta1` in the runtime config as enabling the feature on pre-1.34 clusters.
- `mutatingAdmissionPolicyAPIVersion` always returns `v1alpha1` for pre-1.34 clusters, even when `v1beta1` is set in the runtime config.
- Tests updated to reflect the corrected behavior.

## Test plan
- [x] `go test ./pkg/controller/controlplane/...`


wpross (approved):
/lgtm

PR #1645, opened 2026-01-16
[GEP-33] Allow mixed format Capability CloudProfiles

Roncossek:
**How to categorize this PR?**
/area control-plane
/kind enhancement
/platform aws

**What this PR does / why we need it**:

This PR enables mixed format support for machine image specifications in CloudProfile and NamespacedCloudProfile resources, allowing smooth migration from the legacy architecture-per-region format to the new capabilityFlavors format.

**Background:**
Previously, CloudProfiles could only use one format per profile:
- **Old format**: `regions` with `architecture` field per region
- **New format**: `capabilityFlavors` with capabilities (introduced in GEP-33)

This created a breaking change scenario where users had to migrate all image versions at once.

**Changes:**
1. **CloudProfile Support**: When `spec.machineCapabilities` is defined, both old and new formats are now supported per image version
   - Validation accepts both formats
   - Each version can use either format independently

2. **NamespacedCloudProfile Support**: Removes format transformation to parent CloudProfile format
   - Mixed format is preserved as-is
   - Validation handles both formats

3. **Image Selection Logic**: Updated to handle both formats during worker/bastion machine image selection
   - Automatically converts old format regions to capabilityFlavors internally
   - Maintains backward compatibility for non-capabilities CloudProfiles

**Example:**
```yaml
providerConfig:
  machineImages:
    - name: ubuntu
      versions:
        # New format - with capabilities
        - version: "24.04"
          capabilityFlavors:
            - capabilities:
                architecture: [amd64]
                hypervisor: [baremetal, xen]
              regions:
                - name: eu-west-1
                  ami: ami-456
        # Old format - backwards compatible
        - version: "22.04"
          regions:
            - name: eu-west-1
              ami: ami-123
              architecture: amd64
```

**Which issue(s) this PR fixes**:
Part of gardener/gardener#11301

**Special notes for your reviewer**:
@hebelsan

**Release note**:
``` feature operator
Support mixed format for machine image specifications in CloudProfile and NamespacedCloudProfile, allowing both (legacy) architecture-per-region format and new capabilityFlavors format to coexist within the same CloudProfile for smooth migration.
```

PR #1846, opened 2026-06-26
fix: ValidateWorkersAgainstCloudProfileOnUpdate uses new worker architecture

Roncossek:

/area quality
/kind bug
/platform aws

Fixes a regression in the admission validator's update path. When adding a new worker to an existing shoot, `ValidateWorkersAgainstCloudProfileOnUpdate` passed the *old* worker's `Machine.Architecture` to `NormalizeMachineTypeCapabilities`. For a brand-new worker the old worker is the zero value, so the architecture was `nil` and silently defaulted to `amd64` — causing arm64 workers referencing arm64-only image flavors to be rejected with `could not determine best flavor: no compatible flavor found`. The fix uses the new worker's architecture, matching the already-correct `OnCreation` path. Adds a regression test plus a counter-case that confirms genuine architecture/image mismatches are still rejected.

Fixes #

Regression introduced in #1645 ([GEP-33] Allow mixed format Capability CloudProfiles). Only the update path was affected; new-shoot creation was already correct.

**Release note**:
```bugfix operator
Fix admission validator rejecting new arm64 worker pools added to existing shoots when the CloudProfile defines no `capabilityDefinitions`.
```

### DIFF
--- a/pkg/apis/aws/validation/worker.go
+++ b/pkg/apis/aws/validation/worker.go
@@ -215,11 +215,11 @@ func ValidateWorkersAgainstCloudProfileOnUpdate(
 		if w.Name == "" || newWorker.Machine.Image != w.Machine.Image {
 			machineType := gardencorev1beta1helper.FindMachineTypeByName(machineTypes, newWorker.Machine.Type)
 			if machineType == nil {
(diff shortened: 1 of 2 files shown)


MartinWeindel (commented):
/lgtm

wpross (approved):
/lgtm

PR #1861, opened 2026-06-30
remove replace for cloud.google.com/go

matthias-horne:

/area compliance
/kind enhancement
/platform aws

The replace for `cloud.google.com/go` is no longer required, so we should not pin to an old version.

Fixes #

**Release note**:

```other operator

```


hebelsan (commented):
/lgtm
/approve

PR #1879, opened 2026-07-08
Fix EIP validation to exclude deleted NAT gateways

hebelsan:

/area quality
/kind enhancement
/platform aws

`GetNATGatewayAddressAllocations` retrieves all allocation IDs from NAT gateways tagged with `kubernetes.io/cluster/<shoot-namespace>`, including those in deleted state. AWS retains the AllocationId on deleted NAT gateways in its API response.

This could caused `validateEIPS` to incorrectly consider a previously-used EIP as still belonging to the Shoot's NAT gateways, even after the NAT gateway was deleted and the EIP was disassociated.

Fixes #

**Release note**:

```other operator
Fix EIP validation to exclude deleted NAT gateways
```


hebelsan:
Closing as issue can not be reproduced.

PR #1878, opened 2026-07-08
Add support for Kubernetes 1.36

tobschli:
**How to categorize this PR?**

/area open-source usability
/kind enhancement
/platform aws
/topology garden seed shoot

**What this PR does / why we need it**:
This PR adds support for Kubernetes 1.36 to the extension.

**Which issue(s) this PR fixes**:
Part of gardener/gardener#14653

**Special notes for your reviewer**:

- Functionality validations:
  - [ ] Create new clusters with versions < 1.36
  - [ ] Create new clusters with version = 1.36
  - [ ] Upgrade old clusters from version 1.35 to version 1.36
  - [ ] Delete clusters with versions < 1.36
  - [ ] Delete clusters with version = 1.36

**Release note**:
```feature user
This extension now supports shoot clusters with Kubernetes version 1.36.
You should consider the Kubernetes release notes before upgrading to 1.36.
```

PR #1885, opened 2026-07-13
Revert "Add support for Kubernetes 1.36"

matthias-horne:
Reverts gardener/gardener-extension-provider-aws#1878

The upstream cloud-provider-aws images is not yet available in `registry.k8s.io/provider-aws/cloud-controller-manager` causing our builds to fail.

See also: https://github.com/kubernetes/cloud-provider-aws/issues/1450


AndreasBurger (approved):
lgtm

PR #1891, opened 2026-07-16
Fix fluentbit conflict on hybrid Garden cluster

Wieneo:

/area robustness
/kind bug
/platform aws

If a Garden cluster is also a seed cluster at the same time, the FluentBit `ClusterFilter` resources get applied by the runtime and seed deployment of the extension.
This causes the deletion of the `ControllerInstallation` to get stuck as two `ManagedResource`s fight over the same resources.

If I understood correctly, the `ClusterFilter`s should only be applied to seeds.

**Release note**:

```bugfix operator
Fixed a bug that caused the FluentBit `ClusterFilter`s of this extension to be applied to the Garden cluster.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated AWS cluster filtering so logging filters are omitted when runtime clusters are enabled.
  * Refreshed the AWS provider controller deployment chart configuration.


charts/gardener-extension-provider-aws/templates/clusterfilters-logging.yaml
```diff
@@ -1,3 +1,4 @@
+{{ if not (include "runtimeCluster.enabled" .) }}
```
timuthy, line 1: With this conditional statement, the `ClusterFilter` is not deployed to exclusive Garden Runtime clusters, i.e. those not serving as a seed at the same time.

There is an enhancement in https://github.com/gardener/gardener/pull/15566 that allows distinguishing both cases. After the PR is merged and released, the condition must be adjusted to:

```
{{ if or (and .clusterTypes.gardenRuntimeCluster (not .clusterTypes.seedCluster)) and (not .clusterTypes.gardenRuntimeCluster) .clusterTypes.seedCluster) }}
```

hebelsan: @timuthy, I'm a bit confused about the condition you proposed.
If I'm getting it correctly your condition is XOR, which would skip the hybrid case (gardenRuntimeCluster=true AND seedCluster=true).
I thought `{{- if .Values.gardener.clusterTypes.seedCluster }}` would be sufficient?
Can you clarify: in which cluster types should the ClusterFilters be present?

timuthy: I might have misunderstood the requirement for the `ClusterFilter`, also due to an offline discussion. In the end, the current change looks fine from my side 👍

timuthy (commented):
/lgtm

AndreasBurger (dismissed):

AndreasBurger (approved):
rebased and ran `make generate`

lgtm, thanks!

AndreasBurger:
Seeing these test-failures on master as well. I'll try to figure out whats going on and then fix/rebase this PR

PR #1919, opened 2026-07-30
Add `minStepAbsolute` defaulting to pvc-autoscaler webhook

RadaBDimitrova:

/area storage
/kind enhancement
/platform aws

AWS EBS volumes can only be modified a limited number of times within a rolling 24-hour period. Enforcing a minimum absolute scaling step of `5Gi` (in addition to the cooldown) ensures each scaling operation makes a meaningfully large change, reducing the risk of exhausting the modification quota through many small resizes.

Part of https://github.com/gardener/gardener/issues/14383 and related to #1745

@plkokanov @Kostov6 

**Release note**:

```other operator
Extends the `PersistentVolumeClaimAutoscaler` mutating webhook to default and enforce a minimum `MinStepAbsolute` of `5Gi` on scale-up policies, alongside the existing `CooldownDuration` defaulting.
```

## Summary by CodeRabbit

* **New Features**
  * Scale-up volume policies now enforce a minimum absolute scaling step of 5 GiB.
  * Policies without a configured minimum automatically default to 5 GiB.
  * Configured values below 5 GiB are raised to the minimum, while larger values remain unchanged.
  * This provides consistent minimum growth behavior across all configured volume policies.

### DIFF
--- a/pkg/webhook/pvca/mutator.go
+++ b/pkg/webhook/pvca/mutator.go
@@ -12,12 +12,16 @@ import (
 	extensionswebhook "github.com/gardener/gardener/extensions/pkg/webhook"
 	pvcautoscalingv1alpha1 "github.com/gardener/pvc-autoscaler/api/autoscaling/v1alpha1"
 	"github.com/go-logr/logr"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 // A volume can be modified up to four times within a rolling 24-hour period (https://docs.aws.amazon.com/ebs/latest/userguide/ebs-modify-volume.html)
 const minCooldownDuration = 6 * time.Hour
+
+// minStepAbsolute is the minimum absolute change in capacity enforced during scaling.
+var minStepAbsolute = resource.MustParse("5Gi")
 
 type mutator struct {
 	logger logr.Logger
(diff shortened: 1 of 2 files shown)


pkg/webhook/pvca/mutator.go
```diff
@@ -42,6 +46,11 @@ func (m *mutator) Mutate(_ context.Context, newObj, _ client.Object) error {
 			(pvca.Spec.VolumePolicies[i].ScaleUp.CooldownDuration == nil || pvca.Spec.VolumePolicies[i].ScaleUp.CooldownDuration.Duration < minCooldownDuration) {
 			pvca.Spec.VolumePolicies[i].ScaleUp.CooldownDuration = &metav1.Duration{Duration: minCooldownDuration}
 		}
+		if pvca.Spec.VolumePolicies[i].ScaleUp != nil &&
+			(pvca.Spec.VolumePolicies[i].ScaleUp.MinStepAbsolute == nil || pvca.Spec.VolumePolicies[i].ScaleUp.MinStepAbsolute.Cmp(minStepAbsolute) < 0) {
+			step := minStepAbsolute.DeepCopy()
```
plkokanov, line 51: Why do we need the `DeepCopy` here?

plkokanov:
@hebelsan @wpross, when you have the time, could you check this PR? :)

pkg/webhook/pvca/mutator.go
```diff
@@ -42,6 +46,10 @@ func (m *mutator) Mutate(_ context.Context, newObj, _ client.Object) error {
 			(pvca.Spec.VolumePolicies[i].ScaleUp.CooldownDuration == nil || pvca.Spec.VolumePolicies[i].ScaleUp.CooldownDuration.Duration < minCooldownDuration) {
 			pvca.Spec.VolumePolicies[i].ScaleUp.CooldownDuration = &metav1.Duration{Duration: minCooldownDuration}
 		}
+		if pvca.Spec.VolumePolicies[i].ScaleUp != nil &&
+			(pvca.Spec.VolumePolicies[i].ScaleUp.MinStepAbsolute == nil || pvca.Spec.VolumePolicies[i].ScaleUp.MinStepAbsolute.Cmp(minStepAbsolute) < 0) {
+			pvca.Spec.VolumePolicies[i].ScaleUp.MinStepAbsolute = &minStepAbsolute
```
hebelsan: This hands every defaulted policy a pointer to the same package-level minStepAbsolute global. It's safe on today's code path but if any future caller invokes a pointer-receiver method like String() on the shared value, it becomes shared mutable state across concurrent admission requests.

I'd suggest allocating a fresh copy per policy, which also keeps it consistent with the CooldownDuration line right above that already allocates fresh.

plkokanov: Yeah, I actually suggested the opposite to @RadaBDimitrova, since we don't do it anywhere in the code like that (also in g/g). 
But you're generally right, so if it's 2 against 1, let me revert it :)

plkokanov: Done in https://github.com/gardener/gardener-extension-provider-aws/pull/1919/changes/4c059237d4fb525a3042ad536e37bfb87a23e8b3

hebelsan (commented):
/lgtm
/approve

PR #1930, opened 2026-08-05
Retrieve the natgateway by cluster key only

hebelsan:

/area quality
/kind enhancement
/platform aws

This PR fixes the natgateway retrieval in the byo elastic IP case.
With the byo everything PR we switched from TagValueCluster ("1") to TagValueClusterOwned ("owned").
This should be respected everywhere.
Instead of checking the tag value we only check the tag key from now on.

Fixes #

**Release note**:

```other operator
NONE
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved NAT gateway discovery by recognizing resources with the cluster tag regardless of its tag value.


matthias-horne (commented):
/lgtm
/approve

PR #890, opened 2024-03-05
Upgrade terraformer in v1.51

AndreasBurger:
**Special notes for your reviewer**:
Updates by two minor versions.

**Release note**:

```other operator
Update terraformer to v2.25.0
```

PR #1957, opened 2026-08-21
Support opt-in cleanup of out-of-band IAM role attachments on infrastructure deletion

### LINKED ISSUE #1956: Shoot deletion wedges on out-of-band IAM policy attachments to the nodes role — support opt-in force-detach during infrastructure deletion

/area robustness
/kind enhancement
/platform aws

An **opt-in** mechanism (e.g. a shoot annotation such as
`aws.provider.extensions.gardener.cloud/force-detach-role-policies: "true"`) that makes
the infrastructure **deletion** flow clean up all remaining attachments on the nodes IAM
role (`shoot--<project>--<name>-nodes`) before calling `DeleteRole`:

1. `ListAttachedRolePolicies` → `DetachRolePolicy` for every remaining managed-policy
   attachment (the policies themselves are **not** deleted — only their attachment to
   this one role),
2. `ListRolePolicies` → `DeleteRolePolicy` for remaining inline policies (inline
   policies are role-owned by definition),
3. `ListInstanceProfilesForRole` → `RemoveRoleFromInstanceProfile` for remaining
   instance-profile associations,

each action logged with the role name and policy ARN / policy name / profile name for
auditability. Default behavior stays exactly as today; reconciliation is never affected.

I am happy to contribute the PR — the change is small and self-contained: a new task in
[`buildDeleteGraph()`](https://github.com/gardener/gardener-extension-provider-aws/blob/2ff8d81837ddd58a79e23a579baa43623d43faa6/pkg/controller/infrastructure/infraflow/delete.go#L37-L106) ordered
between the existing IAM cleanup tasks and `delete IAM role`, four thin paginated
wrappers in [`pkg/aws/client`](https://github.com/gardener/gardener-extension-provider-aws/tree/2ff8d81837ddd58a79e23a579baa43623d43faa6/pkg/aws/client), and an annotation constant. Notably, **no new IAM
permissions are required**: the documented minimal policy for the cloud credentials
already contains `iam:ListAttachedRolePolicies`, `iam:DetachRolePolicy`,
`iam:ListRolePolicies`, `iam:ListInstanceProfilesForRole`, and
`iam:RemoveRoleFromInstanceProfile` (see the [documented policy](https://github.com/gardener/gardener-extension-provider-aws/blob/2ff8d81837ddd58a79e23a579baa43623d43faa6/docs/usage/usage.md#L166-L253)).

**Why is this needed**:

On shoot deletion, provider-aws deletes the nodes IAM role it created. AWS returns

```
DeleteConflict: Cannot delete entity, must detach all policies first. (409)
```

if **any** policy is still attached. The deletion flow only removes what it manages
(its own inline policy and instance profile), so any policy attached **out of band**
wedges the entire shoot deletion at "Waiting until shoot infrastructure has been
deleted", surfacing as `Error deleting Infrastructure: ... failed to "delete IAM role":
operation error IAM: DeleteRole ...`.

Out-of-band attachments to the nodes role are common in enterprise AWS accounts and are
usually not under the shoot owner's control:

- **account-governance / security-baseline automation** that attaches a mandatory
  policy (agent permissions, SSM access, etc.) to every instance role in the account —
  in our fleet this affects the vast majority of shoots across dozens of accounts;
- **platform/application teams** attaching functional policies directly to the nodes
  role (backup tooling, KMS unseal, EFS/Route53/S3 access, etc.);
- **operators** attaching e.g. `AmazonSSMManagedInstanceCore` by hand for debugging and
  forgetting to remove it.

Today the operator's options are all bad:

- **Manual runbook:** fetch the shoot's cloud credentials, `aws iam
  list-attached-role-policies` + `detach-role-policy`, then annotate the shoot with
  `gardener.cloud/operation=retry`. Works (deletion completes within seconds once
  detached), but requires credential handling and an operator for every affected
  deletion.
- **Force-deletion** (`confirmation.gardener.cloud/force-deletion`): skips *all* cloud
  cleanup and permanently orphans the IAM role and everything else remaining. We found
  multiple such fossil roles from past wedged deletions in a single account — this
  pattern silently leaks IAM resources fleet-wide.

An opt-in force-detach removes the need for both: annotate + retry, and the normal
deletion path completes cleanly with a full audit trail in the extension log.

Safety properties of the proposal:

- **Deletion-only** — never runs during reconcile; attachments are untouched during the
  shoot's lifetime.
- **Scoped** — acts only on the one role the flow has already decided to delete (same
  guard as the existing `delete IAM role` task); no other role or policy is touched. If
  the role's ARN is recorded in the infrastructure state, the cleanup additionally
  refuses to act on a live role whose ARN does not match (guards against identically
  named roles of another Gardener installation sharing the same account).
- **Detach, not delete** — managed/customer policy objects and their attachments to
  other roles remain intact.
- **Opt-in** — the default is today's behavior; the annotation is a deliberate,
  per-shoot operator action (and, being an annotation, it can still be set after the
  deletion has already started and wedged, unlike anything in the spec).
- **Idempotent** — all steps ignore `NoSuchEntity`, so retries converge.
- **Partition-agnostic** — ARNs are taken from the `ListAttachedRolePolicies` response,
  so it works identically in the `aws` and `aws-us-gov` partitions.

**Prior art in comparable tools**:

This is a well-trodden problem with established solutions elsewhere:

- **Terraform** added an opt-in `force_detach_policies` argument to `aws_iam_role`
  for exactly this scenario — a third-party account-governance tool attaching
  policies to every role and wedging `terraform destroy`
  (hashicorp/terraform-provider-aws#883, implemented in #890). It was later extended
  to also delete inline policies because detaching managed policies alone still left
  deletion blocked (#2279, fixed in #2388) — which is why the proposal above covers
  managed attachments, inline policies, and instance-profile associations together.
  The proposed annotation name deliberately mirrors Terraform's established flag.
- **kOps** goes further and *unconditionally* detaches all attached policies and
  deletes all inline policies before deleting IAM roles during `kops delete cluster`
  (kubernetes/kops#1644; [`DeleteIAMRole` in `pkg/resources/aws/aws.go`](https://github.com/kubernetes/kops/blob/301e80c5dfb5/pkg/resources/aws/aws.go#L1900-L1980)).
- AWS's own `DeleteRole` API documentation prescribes this exact pre-deletion
  cleanup (delete inline policies, detach managed policies, remove the role from
  instance profiles), so the proposed task simply automates the AWS-documented
  deletion order behind an explicit opt-in.

If maintainers prefer, the same behavior could additionally be defaultable via the
ControllerConfiguration for fleets where account-baseline automation makes this the
norm — but the per-shoot annotation alone already solves the operational problem, so
I'd suggest starting there.

amendezsap:

/area robustness
/kind enhancement
/platform aws

AWS rejects `DeleteRole` with a `DeleteConflict` error while any policy is still attached
to the role. The infrastructure deletion flow only removes the attachments it manages
itself, so policies attached to the nodes IAM role out-of-band (e.g. by
account-governance or security-baseline automation, or manually by administrators) leave
the shoot deletion permanently failing at `failed to "delete IAM role"` until an operator
detaches them by hand with the shoot's cloud credentials. The only generic escape today is
force-deletion, which skips cloud cleanup entirely and orphans the role.

This PR adds an opt-in shoot annotation:

```yaml
aws.provider.extensions.gardener.cloud/force-detach-role-policies: "true"
```

When set, the deletion flow runs one additional task right before `delete IAM role` that

- detaches all still-attached managed policies (the policies themselves are never deleted),
- deletes remaining inline policies of the role (inline policies are role-owned by definition),
- removes remaining instance-profile associations of the role,

logging every removed attachment (ARN / policy name / instance profile name) for
auditability.

Safety properties:

- **Opt-in, deletion-only**: default behavior is unchanged; attachments are never touched
  while the shoot exists. Since annotations (unlike `.spec`) can still be changed on a
  shoot that is already marked for deletion, the annotation can also rescue a deletion
  that is already stuck — annotate and retry.
- **Scoped**: the task runs under the same flow-state guard as the existing
  `delete IAM role` task, and is ordered after the flow's own IAM cleanup tasks, so it
  only ever sees genuine out-of-band leftovers. Additionally, if the role's ARN is
  recorded in the infrastructure state, the task refuses to act on a live role whose ARN
  does not match — guarding against identically named roles of another Gardener
  installation sharing the same account.
- **Idempotent**: all steps ignore `NoSuchEntity`, so retries converge.
- **No new permissions**: all required IAM actions (`iam:ListAttachedRolePolicies`,
  `iam:DetachRolePolicy`, `iam:ListRolePolicies`, `iam:ListInstanceProfilesForRole`,
  `iam:RemoveRoleFromInstanceProfile`) are already part of the [documented minimal policy](https://github.com/gardener/gardener-extension-provider-aws/blob/2ff8d81837ddd58a79e23a579baa43623d43faa6/docs/usage/usage.md#L166-L253).
- **Partition-agnostic**: policy ARNs are taken from the `ListAttachedRolePolicies`
  response, so this works identically in the `aws` and `aws-us-gov` partitions.

Prior art: Terraform's `aws_iam_role` offers the analogous opt-in `force_detach_policies`
argument for exactly this scenario (hashicorp/terraform-provider-aws#883, #890, #2388 —
the last one is why inline policies are covered too), and kOps unconditionally detaches
all policies before role deletion during `kops delete cluster` (kubernetes/kops#1644; [`DeleteIAMRole` in `pkg/resources/aws/aws.go`](https://github.com/kubernetes/kops/blob/301e80c5dfb5/pkg/resources/aws/aws.go#L1900-L1980)).
The annotation name deliberately mirrors Terraform's established flag.

Fixes #1956

- The new AWS client methods are thin paginated wrappers following the existing style in
  [`pkg/aws/client`](https://github.com/gardener/gardener-extension-provider-aws/tree/2ff8d81837ddd58a79e23a579baa43623d43faa6/pkg/aws/client); mocks were regenerated with `mockgen`.
- Unit tests for the new flow task cover: no-op without flow state, no-op when the role is
  gone, refusal on ARN mismatch, the full cleanup path, and the no-recorded-ARN fallback.
- If maintainers would prefer an additional landscape-wide default via the
  ControllerConfiguration (annotation as per-shoot override), I'm happy to follow up — the
  per-shoot annotation alone already covers the operational need, so this PR keeps the
  surface minimal.

**Release note**:

```feature user
The new shoot annotation `aws.provider.extensions.gardener.cloud/force-detach-role-policies: "true"` allows the infrastructure deletion flow to detach all remaining (out-of-band) managed policies from the nodes IAM role, delete its remaining inline policies, and remove remaining instance-profile associations before deleting the role. This lets shoot deletion succeed when policies were attached to the nodes role outside of Gardener (e.g. by account-governance automation), without resorting to force-deletion. The annotation can also be set on a shoot whose deletion is already stuck on a `DeleteConflict` error; retry the operation afterwards.
```

## Summary by CodeRabbit

- **New Features**
  - Added an optional cleanup setting for Shoot deletion that removes remaining IAM managed and inline policies and instance-profile associations from the created nodes role.
  - Cleanup validates the role identity and records audit information, helping recover deletions that are already stuck.
  - Cleanup runs only during deletion and is enabled through the documented annotation.

- **Documentation**
  - Added configuration guidance, behavior details, and required IAM permissions for enabling the cleanup.


amendezsap:
@coderabbitai review

amendezsap:
@coderabbitai full review

amendezsap:
@coderabbitai full review

hebelsan:
Thanks for the contribution and the well crafted PR!  🚀
/lgtm

hebelsan:
Verified functionality in the local remote setup ✅

hebelsan (approved):

amendezsap:
Thanks for the review @hebelsan! It looks like prow didn't pick up the `/lgtm` — it seems to only register commands on their own line, so the label was never applied. Would you mind re-issuing it as a standalone line when you get a chance?

PR #1959, opened 2026-08-24
fix: EFS FindFileSystemsByTags pagination

hebelsan:

/area robustness
/kind bug
/platform aws

The AWS API returns the EFS results in pages.
We should go through all of them to not miss something.

Fixes #

**Release note**:

```bugfix operator
fix EFS FindFileSystemsByTags pagination
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved file system discovery to include all paginated results.
  * Ensured all tags are retrieved across multiple pages.
  * Tag retrieval failures are now reported instead of silently skipped.

### DIFF
--- a/pkg/aws/client/client.go
+++ b/pkg/aws/client/client.go
@@ -2382,8 +2382,7 @@ func (c *Client) FindFileSystemsByTags(ctx context.Context, tags Tags) ([]*Elast
 		for _, fs := range output.FileSystems {
 			fsTags, err := c.listFileSystemTags(ctx, fs.FileSystemId)
 			if err != nil {
-				c.Logger.Info("could not get tags for fs %s: %v", ptr.Deref(fs.FileSystemId, ""), err)
-				continue
+				return nil, fmt.Errorf("could not get tags for file system %s: %w", ptr.Deref(fs.FileSystemId, ""), err)
 			}
 
 			if tags.ContainEfsTags(fsTags) {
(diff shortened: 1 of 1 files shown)


pkg/aws/client/client.go
```diff
@@ -2372,26 +2372,49 @@ func (c *Client) GetFileSystem(ctx context.Context, fileSystemID string) (*Elast
...
-			c.Logger.Info("could not get tags for fs %s: %v", ptr.Deref(fs.FileSystemId, ""), err)
-			continue
+			return nil, ignoreNotFound(err)
+		}
+
+		for _, fs := range output.FileSystems {
+			fsTags, err := c.listFileSystemTags(ctx, fs.FileSystemId)
+			if err != nil {
+				c.Logger.Info("could not get tags for fs %s: %v", ptr.Deref(fs.FileSystemId, ""), err)
+				continue
```
kon-angelo: why do we continue in this error ?

hebelsan: `listFileSystemTags` is best-effort here.
If we can't fetch tags for one filesystem, we skip it rather than aborting the entire list operation. Failing the whole call over a single filesystem's tag retrieval error could unnecessarily block reconciliation.

hebelsan: Additional context:
1. `ListTagsForResource` runs per-FS across the whole account. DescribeFileSystems returns every filesystem the credentials can see, not just this shoot's. Erroring would fail our reconcile over a foreign FS.
2. EFS tag APIs get rate-limited; per-FS calls in a large account hit ThrottlingException routinely. Hard-failing on those makes the function flaky exactly when the account is busy.
3. The caller resolves by ID first (FindExisting) and runs in a retry loop, so the tag scan is a fallback.

kon-angelo: How do we know that we did not error out on fetching our shoot's FS?

hebelsan: You are probably right, the risk of leaked resources is too high.
I switched to returning the error instead.
Also if throttling turns out to be a real concern we can still introduce a retry mechanism.

kon-angelo (approved):

PR #1974, opened 2026-09-01
Align with provider local

hebelsan:

/area robustness
/kind enhancement
/platform aws

Aligns the extension-class filtering in all controllers with the approach used in provider-local.
- Removes the early-return skip-guard (if len(classes) == 0 { return nil }) from all controller add.go files. The guard was added defensively but is unnecessary — when no supported classes remain after filtering, the HasClass predicate (which defaults an empty list to [shoot]) and the --controllers name-based switch already handle controller gating correctly.
- Moves slices.Clone of ExtensionClasses to the app.go assignment site (one clone per controller), matching provider-local exactly. This fixes a subtle shared-slice mutation bug: slices.DeleteFunc compacts the backing array in place, so without cloning, filtering in one controller registration path could corrupt the slice seen by a later controller, leading to incomplete controller registration.

Fixes #

**Release note**:

```other operator
NONE
```

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved controller registration when no supported extension classes are configured.
  - Ensured AWS controller options are isolated when shared configuration is used.
  - Prevented valid controllers from being skipped due to empty filtered configuration.

### DIFF
--- a/cmd/gardener-extension-provider-aws/app/app.go
+++ b/cmd/gardener-extension-provider-aws/app/app.go
@@ -8,6 +8,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"slices"
 	"strings"
 	"time"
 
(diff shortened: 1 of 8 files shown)


kon-angelo (approved):
/lgtm

